PR #46, opened 2019-02-03
Fix comparisons to `nil` in queries & make such queries explicit

### LINKED ISSUE #38: Disallow `nil` when querying. Require using `is_nil` instead
If you want to allow a value *or* nil maybe have a method like `is?(something_that_might_be_nil)`. There may be some other name that is better. Will have to think about it a bit

HarrisonB:
Continuation/migration of https://github.com/luckyframework/lucky_record/pull/289. See there for details.

Closes #38.

### DIFF
--- a/src/avram/criteria.cr
+++ b/src/avram/criteria.cr
@@ -17,7 +17,7 @@ class Avram::Criteria(T, V)
   end
 
   def is(_value : Nil)
-    {{ raise "Use `is?` instead of `is` if trying to check against nillable value." }}
+    {{ raise "Use `is?` instead of `is` if trying to check against a nilable value." }}
   end
 
   def is(value)
@@ -48,7 +48,7 @@ class Avram::Criteria(T, V)
   end
 
   def is_not(_value : Nil)
-    {{ raise "Use `is_not?` instead of `is_not` if trying to check against nillable value." }}
+    {{ raise "Use `is_not?` instead of `is_not` if trying to check against a nilable value." }}
   end
 
   def is_not?(value) : T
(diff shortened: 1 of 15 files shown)


HarrisonB:
@paulcsmith Continuing on some questions from the previous PR:

> And actually I'm wondering if we even need `is_not` since we have `.not.eq` and `.not.nilable_eq`.

I think you're totally right here. Let's get rid of them. Do we want to eliminate `not(value)` as well?

paulcsmith:
Hmm great question. I kind of like the shorthand “not” method. But at the same time I like the explicitness of “not.eq”

I think we should try removing it and we can add it back later if we get sick of typing “not.eq”. Does that sound good?

HarrisonB:
Sounds grrrrreaaat!

image

src/avram/criteria.cr
```diff
@@ -16,24 +16,29 @@ class Avram::Criteria(T, V)
...
+
+  def nilable_eq(_value : Nil)
+    add_clause(Avram::Where::Null.new(column))
     rows
   end
 
-  def not(value) : T
-    is_not(value)
+  def nilable_eq(value)
+    add_clause(Avram::Where::Equal.new(column, V::Lucky.to_db!(value)))
```
paulcsmith: I think this might make more sense with the method body swapped with `eq`. So this method body would replace `eq` and this method (`nilable_eq(value)`) would call `eq(value)`. Does that make sense?

HarrisonB:
@paulcsmith Should I open up a separate pull request for that?

paulcsmith:
@HarrisonB Yes that'd be great! It would be on the Lucky repo and not Avram. I'm trying to keep all Lucky framework relevant changes there so people using master now what changed and so when I make upgrade notes I don't forget anything :P

paulcsmith:
Thanks for tackling this 🎉 :D

PR #47, opened 2019-02-03
Allow adding UUID columns (that are not associations)

hanneskaeufler:
Adding a simple UUID column failed with "undefined method add_column"
because the UUID type was not recognized for default values.

A static default for UUIDs is kind of ... unintuitive ... anyway, but this gets me going for now.
What I think should be the next step is to maybe allow passing a block or something to fill_existing_with: ... in order to get a different value for each entry in the db. But I see that as a separate task to tackle?

(Storing a UUID in postgres can be done in UUID, but generating one requires an extension, so I opted out of something like DEFAULT generate_uuid_v1()).

### DIFF
--- a/spec/migrator/alter_table_statement_spec.cr
+++ b/spec/migrator/alter_table_statement_spec.cr
@@ -13,6 +13,7 @@ describe Avram::Migrator::AlterTableStatement do
       add joined_at : Time, default: :now
       add updated_at : Time, fill_existing_with: :now
       add future_time : Time, default: Time.new
+      add new_id : UUID, default: UUID.new("46d9b2f0-0718-4d4c-a5a1-5af81d5b11e0")
       remove :old_field
       remove_belongs_to :employee
     end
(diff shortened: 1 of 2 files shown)


hanneskaeufler:
@edwardloveall thanks for the ping on https://github.com/luckyframework/lucky_record/pull/302#issuecomment-460004160, I rebased and re-opened this PR here.

jwoertink:
You can already [add UUID columns](https://github.com/luckyframework/avram/blob/master/spec/migrator/create_table_statement_spec.cr#L25), but not with a default. This PR allows you to set that UUID default which I think may also be able to help fix https://github.com/luckyframework/avram/issues/92 with a few small extra additions. 

I was able to run this locally, and all specs still pass with latest Avram, so this all looks good to me! Thanks @hanneskaeufler

paulcsmith:
And thanks @hanneskaeufler for working on this :)

PR #48, opened 2019-02-04
Jsonb support

### LINKED ISSUE #18: Add support for Postgres JSON/JSONB column types
Also requested in the Lucky repo: https://github.com/luckyframework/lucky/issues/612. Not sure if this should be tackled along with or separate from some custom (de)serialization for attributes.

tzar:
See https://github.com/luckyframework/lucky_record/pull/275 for prior discussion

Ninja edit: closes #18 and https://github.com/luckyframework/lucky/issues/612

### DIFF
--- a/src/avram/migrator/column_default_helpers.cr
+++ b/src/avram/migrator/column_default_helpers.cr
@@ -1,5 +1,5 @@
 module Avram::Migrator::ColumnDefaultHelpers
-  alias ColumnDefaultType = String | Time | Int32 | Int64 | Float32 | Float64 | Bool | Symbol
+  alias ColumnDefaultType = Object
 
   def value_to_string(type : String.class | Time.class, value : String | Time)
     "'#{value}'"
@@ -43,5 +43,9 @@ module Avram::Migrator::ColumnDefaultHelpers
 
   def default_value(type : Time.class, default : Symbol)
     " DEFAULT #{value_to_string(type, default)}"
+  end
+
+  def default_value(type : JSON::Any.class, default)
+    " DEFAULT '#{default.to_json.gsub(/'/, "''")}'"
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,5 @@ script:
   - scripts/test
 notifications:
   email: false
+addons:
+  postgresql: "9.4"
--- a/db/migrations/20180802180356_create_blobs.cr
+++ b/db/migrations/20180802180356_create_blobs.cr
@@ -1,0 +1,11 @@
+class CreateBlobs::V20180802180356 < Avram::Migrator::Migration::V1
+  def migrate
+    create :blobs do
+      add doc : JSON::Any?, default: { "defa'ult" => "val'ue" }
+    end
+  end
+
+  def rollback
+    drop :blobs
+  end
+end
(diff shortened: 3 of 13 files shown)


paulcsmith:
This looks awesome! Have  tried this in a real app? I’d like to make sure it works as expected before merging it in as I’m not as familiar with JSON and Postgres and I don’t have an app to try it on

tzar:
That's fair enough - when I added this feature, I was just prototyping something out quickly; unfortunately I have been busy with other projects since then so haven't updated it since. In other words, at one point it worked and in my rebases I have relied on the specs :wink:

I'm hoping to get back to that project later this year, so perhaps it would be wise to wait until something has proven it or alternatively set up a simple test app. Cheers

paulcsmith:
Thanks for the heads up and quick response @tzar! I will try to test this out sometime. Even if it doesn’t work in its current state, your work will certainly help move this forward. So thank you! And good luck with your current projects :)

jwoertink:
Ok, I have an app I need this on, so I'm gonna give it a shot and see how it goes. I'll report back after I've had a chance to try it out

src/avram/migrator/column_default_helpers.cr
```diff
@@ -1,5 +1,5 @@
 module Avram::Migrator::ColumnDefaultHelpers
-  alias ColumnDefaultType = String | Time | Int32 | Int64 | Float32 | Float64 | Bool | Symbol
+  alias ColumnDefaultType = Object
```
jwoertink: Hmm.. Not sure about this. It is technically cleaner, but one goal of Avram is to be VERY typesafe, and this would technically allow something like `Tuple` or `Point` to match because `Tuple.is_a?(Object)`.  @paulcsmith what are your thoughts on this?

jwoertink:
Ok, I just finished going over this. For the most part it's pretty good. One thing I'm having issues with is after saving the record, pulling the data back out is a little strange.

```crystal
p = PreferenceQuery.new.first

# This fails with no method overload
PreferenceForm.update!(p, notifications: {"mobile" => true})

# This also fails with no method overload
PreferenceForm.update!(p, notifications: {"mobile" => true}.to_json)

# This saves and updates the record
PreferenceForm.update!(p, notifications: JSON::Any.new({"mobile" => true}.to_json))

p = PreferenceQuery.new.first
p.notifications
# => "{\"mobile\":true}"
p.notifications["mobile"]
# Unhandled exception: Expected Hash for #[](key : String), not String (Exception)
#   from JSON::Any#[]<String>:JSON::Any
#   from __icr_exec__:JSON::Any
#   from __crystal_main
#   from Crystal::main_user_code<Int32, Pointer(Pointer(UInt8))>:Nil
#   from Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32
 #  from main
p.notifications.class
# => JSON::Any
JSON.parse(p.notifications.as_s)
{"mobile" => true}

p = PreferenceQuery.new.first
# This stores it as you'd actually expect 
PreferenceForm.update!(p, notifications: JSON::Any.new({"mobile" => JSON::Any.new(true)}))
p.notifications
# => {"mobile" => true}
p.notifications["mobile"]
# => true
```

So as you can see, saving the field does technically work, but you can save a single string as JSON which means that this could lead to some error with data. If you're storing a ton of data as a string and don't realize until later, this could cause a huge issue. 

I think it would be nice to have a few subtypes from `JSON::Any` to specify something like a field is very specifically an `Array(String)` that gets checked and mapped for json, or this field is very specifically `Hash(String, String)`, for example. What that looks like, I'm not totally sure yet....

paulcsmith:
Closing in favor of https://github.com/luckyframework/avram/pull/108

Thank you so much for your work on this @tzar! We'e making a few small tweaks and then this will be in! 🎉

tzar:
awesome 👍

PR #57, opened 2019-02-26
Log when form fails to save

### LINKED ISSUE #58: Should failed validations log in development?
If I create a form based off of a model, but don’t include all of the required fields in both the html form and the form object, all created will fail silently. 

Perhaps logging failed validations would make this less confusing to developers?

paulcsmith:
Closes #58

### DIFF
--- a/config/database.cr
+++ b/config/database.cr
@@ -1,5 +1,19 @@
 database_name = "avram_dev"
 
+# Adding this struct makes everything work
+#
+# Removing it causes:
+#
+# BUG: trying to upcast Dexter::Formatters::JsonLogFormatter <- Dexter::Formatters::BaseLogFormatter (Exception)
+#   from ???
+#   from ???
+#   from ???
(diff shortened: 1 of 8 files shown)


config/database.cr
```diff
@@ -1,5 +1,19 @@
 database_name = "avram_dev"
 
+# Adding this struct makes everything work
+#
+# Removing it causes:
+#
+# BUG: trying to upcast Dexter::Formatters::JsonLogFormatter <- Dexter::Formatters::BaseLogFormatter (Exception)
+#   from ???
+#   from ???
+#   from ???
```
paulcsmith: I'd like to figure this out before merging

paulcsmith: Actually, this is a super handy feature and I'd like to release it soon. So I'll merge with this hack in place and add an issue to remove it when the bug in Crystal is fixed https://github.com/crystal-lang/crystal/issues/7481

PR #62, opened 2019-02-28
Fix preload bug when loading results twice

### LINKED ISSUE #51: bind message supplies 2 parameters, but prepared statement "" requires 1
I'm having trouble tracking this bug down. Here's a `crystal play` session you can run, as long as you have a database with two tables `users` and `follows`:

```
users   (id, created_at, updated_at)
follows (id, created_at, updated_at, from_id)
```

```crystal
require "avram"

Avram::Repo.configure do |settings|
  settings.url = "postgres://localhost/linkybits_development"
end

class User < Avram::Model
  table users do
  end
end

class Follow < Avram::Model
  table follows do
    belongs_to from : User
  end
end

user = User::BaseQuery.new.first
follows = Follow::BaseQuery.new.preload_from

if !follows.size.zero?
  follows.each do |follows|
    # anything
  end
end
```

The error is just above `# anything`:

> bind message supplies 2 parameters, but prepared statement "" requires 1

Any idea?

paulcsmith:
Closes #51

Here is the problem. We were adding the `in(ids)` every time a preload proc
was called. So the second time `results` was called, it would add
another identical clause. This was a no-no.

Instead we duplicate the query so we can get a fresh copy of the query
that we use just for preloading w/o modifying the original query.
93265c5

cc @edwardloveall


paulcsmith:
More info:

```
# First time it has one where
query.args # => @wheres # => [#<Avram::Where::In:0x7fa5e7aa4180 @column="managers.id", @value=[]>, #<Avram::Where::In:0x7fa5e7a46c90 @column="managers.id", @value=[]>]
[[], []]
# Second time it has a duplicate where::in
@wheres # => [#<Avram::Where::In:0x7fa5e7aa4180 @column="managers.id", @value=[]>, #<Avram::Where::In:0x7fa5e7a46c90 @column="managers.id", @value=[]>]
```

edwardloveall:
Thanks for looking into this! I'm a little confused by your output above https://github.com/luckyframework/avram/pull/62#issuecomment-468100132 It looks like both outputs have duplicate `Where::In` statements, no?

paulcsmith:
I trimmed it down and might have trimmed the wrong stuff :P I promise one of them had just one lol

paulcsmith:
Narrowed it down a tad bit more. The problem only happens when loading results twice (which happens when using `each` or any enumerable method) and only when preloading. If I comment out the preload line it works fine. Trying to figure out what in the preload part of the code is mutating the QueryBuilder... 🤔

src/avram/associations.cr
```diff
@@ -165,7 +165,7 @@ module Avram::Associations
       def preload(preload_query : {{ model }}::BaseQuery)
         add_preload do |records|
           ids = records.map(&.id)
-          {{ assoc_name }} = preload_query.{{ foreign_key }}.in(ids).results.group_by(&.{{ foreign_key }})
+          {{ assoc_name }} = preload_query.dup.{{ foreign_key }}.in(ids).results.group_by(&.{{ foreign_key }})
```
paulcsmith: Here is the problem. We were add the `in(ids)` every time a preload proc was called. So the second time `results` was called, it would add another identical clause. This was a no-no.

Instead we duplicate the query so we can get a fresh copy of the query that we use just for preloading w/o modifying the original query.

paulcsmith:
Gonna do a patch version bump too this is included in the new release of Lucky that went out a couple hours ago

PR #64, opened 2019-02-28
Another failing spec for preload

paulcsmith:


matthewmcgarvey:
Is the problem the spec or does the commented code cause problems? If it is the spec, then it passes now.

paulcsmith:
Oh nice! The commented out part was me messing around briefly. Pushing up a version without comments and just the spec. Will merge once it is green

PR #71, opened 2019-03-15
Add `VirtualForm#errors`

### LINKED ISSUE #29: Add VirtualForm#errors method similar to the one available for regular forms

HarrisonB:
Closes #29.

### DIFF
--- a/src/avram/form_errors.cr
+++ b/src/avram/form_errors.cr
@@ -1,5 +1,5 @@
 module Avram::FormErrors
-  def errors
+  def errors : Hash(Symbol, Array(String))
     fields.reduce({} of Symbol => Array(String)) do |errors_hash, field|
       if field.errors.empty?
         errors_hash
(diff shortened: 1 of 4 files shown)


src/avram/form_errors.cr
```diff
@@ -0,0 +1,12 @@
...
+  def errors
+    fields.reduce({} of Symbol => Array(String)) do |errors_hash, field|
+      if field.errors.empty?
+        errors_hash
+      else
+        errors_hash[field.name] = field.errors
+        errors_hash
+      end
+    end
+  end
```
paulcsmith: Can you add an `abstract def fields` to this just so it is extra clear what this module requires?

src/avram/form_errors.cr
```diff
@@ -0,0 +1,12 @@
+module Avram::FormErrors
+  def errors
```
paulcsmith: Could you add a return type here? If you're not sure what it is you can make it String and then the compiler will tell you what it actually should be

paulcsmith:
Also looks like it needs to be formatted: `crystal tool format src spec config`

HarrisonB:
Should be all good 😄

PR #82, opened 2019-04-26
Raise pretty error message when connection fails

### LINKED ISSUE #81: When there is an issue connecting, try common username/passwords and suggest them
For example, different distros of Linux have different default usernames and passwords. Would be cool to automatically try those and suggest one if it works

jwoertink:
closes #81 (though, this is an alternative to doing that)

Prior to this, you'd get a `DB::ConnectionRefused error` as well as some `PQ` related errors. Now you get a nice error message that helps to explain what went wrong, and possible ways to fix it.

### DIFF
--- a/spec/connection_spec.cr
+++ b/spec/connection_spec.cr
@@ -1,0 +1,11 @@
+require "./spec_helper"
+
+describe Avram::Connection do
+  it "displays a helpful error when failing to connect" do
+    conn = Avram::Connection.new("postgres://root:root@localhost:5432/tacoman")
+    message = Regex.new("Failed to connect to databse 'tacoman' with username root.")
+    expect_raises(Avram::ConnectionError, message) do
+      conn.try_connection!
+    end
+  end
+end
--- a/spec/connection_spec.cr
+++ b/spec/connection_spec.cr
@@ -3,7 +3,15 @@ require "./spec_helper"
 describe Avram::Connection do
   it "displays a helpful error when failing to connect" do
     conn = Avram::Connection.new("postgres://root:root@localhost:5432/tacoman")
-    message = Regex.new("Failed to connect to databse 'tacoman' with username root.")
+    message = Regex.new("Failed to connect to databse 'tacoman' with username 'root'.")
+    expect_raises(Avram::ConnectionError, message) do
+      conn.try_connection!
+    end
+  end
+
+  it "suggests trying a password when no password supplied and connection fails" do
+    conn = Avram::Connection.new("postgres://root@localhost:5432/tacoman")
+    message = Regex.new("You didn't supply a password, did you mean to?")
     expect_raises(Avram::ConnectionError, message) do
       conn.try_connection!
     end
--- a/src/avram/connection.cr
+++ b/src/avram/connection.cr
@@ -9,11 +9,9 @@ class Avram::Connection
   end
 
   def try_connection!
-    begin
-      DB.open(@connection_string)
-    rescue DB::ConnectionRefused
-      raise ConnectionError.new(connection_uri)
-    end
+    DB.open(@connection_string)
+  rescue DB::ConnectionRefused
+    raise ConnectionError.new(connection_uri)
   end
 
   private def connection_uri
--- a/src/avram/connection.cr
+++ b/src/avram/connection.cr
@@ -1,0 +1,22 @@
+# Handles the connection to the DB.
+class Avram::Connection
+  def self.open(connection_string : String) : DB::Database
+    conn = self.new(connection_string)
+    conn.try_connection!
+  end
+
+  def initialize(@connection_string : String)
+  end
+
+  def try_connection!
+    begin
+      DB.open(@connection_string)
+    rescue DB::ConnectionRefused
+      raise ConnectionError.new(connection_uri)
+    end
+  end
+
+  private def connection_uri
+    URI.parse(@connection_string)
+  end
+end
--- a/src/avram/errors.cr
+++ b/src/avram/errors.cr
@@ -59,4 +59,22 @@ module Avram
       super message
     end
   end
+
+  class ConnectionError < AvramError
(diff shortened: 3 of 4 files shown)


spec/connection_spec.cr
```diff
@@ -0,0 +1,11 @@
+require "./spec_helper"
+
+describe Avram::Connection do
+  it "displays a helpful error when failing to connect" do
+    conn = Avram::Connection.new("postgres://root:root@localhost:5432/tacoman")
+    message = Regex.new("Failed to connect to databse 'tacoman' with username root.")
```
paulcsmith: I believe you can pass the string and it will do a substring match. I'm not 100% sure though but worth a try

src/avram/connection.cr
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(@connection_string : String)
+  end
+
+  def try_connection!
+    begin
+      DB.open(@connection_string)
+    rescue DB::ConnectionRefused
+      raise ConnectionError.new(connection_uri)
+    end
```
paulcsmith: You can remove the begin and end and it will still work. This is a handy little shortcut for rescuing errors

src/avram/errors.cr
```diff
@@ -59,4 +59,22 @@ module Avram
...
+
+  class ConnectionError < AvramError
+    def initialize(connection_details : URI)
+      error = String.build do |message|
+        message << "Failed to connect to databse '#{connection_details.path.try(&.[1..-1])}' with username #{connection_details.user}.\n"
+        message << "You can try these steps to help:"
+        message << '\n'
+        message << '\n'
+        message << "  ▸ Ensure you have the correct settings in config/database.cr\n"
+        message << "  ▸ Does the database exist? If not, be sure to create it.\n"
```
paulcsmith: Might be worth including the command `lucky db.create`

src/avram/errors.cr
```diff
@@ -59,4 +59,22 @@ module Avram
...
+        message << "  ▸ Check that you have access to connect to #{connection_details.host} on port #{connection_details.port}\n"
+        if connection_details.password.blank?
+          message << "  ▸ You didn't supply a password, did you mean to?\n"
```
paulcsmith: Nice 👍

src/avram/errors.cr
```diff
@@ -59,4 +59,22 @@ module Avram
...
   end
+
+  class ConnectionError < AvramError
+    def initialize(connection_details : URI)
+      error = String.build do |message|
+        message << "Failed to connect to databse '#{connection_details.path.try(&.[1..-1])}' with username #{connection_details.user}.\n"
+        message << "You can try these steps to help:"
+        message << '\n'
+        message << '\n'
+        message << "  ▸ Ensure you have the correct settings in config/database.cr\n"
```
paulcsmith: Nice. We can add some comments here in the generated code to help people find a suitable username/password for command Postgres installations

src/avram/errors.cr
```diff
@@ -59,4 +59,22 @@ module Avram
...
     end
   end
+
+  class ConnectionError < AvramError
+    def initialize(connection_details : URI)
+      error = String.build do |message|
+        message << "Failed to connect to databse '#{connection_details.path.try(&.[1..-1])}' with username #{connection_details.user}.\n"
+        message << "You can try these steps to help:"
+        message << '\n'
+        message << '\n'
```
paulcsmith: I think these should be double quotes but I am not sure.

spec/connection_spec.cr
jwoertink: I tried that first and it fails. Seems the string does exact match?

src/avram/errors.cr
jwoertink: They should be single because the carriage return is a single character. It's actually more performant to do it this way though negligible when talking about this small of code...

paulcsmith: TIL!

jwoertink:
Screenshot of what the error looks like:

Screen Shot 2019-04-26 at 10 24 06 AM

src/avram/errors.cr
paulcsmith: Also I think this can be shortened. `Be sure the database exists (lucky db.create)`

src/avram/errors.cr
paulcsmith: People don't read (myself included) so maybe we can shorten this to `Check connection settings in 'config/databse.cr'`

src/avram/errors.cr
```diff
@@ -59,4 +59,22 @@ module Avram
       super message
     end
   end
+
+  class ConnectionError < AvramError
+    def initialize(connection_details : URI)
+      error = String.build do |message|
+        message << "Failed to connect to databse '#{connection_details.path.try(&.[1..-1])}' with username #{connection_details.user}.\n"
```
paulcsmith: Let's wrap the username in `'` as well (like with the database name

src/avram/errors.cr
```diff
@@ -59,4 +59,22 @@ module Avram
...
+        message << "You can try these steps to help:"
```
paulcsmith: In most places we have this as `Try this...`

spec/connection_spec.cr
```diff
@@ -0,0 +1,11 @@
...
+    expect_raises(Avram::ConnectionError, message) do
+      conn.try_connection!
+    end
+  end
```
paulcsmith: Could you add a spec for one without password to check that it displays the password hint?

paulcsmith:
Also, I have a feeling you like Tacos...

paulcsmith:
If it isn't too much to ask, could you add some common usernames/passwords to https://github.com/luckyframework/lucky_cli/blob/2cffd3fdd9e288ae20cd06cba8cd82bb12ff7e8c/src/web_app_skeleton/config/database.cr.ecr#L7 ?

PR #86, opened 2019-05-12
Add sequence to Avram::Box

edwardloveall:
_Note: just adding this because I ran into it while developing and it was about as easy to impliment it as it was to make an issue. Feel free to push back 🙂_

Unique values are a good thing and application developers should use
them. However, testing with unique values can be a pain since you need
to create a new value explicitly in tests instead of using a default
value via a Box.

This makes sequences (some code that I found in the Avram tests)
included by default in every box. This seems like a thing that users
shouldn't have to build into every app themselves.

### DIFF
--- a/src/avram/box.cr
+++ b/src/avram/box.cr
@@ -1,6 +1,8 @@
 abstract class Avram::Box
   getter form
 
+  SEQUENCES = {} of String => Int32
+
   macro inherited
     {% unless @type.abstract? %}
       {% form = @type.name.gsub(/Box/, "::BaseForm").id %}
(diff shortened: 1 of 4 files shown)


src/avram/box.cr
```diff
@@ -37,4 +39,22 @@ abstract class Avram::Box
   def self.create_pair
     2.times { new.create }
   end
+
+  # Returns a value with a number to use for unique database values.
```
paulcsmith: I think you can remove “database” since the values are unique whether used for database values or not

edwardloveall: Oh huh, I thought it was database constraint-based. 👍

PR #90, opened 2019-06-03
Updates for crystal 0.29.0

bcardiff:
There a couple of relative paths used in `require`s that needs to be fixed for Crystal 0.29.0


bcardiff:
FYI The build is failing because it uses crystal 0.27.2 .
It should work fine in 0.28.0.

Feel free to close/adapt this PR accordingly.

NB: I think next ameba release will only work in 0.29.0 and will not have support for 0.28.0, although ameba is not used in avram, but in other lucky repos

PR #108, opened 2019-06-15
JSON Support 2

jwoertink:
This replaces #48 with some updates 
* ensure this runs on crystal 0.29.0
* fixes a regression where setting value of anything other than String was ~10x slower
* Puts a type restriction back in place on `ColumnDefaultType`.

One thing to note about this type restriction is that in pr 48 it was set to `Object`, this meant you could set default values on json columns super easy like `default: {"number" => 1}`. That was nice, but it also meant you could set pretty much anything there like `default: HTTP::Params` since as long as it was of type `Object`, it would work. 

Eventually I would like to open this up, but in keeping with the type safe tradition, we can start with a rough version, and then slowly improve it over time.


.gitignore
```diff
@@ -1,7 +1,7 @@
 /docs/
 /lib/
 /bin/lucky
-/bin/gen.migration*
+/bin/lucky.gen.migration*
```
paulcsmith: Nice catch!

spec/json_column_spec.cr
```diff
@@ -0,0 +1,62 @@
+require "./spec_helper.cr"
+
+class BlobQuery < Blob::BaseQuery
+end
+
+class BlobForm < Blob::BaseForm
+  fillable :doc
```
paulcsmith: This will need to be `permit_columns` once rebased. And the `BaseForm` is now `SaveOperation`. Maybe it should be `BaseOperation` though? Anyway we can talk about that since it'll be awhile before this big release comes out I think :D

spec/json_column_spec.cr
```diff
@@ -0,0 +1,62 @@
...
+describe "JSON Columns" do
+  it "should work in boxes" do
+    BlobBox.create
+    blob = BlobQuery.new.first
+    blob.doc.should eq JSON::Any.new({"foo" => JSON::Any.new("bar")})
+
+    blob2 = BlobBox.new.doc(JSON::Any.new(42_i64)).create
+    blob2.doc.should eq JSON::Any.new(42_i64)
+  end
+  it "should be nullable" do
```
paulcsmith: Could you add some newlines between `it` blocks

src/avram/form.cr
```diff
@@ -234,16 +234,25 @@ abstract class Avram::Form(T)
...
 
+  private def cast_value(value : Nil)
+    nil
+  end
+
+  private def cast_value(value : Object)
+    case value
+    when JSON::Any
+      value.to_json
+    else
```
paulcsmith: What else could this object be? I wonder if we should be explicit about what this object type is?

src/avram/migrator/column_default_helpers.cr
```diff
@@ -1,5 +1,5 @@
 module Avram::Migrator::ColumnDefaultHelpers
-  alias ColumnDefaultType = String | Time | Int32 | Int64 | Float32 | Float64 | Bool | Symbol
+  alias ColumnDefaultType = String | Time | Int32 | Int64 | Float32 | Float64 | Bool | Symbol | JSON::Any
```
paulcsmith: Nice. Much better than Object 👍

src/avram/query_builder.cr
```diff
@@ -39,10 +39,13 @@ class Avram::QueryBuilder
...
   private def param_values(params)
     params.values.map do |value|
-      if value.nil?
+      case value
+      when Nil
         nil
-      else
+      when String, UUID
         value.to_s
+      else
```
paulcsmith: I'm nervous about this because what if in the future we have different types of values? This might be hard to debug in the future. Could we do the opposite like `when JSON::Any -> value.to_json` and the else does `to_s`?

I'm really not sure here since it's been awhile since I looked at this code 🤷‍♂️

src/avram/form.cr
jwoertink: This can be any object that the form supports. Before it was just doing

```
if field.value.nil?
  nil
else
  field.value.to_s
end
```

This just adds in the catch for `JSON::Any` to call `to_json` instead of `to_s`, but otherwise works exactly the same as before.

paulcsmith: Gotcha! Thanks for clarifying. It's been awhile since I wrote this lol

src/avram/query_builder.cr
jwoertink: Yup. I didn't catch this one. Should be same as in the form cast value.

spec/json_column_spec.cr
jwoertink: `BaseOperation` sounds cool like it's all secret military or something. Do you want me to merge master in to this to make sure it's all working with latest stuff first? Or just wait until after this is merged to update?

paulcsmith: I think rebasing master would be good to make sure it all works ok. And yeah `BaseOperation` does sound kinda cool haha

paulcsmith:
I don't get notified for updates/force pushes so can you ping me when you're ready for me to take another look?

jwoertink:
Weird. Yeah, I'm getting ready to work on this now, so it'll be here in a bit.

jwoertink:
@paulcsmith this adds a simple example on doing a where query with the json. I'll do a new PR to add in fancy query with json.

paulcsmith:
Looks great! Just needs a rebase and then I'll merge it in!

jwoertink:
@paulcsmith nothing to rebase.

paulcsmith:
Hmm looks like there is because I cannot merge 
Screen Shot 2019-06-19 at 1 14 29 PM

Maybe you need to update master first?

jwoertink:
Screen Shot 2019-06-19 at 10 17 25 AM


```
[10:11AM] avram (pr48)$ git checkout master
Switched to branch 'master'
Your branch is ahead of 'origin/master' by 13 commits.
  (use "git push" to publish your local commits)
[10:12AM] avram (master)$ git fetch upstream
[10:12AM] avram (master)$ git merge upstream/master
Already up to date.
[10:12AM] avram (master)$ git checkout pr48
Switched to branch 'pr48'
[10:12AM] avram (pr48)$ git merge master
Already up to date.
```

I'm not great at git. Maybe I'm missing something?

paulcsmith:
Ok...maybe GitHub is having issues lol. Because I still see that message that there are conflicts. Can you go ahead and merge it since I can't 🤣

PR #114, opened 2019-06-23
Allow customizing the setup step in models

### LINKED ISSUE #44: Allow customizing modules and inheritance
```crystal
class BaseModel < LuckyRecord::Model
  # Will make it so models use these base classes instead of the default lucky ones
  query_inherits_from BaseQuery
  form_inherits_from BaseForm 

  include_in_query Pagination
  include_in_form SomeCustomValidator
end

class User < BaseModel
end
```

paulcsmith:
Closes #44


paulcsmith:
This will make it a lot easier for other libs to hook into the model setup process since internally that's what Avram now does! Just register a macro

PR #118, opened 2019-06-24
Ensure proper ordering in the `first?` method

### LINKED ISSUE #115: last? orderes records by ID but first? does not
Here's the relevant source for `last?` and `first?`

```crystal
  def first?
    query.limit(1)
    results.first?
  end

  def last?
    ordered_query.reverse_order.limit(1)
    results.first?
  end
```

The important difference is `ordred_query` which ads the ordering by `id` if no ordering exists:

```crystal
  private def ordered_query
    if query.ordered?
      query
    else
      query.order_by(:id, :asc)
    end
  end
```

I think last and first should order the same way to ensure consistent results. The fix would be to add `ordered_query` to the `first?` method.

smadeja:
Make it match the ordering that the `last?` method uses. Fixes #115.

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -101,9 +101,10 @@ describe Avram::Query do
       UserBox.new.name("First").create
       UserBox.new.name("Last").create
 
-      user = UserQuery.new.first?
+      user = (user_query = UserQuery.new).first?
       user.should_not be_nil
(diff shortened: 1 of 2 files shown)


paulcsmith:
🎉

jwoertink:
Thanks for adding this in @smadeja!

PR #119, opened 2019-06-27
Refactor adding columns

paulcsmith:
Previously to add a column we had to add a bunch of methods and method
overloads to multiple modules. This made things hard to follow and
error-prone.

Now we have classes that handle adding columns. It makes it easy to add
new columns and even third parties can add new columns by making a new
class. I think there are still some changes we can make to make this
easier, but this is a step in a good direction IMO.
cb6af61


paulcsmith:
I still want to remove the `ColumnDefaultHelpers` and rework the `PrimaryKeyType` enum and values

.gitignore
```diff
@@ -1,7 +1,6 @@
 /docs/
 /lib/
 /bin/*
-/db/migrations/
```
paulcsmith: I'm not sure why this was ignored since it is needed for script/test

spec/migrator/alter_table_statement_spec.cr
```diff
@@ -19,25 +19,25 @@ describe Avram::Migrator::AlterTableStatement do
...
+      ADD email text NOT NULL,
       ADD nickname text NOT NULL,
       ADD age int NOT NULL DEFAULT 1,
       ADD num bigint NOT NULL DEFAULT 1,
       ADD amount_paid decimal(10,5) NOT NULL DEFAULT 1.0,
       ADD completed boolean NOT NULL DEFAULT false,
       ADD meta jsonb NOT NULL DEFAULT '{"default":"value"}',
       ADD joined_at timestamptz NOT NULL DEFAULT NOW(),
-      ADD updated_at timestamptz,
+      ADD updated_at timestamptz NOT NULL,
```
paulcsmith: This refactor actually caught a bug in the alter statement. These should have been NOT NULL from the start

spec/migrator/alter_table_statement_spec.cr
```diff
@@ -53,7 +53,7 @@ describe Avram::Migrator::AlterTableStatement do
         add_belongs_to post : Post?, on_delete: :restrict
         add_belongs_to category_label : CategoryLabel, on_delete: :nullify, references: :custom_table
         add_belongs_to employee : User, on_delete: :cascade
-        add_belongs_to line_item : LineItem, on_delete: :cascade, foreign_key_type: Avram::Migrator::PrimaryKeyType::UUID
+        add_belongs_to line_item : LineItem, on_delete: :cascade, foreign_key_type: UUID
```
paulcsmith: New shorter syntax here

paulcsmith:
@jwoertink Mind taking a look at this?

src/avram/migrator/columns/add_string.cr
```diff
@@ -0,0 +1,14 @@
...
+    @default : String? = nil
+
+    def initialize(@name, @nilable, @default)
+    end
+
+    def column_type
+      "text"
+    end
+  end
+end
```
jwoertink: Should this class support a length option?

spec/migrator/alter_table_statement_spec.cr
```diff
@@ -19,25 +19,25 @@ describe Avram::Migrator::AlterTableStatement do
...
     ALTER TABLE users
       ADD name text,
-      ADD email text,
+      ADD email text NOT NULL,
       ADD nickname text NOT NULL,
-      ADD age int NOT NULL DEFAULT 1,
-      ADD num bigint NOT NULL DEFAULT 1,
-      ADD amount_paid decimal(10,5) NOT NULL DEFAULT 1.0,
-      ADD completed boolean NOT NULL DEFAULT false,
+      ADD age int NOT NULL DEFAULT '1',
```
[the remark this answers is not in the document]
paulcsmith: It does :D I added the `TestDefaults` migration just to be sure

src/avram/migrator/columns/add_string.cr
paulcsmith: It could, but I think I'd rather keep this PR purely refactoring and not  adding new stuff. I'm also not sure it should be added since `text` columns can be any length. But I dunno. Probably wouldn't be too bad to  add later

jwoertink: True. I definitely think it should be added, but we can go through all that stuff and start adding some fancy add-ons later.

paulcsmith: That sounds fair to me! Luckily this refactor should make this stuff a lot easier to do! I'm sure we could add some time options and stuff too

src/avram/migrator/columns/json_column.cr
```diff
@@ -0,0 +1,18 @@
+require "./base"
+
+module Avram::Migrator::Columns::JSON
+  class AnyColumn < Base
```
jwoertink: you clever guy you. It took me a sec to figure out how this one would work.

paulcsmith: Hehe thanks. Makes me wonder if we should instead use just `JSON` as the column? `add meta : JSON`?

jwoertink: We'd end up having to do some mutation stuff with saving and pulling data out most likely. Since things will return `JSON::Any`. It would be nice if it was shorter, but that could lead to some other issues.

paulcsmith: That makes sense! Let’s keep it as is then. But maybe we can add a JSONColumn class that if used raises an error telling you to use JSON::Any just in case people forget

Issue #89, opened 2019-05-31
Default primary key to Int64 instead of Int32


paulcsmith:
https://github.com/luckyframework/avram/issues/87#issuecomment-497837244

jwoertink:
Just to add to this, right now the tables created generate a `SERIAL` for Int32. With Int64, they should generate `BIGSERIAL`. This should also be configurable to use either (as well as other types), but still default to Int64.

paulcsmith:
Good point! BTW could you try this out? I'm curious if setting Int64 to default will still work with old Int32 columns. 

The reason I ask is I wonder if we can safely change the default types in Crystal to Int64 and still work with old columns created as Int32. I *think* it should work, in which case this will be super easy to make the default. I just haven't had time to try this out

jwoertink:
Yup. I'll give that a shot and report back.

jwoertink:
There's quite a few spots that mention `Int32`. I tried to change the ones I thought were relevant to `Int64`, but  it seems PG didn't like that:

```
icr(0.28.0) > u = UserQuery.new.first
Unhandled exception: PG::ResultSet#read returned a Int32. A Int64 was expected. (Exception)
  from PG::ResultSet@DB::ResultSet#read<Int64.class>:Int64
  from User#initialize<PG::ResultSet>:Int32
  from User::new<PG::ResultSet>:User
  from User::from_rs<PG::ResultSet>:Array(User)
  from UserQuery@Avram::Queryable(T)#exec_query:Array(User)
  from UserQuery@Avram::Queryable(T)#results:Array(User)
  from UserQuery@Avram::Queryable(T)#first?:(User | Nil)
  from UserQuery@Avram::Queryable(T)#first:User
  from __icr_exec__:User
  from __crystal_main
  from Crystal::main_user_code<Int32, Pointer(Pointer(UInt8))>:Nil
  from Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32
  from main
```

I think adding in the BIGSERIAL as Int64 may still work without borking current setup DBs. My guess is PG would see SERIAL, and maybe the crystal-db would know to map this to Int32 where if it saw BIGSERIAL, it could map to Int64.

paulcsmith:
Yeah maybe that could work. Another alternative would be to build in a method or figure out some SQL so people can easily switch to Int64/bigserial.

Issue #96, opened 2019-06-10
Custom Primary Key support


jwoertink:
Ref: https://gitter.im/luckyframework/Lobby?at=5cfedb153dcdab400324037a

Related: 
https://github.com/luckyframework/avram/issues/6
https://github.com/luckyframework/avram/issues/21
https://github.com/luckyframework/avram/issues/87
https://github.com/luckyframework/avram/issues/89

The current support for primary keys in Avram is limited to `UUID` UUID and `SERIAL` Int32. Other options requested are:

* [x] - Int32 (`id SERIAL`)
* [x] - UUID (`id UUID`)
* [x] - Int64 (`id BIGSERIAL`) #121 
* [ ] - String (`id character varying(18) NOT NULL`) or something like (KSUID)
* [ ] - Composite (`PRIMARY KEY (id_1, id_2)`)
* [ ] - Int16 (`id SMALLSERIAL`)
* [ ] - no primary key (table with no `id` field.)
* [x] - renaming the primary_key column to something else.

Since several people request this, and there's already quite a few issues open about it, I wanted to put them all in to 1 location to track. If you have a need for a custom primary key, let us know what your use case is, and if the current support is blocking or not for you. This helps us to prioritize and/or gauge what is wanted.

paulcsmith:
Great idea putting this all in one spot 👍

watzon:
It would be really nice to have the `no` option and then have a `primary` option on the `add` method. For instance:

```crystal
class CreateUsers::V20190610161436 < Avram::Migrator::Migration::V1
  def migrate
    create :users, :none do
      add id : Int64, primary: true
    end
  end

  def rollback
    drop :users
  end
end
```
Mainly I want it to not automatically generate id's, but still allow the creation of a primary key without using a custom SQL query.

jwoertink:
Adding in my own personal use cases:

I have an app that needed to migrate data from an old database in mysql with int32, to a new postgres app. We had to differentiate between the two by using `id character varying(18) NOT NULL` for all of our new records, and then had a secondary column that stored the Int for the old DB. KSUID would have worked better had we known what that was at the time.

Some of the join tables in this rails app have no primary key at all. So the current setup throws an error when trying to fetch an `id` column that doesn't exist. We could just add an id column, but a SQL dump of this DB is 62GB, and this would just increase the size for no reason. 

So supporting custom String, and no id are both options I would like to see.

paulcsmith:
[...]

PR #121, opened 2019-06-30
Customize defaults and primary key

### LINKED ISSUE #87: Allow disabling default columns
It should be possible to disable default columns (`id`, `created_at`, and `updated_at`), both in the migrations and the models.

As a reference, the way Laravel does this is:
* adding id and timestamps to a table in a migration is explicit (but included by default in an `artisan`-generated migration),
* disabling timestamps in the model is possible through a field (`public $timestamps = false;`),
* changing/deleting the primary key in the model is also possible through a field (`protected $primaryKey = "column_name"`/`null;`).

Related: https://github.com/luckyframework/avram/issues/6

paulcsmith:
Part of #96 #121. Still needs to work with different primary key column name on the model side

Closes https://github.com/luckyframework/avram/issues/89
Closes #87

While you can add different name for the primary key it currently will not work in the model. That will require a lot more work. This is already rather large though so I think this is still a big step forward. Next PR can handle allowing a differently named primary key


jwoertink:
This is gonna be huge! I looked it over briefly, and it looks great so far!

paulcsmith:
This also gets us close to no longer needing explicit requires in models. There is just one more `model.resolve` in there, but if I can figure out a way to remove that then we won't need the requires!

src/avram/save_operation.cr
```diff
@@ -153,7 +153,7 @@ abstract class Avram::SaveOperation(T)
     def after_prepare
       validate_required *required_attributes
 
-      {% if primary_key_type == :uuid %}
+      {% if primary_key_type.id == UUID.id %}
         id.value ||= UUID.random()
```
jwoertink: I wonder if this should be handled on the postgres side?

src/avram/migrator/columns/primary_keys/base.cr
```diff
@@ -0,0 +1,11 @@
+abstract class Avram::Migrator::Columns::PrimaryKeys::Base
+  macro inherited
+    private getter name : String
+  end
+
+  abstract def column_type
+
+  def build : String
+    %(  #{name} #{column_type} PRIMARY KEY)
```
jwoertink: What are your thoughts on adding an extra method for default value here?

```crystal
%(  #{name} #{column_type} PRIMARY KEY #{default_values})
```

In the case of Int32 and Int64 being serial and bigserial, their methods would define nothing. But in the case of UUID, it would define 

```crystal
def default_values
  "DEFAULT uuid_generate_v4()"
end
```

One thing I'm also thinking about is later in another PR may be the ability to just define a String as primary key, or possibly composite keys. We will need a way to tell it where it gets it's default value from.

jwoertink:
Ok, this all looks good to me. I added some comments, but nothing I see required really. Feel free to merge if you think it's ready!

src/avram/save_operation.cr
paulcsmith: I kinda think so, but it does require an extension to be enabled. Maybe that is fine? I'd much rather have it on the Postgres side if possible. I'll open another issue for it though

paulcsmith: https://github.com/luckyframework/avram/issues/122

src/avram/migrator/columns/primary_keys/base.cr
paulcsmith: I think in those cases they can override the `build` method so they can do whatever they want. But I agree it should do this by default for uuid. I added an issue here https://github.com/luckyframework/avram/issues/122 to tackle that.

paulcsmith:
Thanks for the review!

lesderid:
Thanks a lot for the work, this PR is a great improvement!

paulcsmith:
Thanks for the kind words and the suggestions that helped prompt this :) @lesderid

PR #123, opened 2019-07-01
Adds new table_for macro

### LINKED ISSUE #120: Add table_for(Model) to migrations
This will return the table name. e.g. `table_for(User)` -> `:users`

Which can then be used in migrations and the migration generator

`create table_for(User) do/end`. This would work in tandem with the `table` inference in models so they always match up correctly

jwoertink:
fixes #120 

Currently I'm not able to run specs because I get an exception

```
Running specs

Error in line 24: while requiring "./spec/query_spec.cr"

in spec/query_spec.cr:480: undefined method 'join_posts' for Comment::BaseQuery

      query = Comment::BaseQuery.new.join_posts
                                     ^~~~~~~~~~
```

This may be related to previous commits, but I'm not sure.


paulcsmith:
Could you open another PR that uses the old run macro instead of `table_for` in the models? That way we can merge the Lucky PR and work on this later

jwoertink:
The model stuff can be added in a separate PR later

PR #124, opened 2019-07-02
Handle renamed primary key

paulcsmith:
You can use a different primary key than `id` now. It is still highly recommended to use `id` for all new tables, but if you have legacy tables this can be a handy option


db/migrations/20171006163153_create_posts_and_comments.cr
```diff
@@ -1,13 +1,13 @@
...
     create :posts do
-      primary_key id : Int64
+      primary_key custom_id : Int64
       add_timestamps
       add title : String
     end
 
     create :comments do
-      primary_key id : Int64
+      primary_key custom_id : Int64
```
paulcsmith: Using custom keys in a number of relationships to make sure everything works fine

db/migrations/20180802180358_different_defaults.cr
```diff
@@ -0,0 +1,14 @@
+class DifferentDefaults::V20180802180358 < Avram::Migrator::Migration::V1
+  def migrate
+    create :table_with_different_default_columns do
```
paulcsmith: A specific table for testing find, delete, and timestamps

script/test
```diff
@@ -8,25 +8,31 @@ docker-compose up -d
...
-
-printf "\nChecking that tasks build correctly\n\n"
-$COMPOSE shards build
-
-printf "\nChecking code formatting\n\n"
-if ! $COMPOSE crystal tool format --check src spec config > /dev/null; then
-    printf "\nCode is not formatted.\n"
-    printf "\nFormat the code with: docker-compose run --rm app crystal tool format src spec config\n\n"
-    exit 1
+$COMPOSE crystal spec "$@"
```
paulcsmith: You can now pass a path to `script/test` and have it run just those tests

spec/query_spec.cr
```diff
@@ -104,7 +104,7 @@ describe Avram::Query do
       user = (user_query = UserQuery.new).first?
       user.should_not be_nil
       user.not_nil!.name.should eq "First"
-      user_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users ORDER BY id ASC LIMIT 1"
+      user_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users ORDER BY users.id ASC LIMIT 1"
```
paulcsmith: This is more accurate. Referencing the table means it is less likely to get confused with a different table's id column

src/avram/base_query_template.cr
```diff
@@ -1,11 +1,23 @@
...
     class ::{{ type }}::BaseQuery < Avram::Query
       include Avram::Queryable({{ type }})
 
       @@table_name = :{{ table_name }}
       @@schema_class = {{ type }}
 
+      # If not using default 'id' primary key
+      {% if primary_key_name.id != "id".id %}
+        # Then point 'id' to the primary key
+        def id(*args, **named_args)
```
paulcsmith: There is an `id` method defined so that things depending on it still work as expected

src/avram/join.cr
```diff
@@ -4,7 +4,14 @@ module Avram::Join
...
 
-    def initialize(@from : Symbol, @to : Symbol, @primary_key : Symbol? = nil, @foreign_key : Symbol? = nil, @comparison : String? = "=", @using : Array(Symbol) = [] of Symbol)
+    def initialize(
+      @from : Symbol,
+      @to : Symbol,
+      @primary_key : Symbol | String | Nil = nil,
+      @foreign_key : Symbol | String | Nil = nil,
+      @comparison : String? = "=",
+      @using : Array(Symbol) = [] of Symbol
+    )
```
paulcsmith: A little reformatting

src/avram/model.cr
```diff
@@ -44,7 +44,7 @@ class Avram::Model
     end
   end
 
-  def_equals @id
+  def_equals id
```
paulcsmith: Use the method since the instance var is not defined if using a different primary key name

src/avram/queryable.cr
```diff
@@ -151,7 +153,7 @@ module Avram::Queryable(T)
     if query.ordered?
       query
     else
-      query.order_by(:id, :asc)
+      id.asc_order.query
```
paulcsmith: Type safe, and also adds the table name when ordering

src/avram/join.cr
```diff
@@ -4,7 +4,14 @@ module Avram::Join
   abstract class SqlClause
     getter :from, :to, :from_column, :to_column
 
-    def initialize(@from : Symbol, @to : Symbol, @primary_key : Symbol? = nil, @foreign_key : Symbol? = nil, @comparison : String? = "=", @using : Array(Symbol) = [] of Symbol)
+    def initialize(
+      @from : Symbol,
+      @to : Symbol,
+      @primary_key : Symbol | String | Nil = nil,
```
jwoertink: Not sure if this is related, but in the model you set the `primary_key_name` to return Symbol specifically; should we look at typing this to `Symbol?` instead?

src/avram/queryable.cr
jwoertink: nice!

jwoertink:
This is awesome! This makes Avram so much more versatile.

src/avram/join.cr
paulcsmith: Yeah that makes sense!

PR #127, opened 2019-07-02
Adds new table_for macro 2

jwoertink:
supersedes #123 

This skips adding to the model due to some issues as shown in the other PR.


src/avram/tasks/gen/migration.ecr
```diff
@@ -3,7 +3,10 @@ class <%= migration_class_name %> < Avram::Migrator::Migration::V1
     <%- if formatted_migrate_contents -%>
     <%= formatted_migrate_contents %>
     <%- else -%>
-    # create :things do
+    # Read more on migrations
+    # https://www.luckyframework.org/guides/database/migrations
```
paulcsmith: I'm not sure if this will get too busy after time, but I think it's worth adding and seeing how it does! Maybe we have a "no comments" config option in the future that people can use once they're comfortable

jwoertink: Yeah. I don't like when it gets too busy, but I wanted to make sure it was clear that the second you open the file and you're like "ok, now what?" you at least have a direction to go... I have an idea though....

paulcsmith: I kind of like `.luckyconfig` more than .rc but yeah I think that is an awesome idea

paulcsmith:
This might be another one of those "I broke GitHub moments" but it says it still needs to be rebased :P

Screen Shot 2019-07-02 at 2 50 32 PM

Once CI passes and if you can merge, feel free to do so. I can't 😂

paulcsmith:
I guess I just have bad luck with GitHub's caching :P

jwoertink:
Yeah, it said it was fine on my end lol

PR #128, opened 2019-07-02
adding in right_join_X and full_join_X type safe query methods

### LINKED ISSUE #116: Right and Full joins
Avram [technically supports](https://github.com/luckyframework/avram/blob/master/src/avram/join.cr#L45) these, but they implemented with safety.

https://github.com/luckyframework/avram/blob/b32b5a9b53688762e22c063ebad9f858cba636c0/src/avram/base_query_template.cr#L41

jwoertink:
fixes #116

These joins already exist as classes, they were just missing their type-safe helper method equivalents.


jwoertink:
There's these specs https://github.com/luckyframework/avram/blob/master/spec/join_spec.cr#L13-L19 already, but I didn't actually see any specs for any of these methods. I will add all of those in though, we should probably have them anyway 😂

paulcsmith:
Haha yeah thank you. Guess someone forgot to add those before....not sure who 🤐

spec/query_associations_spec.cr
```diff
@@ -21,4 +21,96 @@ describe "Query associations" do
...
+      .post_id(post_with_matching_comment.id)
+      .create
+
+    posts = Post::BaseQuery.new.left_join_comments.comments(&.body.eq("matching"))
+    posts.to_sql[0].should contain "LEFT JOIN"
+    posts.results.should eq([post_with_matching_comment])
+
+    posts = Post::BaseQuery.new.left_join_comments.comments(&.body("matching"))
+    posts.to_sql[0].should contain "LEFT JOIN"
+    posts.results.should eq([post_with_matching_comment])
```
paulcsmith: I'm not sure I see the different between these two assertions. Maybe accidental duplication?

spec/query_associations_spec.cr
```diff
@@ -21,4 +21,96 @@ describe "Query associations" do
...
+    posts.to_sql[0].should contain "INNER JOIN"
+    posts.results.should eq([post_with_matching_comment])
+
+    posts = Post::BaseQuery.new.inner_join_comments.comments(&.body("matching"))
+    posts.to_sql[0].should contain "INNER JOIN"
+    posts.results.should eq([post_with_matching_comment])
+  end
+
+  it "can query associations with left_join specified" do
+    post_with_matching_comment = PostBox.create
```
paulcsmith: Also might be good to have another Post called `post_without_matching_comment` that can be used with the second comment. That way we know it is excluded.

```crystal
CommentBox.new
  .body("not-matching")
  .post_id(post_without_matching_comment)
  .create

# Your current assertion should still pass if the join worked right
posts.results.should eq([post_with_matching_comment])
```

spec/query_associations_spec.cr
paulcsmith: Oh it looks like it is testing `body` and `body.eq`. I think just one should be ok since we have other spots that should catch that. It doesn't apply to joins in particular as it is just a normal criteria method

spec/query_associations_spec.cr
jwoertink: that's already in the spec though 🤔 All of them have that... Or am I missing something?

paulcsmith: Right now there are two comments, but just one post. So the second comment isn't really testing much. By creating another post and making the second comment attached to it you can be sure that the join query is working expected and only returning the post related to the matching comment and not the second post. Does that make sense?

jwoertink: There's 2 posts though... On all of them. Yeah, I'm still a bit confused.

```crystal
  # first post and comment
    post_with_matching_comment = PostBox.create
    matching_comment = CommentBox
      .new
      .body("matching")
      .post_id(post_with_matching_comment.id)
      .create

  # second post and comment
    post_without_matching_comment = PostBox.create
    CommentBox
      .new
      .body("not-matching")
      .post_id(post_with_matching_comment.id)
      .create
```

paulcsmith: You are *totally* right. I didn't see the second post, but also the second comment calls `post_with_matching_comment` in every spec which I think is a mistake. The second post is never used in these. Maybe add a newline after the first comment as well so it is easier to see? Also I guess I need glasses 🤣

jwoertink: hahahah I missed that too. Ok. I'll update that. 😆

jwoertink:
It says it's still waiting on travis, but I can override it. I'll try to do that.

PR #132, opened 2019-07-03
Table name is guessed. No requires required

paulcsmith:
This improves an issue where previously you had to require associations.
Now we don't because we no longer resolve the type in the macro.

The one downside is that if you have a table name that is not the
default pluralized snake case, you will need to provide it in
`belongs_to`. I think this is a great tradeoff since the require issue
is annoying and causes a lot of issues. Also most people use those
defaults so it should be fine.

### DIFF
--- a/spec/support/comment.cr
+++ b/spec/support/comment.cr
@@ -1,5 +1,3 @@
-require "./post"
-
 class Comment < Avram::Model
   skip_default_columns
 
@@ -8,5 +6,18 @@ class Comment < Avram::Model
     timestamps
     column body : String
     belongs_to post : Post
+  end
+end
+
+class CommentForCustomPost < Avram::Model
+  skip_default_columns
+
+  table :comments do
+    primary_key custom_id : Int64
+    timestamps
+    column body : String
+    belongs_to post_with_custom_table : PostWithCustomTable,
+      table: :posts,
+      foreign_key: :post_id
   end
 end
--- a/spec/support/post.cr
+++ b/spec/support/post.cr
@@ -1,5 +1,3 @@
-require "./comment"
-
 class Post < Avram::Model
   skip_default_columns
 
@@ -12,5 +10,20 @@ class Post < Avram::Model
     has_many comments : Comment
     has_many taggings : Tagging
     has_many tags : Tag, through: :taggings
+  end
+end
+
+# This is a regular post, but with a custom table name
+# This is to test that 'belongs_to' can accept a 'table' in the
+# CommentForCustomPost model
+class PostWithCustomTable < Avram::Model
+  skip_default_columns
+
+  table :posts do
+    primary_key custom_id : Int64
+    timestamps
+
+    column title : String
+    column published_at : Time?
   end
 end
--- a/src/avram/associations.cr
+++ b/src/avram/associations.cr
@@ -56,9 +56,8 @@ module Avram::Associations
     define_has_one_base_query({{ assoc_name }}, {{ model }}, {{ foreign_key }})
   end
 
-  macro belongs_to(type_declaration)
+  macro belongs_to(type_declaration, foreign_key = nil, table = nil)
     {% assoc_name = type_declaration.var %}
-    {% foreign_key = "#{assoc_name}_id".id %}
 
     {% if type_declaration.type.is_a?(Union) %}
       {% model = type_declaration.type.types.first %}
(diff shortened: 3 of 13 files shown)


src/avram/base_query_template.cr
```diff
@@ -38,31 +38,31 @@ class Avram::BaseQueryTemplate
       {% end %}
 
       {% for assoc in associations %}
-        def join_{{ assoc[:name] }}
-          inner_join_{{ assoc[:name] }}
+        def join_{{ assoc[:table_name] }}
+          inner_join_{{ assoc[:table_name] }}
```
paulcsmith: Also renamed this so it is clear that assoc name was actually the table name

spec/support/comment.cr
```diff
@@ -10,3 +8,16 @@ class Comment < Avram::Model
     belongs_to post : Post
   end
 end
+
+class CommentForCustomPost < Avram::Model
+  skip_default_columns
+
+  table comments do
```
jwoertink: To keep things [consistent](https://github.com/luckyframework/avram/pull/127), can this be a Symbol? `:comments`

spec/support/post.cr
```diff
@@ -14,3 +12,18 @@ class Post < Avram::Model
...
   end
 end
+
+# This is a regular post, but with a custom table name
+# This is to test that 'belongs_to' can accept a 'parent_table' in the
+# CommentForCustomPost model
+class PostWithCustomTable < Avram::Model
+  skip_default_columns
+
+  table posts do
```
jwoertink:
```suggestion
  table :posts do
```

src/avram/associations.cr
```diff
@@ -68,18 +67,26 @@ module Avram::Associations
       {% nilable = false %}
     {% end %}
 
-    column {{ assoc_name.id }}_id : {{ model }}::PrimaryKeyType{% if nilable %}?{% end %}
+    {% if !foreign_key %}
+      {% foreign_key = "#{assoc_name}_id".id %}
+    {% end %}
+
+    {% if !parent_table %}
+      {% parent_table = run("../run_macros/infer_table_name.cr", model.id) %}
```
jwoertink: This will be a good spot for that `table_for`. Should that go in here? Or wait until we add it in to that model stuff later?

jwoertink:
This is sweet! Gonna make things a lot cleaner.

src/avram/associations.cr
edwardloveall: What's `table_for`?

src/avram/associations.cr
```diff
@@ -56,9 +56,8 @@ module Avram::Associations
     define_has_one_base_query({{ assoc_name }}, {{ model }}, {{ foreign_key }})
   end
 
-  macro belongs_to(type_declaration)
+  macro belongs_to(type_declaration, foreign_key = nil, parent_table = nil)
```
edwardloveall: `parent_table` makes me think there's inheritance going on, but I don't believe that's how this works at all. I was also wondering about `table_for` or `table_name`. Sounds like the former might be reserved for something else?

spec/support/comment.cr
paulcsmith: Yeah for sure!

src/avram/associations.cr
paulcsmith: Eating mostly

paulcsmith: But really, `table_for` doesn't work because we don't have access to it inside the macro interpolation. That's why I think it wasn't working on the other PR and I don't think there is a way around it.

@edwardloveall `table_for` was just added in #123 for migrations. It infers the table name based on the model class. It uses the same `run_macro` that is used ehre

src/avram/associations.cr
paulcsmith: `table_for` is a macro to get the table name from a model class. `parent` is more like the belongs_to/has_one relationship. So Post (parent) and Comments (children). Maybe there is a better way to describe this though?

paulcsmith: Maybe we just call it `assoc_table`? Sometimes adding more words just makes things more confusing so maybe just `table`?

src/avram/associations.cr
paulcsmith: Actually. Maybe I can get it working. Let me try one more time :)

paulcsmith: Nope. Here is the issue. At macro compile time we just get the raw `table_for(MyModel)` not the result of that macro call. So it doesn't work for stuff like this.

Here's an example: https://play.crystal-lang.org/#/r/76bh

jwoertink: ZING!

src/avram/associations.cr
paulcsmith: I think I'll go with `table` for now, but you have a suggestion comment here. Happy to  change it if we find something great!

PR #133, opened 2019-07-03
add rollback_to feature

KCErb:
Wow, that was easier than I thought!

But I don't understand the specs well enough to be sure that these tasks don't have specs. Is there somewhere I'm missing that a task spec belongs?

Also you'll notice that I changed the messaging a bit in the test script. Since I put this before `rollback_all` I wanted to make sure I could visually compare (assuming that's our test method) what the message says we are rolling back to vs what actually got rolled back.

If this goes in, I'll be happy to update the docs (over on the lucky webpage right?). The doc should say something like "rollback_to <last stable integer timestamp>" and make it clear that we are *not* rolling back the migration named in the time stamp, we are rolling back *to* the state that that migration provided.

### DIFF
--- a/script/test
+++ b/script/test
@@ -14,6 +14,9 @@ then
     $COMPOSE lucky db.drop
     $COMPOSE lucky db.create
     $COMPOSE lucky db.migrate
+    printf "\nRolling back to 20180802180356\n"
+    $COMPOSE lucky db.rollback_to 20180802180356
+    printf "\nRolling back remainder\n"
     $COMPOSE lucky db.rollback_all
     $COMPOSE lucky db.migrate.one
     $COMPOSE lucky db.migrate
@@ -26,7 +29,7 @@ if [ $# -eq 0 ]
 then
     printf "\nChecking that tasks build correctly\n\n"
     $COMPOSE shards build
-    
+
     printf "\nChecking code formatting\n\n"
     if ! $COMPOSE crystal tool format --check src spec config > /dev/null; then
         printf "\nCode is not formatted.\n"
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -122,6 +122,14 @@ class Avram::Migrator::Runner
     end
   end
 
+  def rollback_to(last_version : Int64)
+    setup_migration_tracking_tables
+    subset = migrated_migrations.select do |mm|
+      mm.new.version.to_i64 > last_version
+    end
+    subset.reverse.each &.new.down
+  end
+
   def ensure_migrated!
     if pending_migrations.any?
       raise "There are pending migrations. Please run lucky db.migrate"
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -128,6 +128,7 @@ class Avram::Migrator::Runner
       mm.new.version.to_i64 > last_version
     end
     subset.reverse.each &.new.down
+    puts "Done rolling back to #{last_version}".colorize(:green)
   end
 
   def ensure_migrated!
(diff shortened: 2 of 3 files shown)


script/test
```diff
@@ -26,7 +29,7 @@ if [ $# -eq 0 ]
 then
     printf "\nChecking that tasks build correctly\n\n"
     $COMPOSE shards build
-    
+
     printf "\nChecking code formatting\n\n"
```
KCErb: Whoops, didn't notice this whitespace change. I guess my editor automatically kills whitespace at the end of lines. I can revert this if desired.

paulcsmith: No this is great! I'm glad your editor caught it :D

script/test
```diff
@@ -14,6 +14,9 @@ then
     $COMPOSE lucky db.drop
     $COMPOSE lucky db.create
     $COMPOSE lucky db.migrate
+    printf "\nRolling back to 20180802180356\n"
+    $COMPOSE lucky db.rollback_to 20180802180356
+    printf "\nRolling back remainder\n"
```
paulcsmith: This is great :)

src/avram/tasks/db/rollback_to.cr
```diff
@@ -0,0 +1,16 @@
...
+class Db::RollbackTo < LuckyCli::Task
+  summary "Rollback to a specific migration"
+
+  def call(version : String? = nil)
+    Avram::Migrator.run do
+      version = version || ARGV.first?
+      if version && version.to_i64?
+        Avram::Migrator::Runner.new.rollback_to(version.to_i64)
+      else
+        raise "Migration version is required. Example: lucky db.rollback_to 20180802180356".colorize(:red).to_s
```
paulcsmith: 👍

src/avram/migrator/runner.cr
```diff
@@ -122,6 +122,14 @@ class Avram::Migrator::Runner
     end
   end
 
+  def rollback_to(last_version : Int64)
+    setup_migration_tracking_tables
+    subset = migrated_migrations.select do |mm|
+      mm.new.version.to_i64 > last_version
+    end
+    subset.reverse.each &.new.down
```
paulcsmith: What do you think about printing a message after this like "Done rolling back to #{last_ version}"?

src/avram/tasks/db/rollback_to.cr
```diff
@@ -0,0 +1,16 @@
+require "colorize"
+
+class Db::RollbackTo < LuckyCli::Task
+  summary "Rollback to a specific migration"
+
+  def call(version : String? = nil)
```
paulcsmith: It looks like this argument is not being used so I think it can be removed

src/avram/migrator/runner.cr
KCErb: Fine by me

src/avram/tasks/db/rollback_to.cr
KCErb: I copied the pattern from `gen.migration`. I was assuming there was some reason to provide both a method argument and from ARGV in general. I'll remove it.

paulcsmith: Probably left on accident :S I think we can use something like this to test with in the future though. I need to go through and cleanup specs and do some refactoring at some point :)

KCErb: So should I leave it in in case of future test or leave it out for now? I'm happy both ways :)

paulcsmith: Sorry I wasn't clear there at all 🤣  you can go ahead and  remove it for now. I'll add it back in when I spruce up the tests

paulcsmith:
Travis is having an incident so I ran this locally and specs pass. Merging now. Thanks @KCErb :D

KCErb:
:tada: sweet

PR #134, opened 2019-07-03
renamed destroy_all to truncate.

### LINKED ISSUE #110: Rename `destroy_all` -> `truncate`
Because that's what it does 

Also it should be a class method because it ignores any other where clauses

jwoertink:
fixes #110

I also documented the Repo.truncate just so we don't get the two confused.

### DIFF
--- a/src/avram/query.cr
+++ b/src/avram/query.cr
@@ -2,7 +2,7 @@ require "./queryable"
 
 abstract class Avram::Query
   # runs a SQL `TRUNCATE` on the current table
-  def truncate
+  def self.truncate
     Avram::Repo.run do |db|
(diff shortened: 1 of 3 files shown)


src/avram/query.cr
```diff
@@ -1,7 +1,8 @@
 require "./queryable"
 
 abstract class Avram::Query
-  def destroy_all
+  # runs a SQL `TRUNCATE` on the current table
+  def truncate
```
paulcsmith: Could you make this a class method as well? Since it ignores any other conditions that might be good

paulcsmith:
Ran specs locally and merging. Hopefully Travis fixes their issues soon :)

Issue #126, opened 2019-07-02
Add lucky db.migrations.status


paulcsmith:
Prints the migration name, timestamp, and whether it has been migrated or not. Can probably use the shell-table shard like we do in Lucky for the route helper: https://github.com/luckyframework/lucky/blob/32db2b499227541ab43b5927bf88e1b95a3e53d8/tasks/routes.cr#L9-L17

smadeja:
Looks like this can be closed.

paulcsmith:
Ah yes thank you!

PR #135, opened 2019-07-04
Add lucky db.migrations.status CLI task

DevinRiley:
### 👋 

This is my first attempt at a contribution to Lucky and pretty much my first time writing Crystal code, so any / all feedback is super appreciated.

I took at a stab at adding the `lucky db.migrations.status` task from https://github.com/luckyframework/avram/issues/126

A few things to note:
- I added the `shell-table` shard as a dependency to help format the output.
- The `shell-table` shard is pointing to a branch on the fork at luckyframework. I didn't have a good reason to do this other than that's the version lucky itself is using (https://github.com/luckyframework/lucky/blob/32db2b499227541ab43b5927bf88e1b95a3e53d8/shard.yml#L47-L49). Happy to update this.
- It doesn't look like there are existing tests for CLI tasks so I didn't add any, but please point me in the right direction if I'm mistaken here.

Example output:
Screen Shot 2019-07-06 at 6 41 37 AM

Thanks!

### DIFF
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -137,18 +137,18 @@ class Avram::Migrator::Runner
     end
   end
 
+  def setup_migration_tracking_tables
+    DB.open(Avram::Repo.settings.url) do |db|
+      db.exec create_table_for_tracking_migrations
+    end
+  end
+
   private def migrated_migrations
     @@migrations.select &.new.migrated?
   end
 
   private def pending_migrations
     @@migrations.select &.new.pending?
-  end
-
-  private def setup_migration_tracking_tables
-    DB.open(Avram::Repo.settings.url) do |db|
-      db.exec create_table_for_tracking_migrations
-    end
   end
 
   private def prepare_for_migration
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -75,6 +75,21 @@ class Avram::Migrator::Runner
     end
   end
 
+  def self.setup_migration_tracking_tables
+    DB.open(Avram::Repo.settings.url) do |db|
+      db.exec create_table_for_tracking_migrations
+    end
+  end
+
+  private def self.create_table_for_tracking_migrations
+    <<-SQL
+    CREATE TABLE IF NOT EXISTS #{MIGRATIONS_TABLE_NAME} (
+      id serial PRIMARY KEY,
+      version bigint NOT NULL
+    )
+    SQL
+  end
+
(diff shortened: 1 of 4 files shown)


src/avram/migrator/runner.cr
```diff
@@ -137,6 +137,12 @@ class Avram::Migrator::Runner
     end
   end
 
+  def setup_migration_tracking_tables
```
paulcsmith: Could you make this a class method instead? I think that might make more sense since it doesn't need an instance variables

src/avram/tasks/db/migrations_status.cr
```diff
@@ -0,0 +1,42 @@
+require "shell-table"
+
+class Db::Migrations::Status < LuckyCli::Task
+  summary "Print the current status of migrations"
+
+  def call
+    if migrations.none?
```
paulcsmith: I love the use of `none?` 👍

src/avram/tasks/db/migrations_status.cr
```diff
@@ -0,0 +1,42 @@
...
+      labels: ["Migration", "Status"],
+      label_color: :white,
+      rows: migration_statuses
+    )
+  end
+
+  private def migration_statuses
+    migrations.map do |migration|
+      status = if migration.new.migrated?
+                 "Migrated".colorize(:green)
```
paulcsmith: I think the yellow and green clash a bit. Could you leave this the default color?

src/avram/migrator/runner.cr
DevinRiley: Updated in https://github.com/luckyframework/avram/pull/135/commits/f5dd3b12bd18c229595a83777f649b5757997636

Note that I also pulled up `create_table_for_tracking_migrations` into a private class method, since it too does not depend on instance variables but is called by `setup_migration_tracking_tables`

src/avram/tasks/db/migrations_status.cr
DevinRiley: Fixed in https://github.com/luckyframework/avram/pull/135/commits/69e31114c0065ed2038b4f645e2f335da39a5823

Now looks like this:
Screen Shot 2019-07-06 at 6 41 37 AM

src/avram/tasks/db/migrations_status.cr
DevinRiley: Crystal has been awesome to work in so far, I love little methods like this

DevinRiley:
@paulcsmith This is ready for another look

PR #136, opened 2019-07-06
Multi database support

### LINKED ISSUE #9: Multi-database support
- [ ] ~How to  handle migrations?~ #137 
- [ ] Allow specifying separate read/write databases per model
- [x] How to configure multiple database

## Ideas for migrations

```crystal
PrimaryDatabase.configure do |settings|
  settings.url = "some_url"
  settings.migration_folder = "db/migrations/primary" # default is "db/migrations"
end

SpecialDatabase.configure do |settings|
  settings.url = "some_url"
  settings.migration_folder = "db/migrations/special"
end
```

Then the Migration file can look at the file path to determine which repo it is for.

## Configure database

```crystal
abstract class BaseModel < Avram::Model
  def self.database
     PrimaryDatabase
  end
end
```

## Prior art

https://edgeguides.rubyonrails.org/active_record_multiple_databases.html

paulcsmith:
Closes #9

### DIFF
--- a/config/database.cr
+++ b/config/database.cr
@@ -1,10 +1,28 @@
 database_name = "avram_dev"
 
-Avram::Database.configure do |settings|
+class TestDatabase < Avram::Database
+end
+
+class DatabaseWithIncorrectSettings < Avram::Database
+end
+
+TestDatabase.configure do |settings|
   settings.url = ENV["DATABASE_URL"]? || Avram::PostgresURL.build(
     hostname: "db",
     database: database_name,
     username: "lucky",
     password: "developer"
   )
+end
+
+DatabaseWithIncorrectSettings.configure do |settings|
+  settings.url = Avram::PostgresURL.build(
+    hostname: "db",
+    database: database_name,
+    username: "incorrect"
+  )
+end
+
+Avram.configure do |settings|
+  settings.database_to_migrate = TestDatabase.new
 end
--- a/config/database.cr
+++ b/config/database.cr
@@ -1,6 +1,6 @@
 database_name = "avram_dev"
 
-Avram::Repo.configure do |settings|
+Avram::Database.configure do |settings|
   settings.url = ENV["DATABASE_URL"]? || Avram::PostgresURL.build(
     hostname: "db",
     database: database_name,
--- a/spec/multi_database_spec.cr
+++ b/spec/multi_database_spec.cr
@@ -1,0 +1,24 @@
+require "./spec_helper"
+
+private class ModelWithBadDatabase < BaseModel
+  table do
+  end
+
+  # Use a different database for this fake model
+  def self.database
+    DatabaseWithIncorrectSettings
+  end
+end
+
+describe "Configuring and connecting to different databases" do
+  it "tries to connect to the configured database" do
+    # It will fail to connect which is what we expect since we configured
+    # a database with an incorrect URL
+    #
+    # If it does not raise an error then the connection is good,
+    # which is not what we configured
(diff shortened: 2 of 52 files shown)


paulcsmith:
@jwoertink this now  allows configuring multiple databases and allows configuring what database each model uses. Does this look like it would work for your needs?

config/database.cr
```diff
@@ -1,10 +1,17 @@
 database_name = "avram_dev"
 
-Avram::Repo.configure do |settings|
+class TestDatabase < Avram::Database
```
paulcsmith: Here  is how to create a new Database

config/database.cr
```diff
@@ -1,10 +1,17 @@
...
+end
+
+TestDatabase.configure do |settings|
```
paulcsmith: Configure databases

spec/support/base_model.cr
```diff
@@ -0,0 +1,5 @@
+class BaseModel < Avram::Model
+  def self.database
+    TestDatabase
+  end
```
paulcsmith: Here is how you set what database a model uses. Can be done on the abstract class or could be overridden per model

paulcsmith:
Yeah I need to add some more specs and a few other things but wanted to make sure this would actually address the problem. Looks like it will so I'll finish it up so we can get you guys on Avram :D

paulcsmith:
Spec added. Ready for another review!

spec/multi_database_spec.cr
```diff
@@ -0,0 +1,24 @@
...
+    DatabaseWithIncorrectSettings
+  end
+end
+
+describe "Configuring and connecting to different databases" do
+  it "tries to connect to the configured database" do
+    # It will fail to connect which is what we expect since we configured
+    # a database with an incorrect URL
+    #
+    # If does not raise an error then that means it is using the good connection,
```
jwoertink:
```suggestion
    # If it does not raise an error then the connection is good,
```

spec/multi_database_spec.cr
```diff
@@ -0,0 +1,24 @@
...
+    # which is not what we configured
+    expect_raises Avram::ConnectionError do
```
jwoertink: Should this raise an exception error message letting us know which one it wasn't able to connect to? Our app could have up to 4 PG connections technically (a bit absurd, I know), so if for some reason one of them didn't connection, it would be nice to see `Unable to connect with DatabaseWithIncorrectSettings` or something.

src/avram/database.cr
```diff
@@ -1,38 +1,65 @@
...
+  def self.rollback
+    new.rollback
+  end
+
+  # runs a SQL `TRUNCATE` on all tables in the database
+  def self.truncate
+    new.truncate
+  end
+
+  def self.transaction
```
jwoertink: Could you add a quick comment on what this does? Will make it easier to add to guides later.

spec/multi_database_spec.cr
paulcsmith: Love it

spec/multi_database_spec.cr
paulcsmith: Yes great point!

src/avram/database.cr
paulcsmith: Yup!

paulcsmith:
This will require changes to the guides and also will need changes in Lucky CLI to generate a default database. I think we'll call it `AppDatabase` unless you've got other ideas

PR #141, opened 2019-07-10
Use underscored model class as param_key

### LINKED ISSUE #102: Avram::SaveOperation#form_name/param_key should use the model as the param key
Instead of using the SaveOperation name it should use the model. That way if you have a `SignUserUp` operation the param key would still be `user`. You can always override with the `param_key` macro though

paulcsmith:
Underscore class and name for  VirtualOperation

Closes #102

### DIFF
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -4,7 +4,7 @@ require "./nested_save_operation"
 require "./needy_initializer_and_save_methods"
 require "./virtual"
 require "./mark_as_failed"
-require "./param_key"
+require "./param_key_override"
 require "./save_operation_errors"
 require "./inherit_attributes"
 
@@ -17,7 +17,7 @@ abstract class Avram::SaveOperation(T)
   include Avram::MarkAsFailed
   include Avram::SaveOperationErrors
   include Avram::InheritAttributess
-  include Avram::ParamKey
+  include Avram::ParamKeyOverride
 
   enum SaveStatus
     Saved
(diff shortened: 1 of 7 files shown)


src/avram/save_operation.cr
```diff
@@ -44,6 +44,10 @@ abstract class Avram::SaveOperation(T)
   abstract def primary_key_name
   abstract def database
 
+  def self.param_key
+    T.name.underscore
+  end
```
paulcsmith: Now we use the model class. So `TaxId` model would have a `param_key` of `tax_id` no matter what the operation is named. I think this makes a lot more sense and keeps things consistent and easy to learn

src/avram/virtual_operation.cr
```diff
@@ -31,4 +31,8 @@ class Avram::VirtualOperation
   def attributes
     virtual_attributes
   end
+
+  def self.param_key
+    name.underscore
```
paulcsmith: This is a little trickier. Since  `VirtualForm`  does not have a model to work with it is hard to tell what this should be. 

Should we underscore the operation name like I did here? `SignInUser` -> `sign_in_user`, `ResetPassword` -> `reset_password`

Or should we not even try to guess it and have people set a param_key when using a VirtualOperation? 

I think underscoring the class name is fine but would love opinions @jwoertink @edwardloveall

edwardloveall: I think default of underscored operation class name is fine, with a way to override it maybe?

paulcsmith: You can override now with `param_key :my_custom_key`! 👍

jwoertink: Yeah, you can always override it anyway. Path of least resistance.

paulcsmith: Thanks for the input both of you! I’m glad we all agree. I’ll merge this in!

PR #142, opened 2019-07-10
/bin/bash is not available everywhere

peterhoeg:
Some distributions do not have /bin/bash so use /usr/bin/env to find it.


peterhoeg:
Sure, will do.

jwoertink:
I've verified these through `checkbashism`. Only "warning" is the use of `command` without the `-p` option. However, I don't think that's really an issue since `docker-compose` is added to `$PATH`.

PR #146, opened 2019-07-11
Handle more time formats and allow custom formats

### LINKED ISSUE #40: Handle more input types for the Time type
Especially for the date time helpers in HTML. That way they work out of the box

paulcsmith:
Closes #40

### DIFF
--- a/src/avram/charms/time_extensions.cr
+++ b/src/avram/charms/time_extensions.cr
@@ -3,14 +3,52 @@ struct Time
     alias ColumnType = Time
     include Avram::Type
 
+    TIME_FORMATS = [
+      Time::Format::ISO_8601_DATE_TIME,
+      Time::Format::RFC_2822,
+      Time::Format::RFC_3339,
+      # Dates and times go last, otherwise it will parse strings with both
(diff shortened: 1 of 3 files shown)


src/avram/charms/time_extensions.cr
```diff
@@ -3,16 +3,51 @@ struct Time
...
+      end.try do |format|
+        SuccessfulCast.new format.parse(value).to_utc
+      end
+    end
+
+    def self.try_parsing_with_string_formats(value)
+      Avram.settings.time_formats.find do |format|
+        begin
+          Time.parse(value, format, Time::Location.load("UTC"))
+        rescue e : Time::Format::Error
```
paulcsmith: I wish Crystal had some way to return `nil`  but it only raises when parsing so we have to do this kind of hacky begin/rescue dance :(

jwoertink: Working with Time in crystal is actually really clunky.

jwoertink:
Looks good to me!

paulcsmith:
Thanks for the reviews. And thanks for the kind words Edward!

PR #149, opened 2019-07-14
Improve callbacks and operation setup and validation 

paulcsmith:
- [x]  Needs to accept a block in callbacks too. 
- [x] call `before_save` in `valid?`
- [x] add specs to ensure after commit runs after the transaction 
- [x] add docs for callbacks. Explain when they run. Whether inside a transaction, etc. 
- [x] document `valid?`. Explain that it runs `before_save` callbacks but doesn’t save
- [x] Rename `VirtualOperation` to `Operation`

## Do later

This requires more work, refactoring, and possible extracting to another shard for `LuckyOperation` or something. Can be figured out later

- [ ] add abstract `run` method to operations. And yield the operation and the result
- [ ] change SaveOperation to use `run` under the hood

### DIFF
--- a/spec/save_operation_callbacks_spec.cr
+++ b/spec/save_operation_callbacks_spec.cr
@@ -80,7 +80,7 @@ describe "Avram::SaveOperation callbacks" do
 
   it "does not run after_commit if rolled back" do
     post = PostBox.create
-    operation = CallbacksSaveOperation.new(post)
+    operation = CallbacksSaveOperation.new(post, rollback: true)
     operation.callbacks_that_ran.should eq([] of String)
 
     operation.save
--- a/spec/save_operation_callbacks_spec.cr
+++ b/spec/save_operation_callbacks_spec.cr
@@ -1,35 +1,21 @@
 require "./spec_helper"
 
 private class CallbacksSaveOperation < Post::SaveOperation
+  needs rollback : Bool = false
+
   @callbacks_that_ran = [] of String
   getter callbacks_that_ran
 
+  before_save setup_required_attributes
+
   before_save :run_before_save
-  before_save run_before_save_again
-
-  before_create :run_before_create
-  before_create run_before_create_again
-
-  before_update :run_before_update
-  before_update :run_before_update_again
+  before_save { run_before_save_again }
 
   after_save :run_after_save
   after_save run_after_save_again
 
-  after_create :run_after_create
-  after_create run_after_create_again
-
-  after_update :run_after_update
-  after_update :run_after_update_again
-
-  def prepare
-    setup_required_attributes
-    mark_callback "prepare"
-  end
-
-  def after_prepare
-    mark_callback "after_prepare"
-  end
+  after_commit :run_after_commit
+  after_commit :run_after_commit_again
 
   def run_before_save
     mark_callback "before_save"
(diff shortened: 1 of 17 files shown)


spec/save_operation_callbacks_spec.cr
```diff
@@ -22,15 +24,14 @@ private class CallbacksSaveOperation < Post::SaveOperation
   after_update :run_after_update
   after_update :run_after_update_again
 
+  after_commit :run_after_commit
+  after_commit :run_after_commit_again
```
paulcsmith: Add spec to make sure this is not run if transaction is rolled back

spec/save_operation_callbacks_spec.cr
```diff
@@ -89,12 +98,13 @@ private class CallbacksSaveOperation < Post::SaveOperation
 end
 
 describe "Avram::SaveOperation callbacks" do
-  it "does not run the save callbacks if just validating" do
+  it "does not run any callbacks if calling valid?" do
```
paulcsmith: Revert this change and run before save

spec/save_operation_callbacks_spec.cr
```diff
@@ -1,35 +1,21 @@
...
+  before_save { run_before_save_again }
 
   after_save :run_after_save
   after_save run_after_save_again
 
-  after_create :run_after_create
-  after_create run_after_create_again
-
-  after_update :run_after_update
-  after_update :run_after_update_again
```
paulcsmith: Removing all the `_update|_create` callbacks. Instead use different objects. If  we need something like this later we can add methods to `SaveOperation` like `creating?` and `updating?` so you can use them inside the callbacks if really necessary

src/avram/callbacks.cr
```diff
@@ -1,61 +1,128 @@
...
+      {{ method_name.id }}(object)
     end
   end
 
-  macro before_update(method)
-    def before_update
-      {% if @type.methods.map(&.name).includes?(:before_update.id) %}
-        previous_def
-      {% end %}
+  {% for callbacks_without_block in [:after_save, :after_commit] %}
```
paulcsmith: Can't use callbacks with a block because  there is no way to set the block arg.

```crystal
macro  after_save 
  {{ yield }}
end

# Can be called with any number of block args.  They're all ignored
before_save do |foo, bar, baz|
end
```

So we can only allow named methods. I think this is ok since  those  steps typically benefit from a  well named method

src/avram/save_operation_template.cr
```diff
@@ -7,6 +7,14 @@ class Avram::SaveOperationTemplate
     end
 
     class ::{{ type }}::SaveOperation < Avram::SaveOperation({{ type }})
+      {% if primary_key_type.id == UUID.id %}
+        before_save set_uuid
```
paulcsmith: Doing this in a  `before_save` now. No special 'after_prepare' method

paulcsmith:
@jwoertink No I don't think we do because I don't think we handle that case yet. That may require re-working how params work :S. 

There is a spec for save_nested_operation_spec.cr that handles nested operations, but not array params

paulcsmith:
Thanks or the review! Gonna go ahead and merge this

Issue #75, opened 2019-03-23
Multi-column indices


HarrisonB:
While it looks like `Avram::Migrator::CreateIndexStatement` supports multi-column indices, it doesn't appear that `Avram::Migrator::IndexStatementHelpers#add_index` supports it. The change seems like a pretty easy one. Just need to change
```crystal
def add_index(column : Symbol, unique = false, using : Symbol = :btree)
```
to
```crystal
def add_index(columns : Columns, unique = false, using : Symbol = :btree)
```
and resolve other references that might need to change (and add tests).

jwoertink:
This one is a little strange. I think `add_index` is just being used for when you define a column, and want an index on that one.

https://github.com/luckyframework/avram/blob/master/src/avram/migrator/index_statement_helpers.cr#L7
```crystal
add name : String, index: true
```

But if you wanted to create a multi-column index, you can do that with `create_index` https://github.com/luckyframework/avram/blob/master/src/avram/migrator/statement_helpers.cr#L34

```crystal
create_index :users, [:first_name, :last_name]
```

I wonder if `add_index` is meant to be exposed like that or if it should just be internal...

```crystal
create :users do
  add first_name : String
  add last_name : String
  add_index [:first_name, :last_name]
end

create_index :users, [:first_name, :last_name]
```

I guess I can see both ways being nice to use.

Issue #145, opened 2019-07-10
Polymorphic associations


paulcsmith:
```crystal
class Comment < BaseModel
  table do
    polymorphic_belongs_to commentable : Post | Video | Photo
  end
end

class Post < BaseModel
  table do
    # Maybe we can figure out a way to look the comment model and know that it is polymorphic,      
    # but if not then we can pass the option like this:
    has_many comments : Comment, polymorphic: :commentable
  end
end
```

These would also add methods for querying and saving:

```crystal
CommentQuery.new.commentable(some_photo) # Comments for the photo
SaveComment.create!(commentable: some_photo)
```

## Migrations

`add_polymorphic_belongs_to :commentable`

## How do it manually in the meantime

https://gist.github.com/paulcsmith/39d2835795f184342296690e6ca324fc

paulcsmith:
Here's how to do this manually right now https://gist.github.com/paulcsmith/39d2835795f184342296690e6ca324fc

PR #150, opened 2019-07-15
Polymorphism support

fernandes:
Based on https://github.com/luckyframework/avram/issues/145 and the [gist](https://gist.github.com/paulcsmith/39d2835795f184342296690e6ca324fc)

To get this PR ready needs to:

- [x] Create sugar method to add polymorphism on receiver class

Do

```crystal
  table do
    column commentable : (Company | Employee)
  end
```

or a macro ta add to table AND add the `commentable` method

Instead of

```crystal
  table do
    column commentable_type : String?
    column commentable_id : Int32?
  end
```

- [x] Create sugar method to add polymorphism on migration

Do

```crystal
  add_polymorphic commentable
```

Instead of:

```crystal
  add commentable_type : String?
  add commentable_id : Int32?
```

- [x] Add ENUM type to Avram (maybe a future PR), enum would make easy (and safe) at DB level, which values are allowed in the `polymorphic_type` field


db/migrations/20190715125635_add_commentable_to_comment.cr
```diff
@@ -0,0 +1,15 @@
+class AddCommentableToComment::V20190715125635 < Avram::Migrator::Migration::V1
+  def migrate
+    alter :comments do
+      add commentable_type : String?
+      add commentable_id : Int64?
```
paulcsmith: I think it would be nice to have an `add_polymorphic_belongs_to :commentable` that generates these for you. It could take a `foreign_key_type` option that defaults to `Int64`. Also would be great to specify optional or not with `optional: true`

db/migrations/20190715125635_add_commentable_to_comment.cr
```diff
@@ -0,0 +1,15 @@
...
+    end
+  end
+
+  def rollback
+    alter :comments do
+      remove :commentable_type
+      remove :commentable_id
```
paulcsmith: `remove_polymorphic_belongs_to :commentable` would be wonderful to cover this

spec/polymorphism_spec.cr
```diff
@@ -0,0 +1,64 @@
+require "./spec_helper.cr"
+
+private class CommentForm < Comment::SaveOperation
+  permit_columns commentable_id
+  permit_columns commentable_type
+
+  def prepare
+    # Make sure we only save values we understand
+    validate_inclusion_of commentable_type, in: ["Employee", "Company"]
```
paulcsmith: This is great, but I believe should be done automatically when specifying in the model that it is polymorphic with `polymorphic_belongs_to` as described in #145

spec/polymorphism_spec.cr
```diff
@@ -0,0 +1,64 @@
...
+    employee = EmployeeBox.create
+
+    comment1 = CommentBox.create &.post_id(post.id).body("Comment 1")
+    CommentForm.update!(comment1, {"commentable_type" => "Company", "commentable_id" => company.id.to_s})
+
+    comment2 = CommentBox.create &.post_id(post.id).body("Comment 2")
+    CommentForm.update!(comment2, {"commentable_type" => "Company", "commentable_id" => company.id.to_s})
+
+    comment3 = CommentBox.create &.post_id(post.id).body("Comment 3")
+    CommentForm.update!(comment3, {"commentable_type" => "Employee", "commentable_id" => employee.id.to_s})
```
paulcsmith: Rather than a hash you can pass named args, for example `commentable_type: "Employee"`

paulcsmith:
Oh sorry! Should have read your description. I see that you are planning to do that later. LMK when you've started on that and I can take a look!

jwoertink:
This will be awesome! I definitely need this in my app. Thanks for tackling it..

Just one thing to note in reference to your PR description:

```crystal
# If you have this
table do
  column commentable : (Company | Employee)
end

# it should do this
table do
  # not nillable
  column commentable_type : String
  column commentable_id : Int32
end
```

I think this may require changing a few things, but right now there's a lot of assumptions that a Union means it's nillable. This would create a union that's not nillable. However, you may want a nillable polymorphic association like


```crystal
# If you have this
table do
  # can be nillable
  column commentable : (Company | Employee | Nil)
end

# it should do this
table do
  column commentable_type : String?
  column commentable_id : Int32?
end
```

Lastly, keep in mind that `commentable_id` could be any type that a primary key could be (i.e. Int64, UUID, etc...)

fernandes:
hello guys! yeah, that was the idea of the PR, more ideas/feedback.. thanks for hoping in! 😉

db/migrations/20190715125635_add_commentable_to_comment.cr
fernandes: added! 😄

db/migrations/20190715125635_add_commentable_to_comment.cr
fernandes: added! 😄

spec/polymorphism_spec.cr
fernandes: solved, thanks for the tip heheh you told me this before, but my memory usually fails hahaha

spec/polymorphism_spec.cr
fernandes: tackling this next!

fernandes: couldn't figure out how to add this validation on the default operation, could you please give me some direction?

src/avram/model.cr
```diff
@@ -184,6 +184,44 @@ class Avram::Model
     {% COLUMNS << {name: type_declaration.var, type: data_type, nilable: nilable.id, autogenerated: autogenerated} %}
   end
 
+  macro polymorphic_belongs_to(type_declaration, autogenerated = false, optional = false)
+
+    {{ nilable = !optional }}
+
+    {% COLUMNS << {name: type_declaration.var + "_id", type: Int64, nilable: nilable.id, autogenerated: autogenerated} %}
+    {% COLUMNS << {name: type_declaration.var + "_type", type: String, nilable: nilable.id, autogenerated: autogenerated} %}
+
```
paulcsmith: To validate_inclusion_of you can add a before_save in here:

```crystal
macro finished
  class SaveOperation
    before_save :validate_polymorphic_association
       
    private def validate_polymorphic_association
      validate_inclusion_of {{ type_declartion.var }}_type, in: []
    end
  end
end
```

Where the `in` is the underscored class name of the passed in  types

fernandes: `Expected Avram::InvalidSaveOperationError but nothing was raised`

not if I put the validation direct on the method `Expected Avram::InvalidSaveOperationError but nothing was raised`

ps: It's executing this method on the callback

paulcsmith: Hmm try this instead:

```crystal
def prepare
  previous_def
  validate_inclusion #...
end
```

I'm redoing how before_save works, but it is not in yet, so doing it in prepare might be better

fernandes: not even: `{{ type_declaration.var }}_type.add_error "test error"` raised exception hehe

paulcsmith: Hmm maybe need to add `class {{ @type }}::SaveOperation` also you can rebase and try `before_save` again. The new behavior has just been merged. If that still doesn't work I can pull it down and try to figure out what's up :P

fernandes: solved! 🎉

fernandes:
PR is ready for review 😄

spec/polymorphism_spec.cr
fernandes: solved! 🎉

src/avram/migrator/alter_table_statement.cr
```diff
@@ -127,6 +127,27 @@ class Avram::Migrator::AlterTableStatement
     {% end %}
   end
 
+  alias POLYMORPHIC_TYPES = Int16.class | Int32.class | Int64.class | UUID.class
```
jwoertink: Is this used anywhere?

fernandes: I think it was using before `foreign_key_type` hahahah good catch, thanks!

src/avram/migrator/index_statement_helpers.cr
```diff
@@ -4,13 +4,18 @@ module Avram::Migrator::IndexStatementHelpers
   private getter index_statements = [] of String
 
   # Generates raw sql for adding an index to a table column. Accepts 'unique' and 'using' options.
-  def add_index(column : Symbol, unique = false, using : Symbol = :btree)
+  def add_index(column : Columns, unique = false, using : Symbol = :btree)
```
jwoertink: This actually fixes #75

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+          validate_required {{ type_declaration.var }}_type
+          validate_required {{ type_declaration.var }}_id
+          validate_inclusion_of {{ type_declaration.var }}_type, in: {{type_declaration.type.types.map(&.stringify)}}
+          {% end %}
+        end
+      end
+    end
+
+    def {{type_declaration.var}} : {{type_declaration.type}}{% if nilable %}?{% end %}
+    {% if nilable %}
```
jwoertink: I'm horrible at reading macros, so I'm going to assume this is all good 😂

src/avram/migrator/index_statement_helpers.cr
fernandes: ohhh good you pointed the issue, I had this problem and changed (the `Columns` constant was already there heheh) also added the `drop_index` helper

src/avram/model.cr
fernandes: 😂 😂 😂 it's hard to read, needs to assume it's all good reading the tests that covers it hahaha

fernandes:
> Thanks for all the work on this! This is huge!

thanks, my pleasure I can contribute a little bit 😉

spec/support/comment.cr
```diff
@@ -6,6 +6,8 @@ class Comment < BaseModel
     timestamps
     column body : String
     belongs_to post : Post
+    polymorphic_belongs_to commentable : Employee | Company
+    polymorphic_belongs_to commentable_opt : Employee | Company | Nil
```
paulcsmith: Could you call this column `optional_commentable`? I kept thinking this was `otp` like a one time password

src/avram/associations.cr
```diff
@@ -257,6 +261,12 @@ module Avram::Associations
           end
           .preload_{{ through.id }}
           .results
+      {% elsif polymorphic != false %}
```
paulcsmith: You could do `polymorphic` without `!= false`

src/avram/associations.cr
```diff
@@ -257,6 +261,12 @@ module Avram::Associations
...
+        {{ model }}::BaseQuery
+          .new
+          .{{ polymorphic.id }}_type(self.class.to_s)
```
paulcsmith: Could yo do `self.class.name` just to make this super clear?

src/avram/migrator/alter_table_statement.cr
```diff
@@ -144,4 +163,14 @@ class Avram::Migrator::AlterTableStatement
     {% end %}
     remove {{ association_name }}_id
   end
+
+  macro remove_polymorphic_belongs_to(association_name)
+    {% unless association_name.is_a?(SymbolLiteral) %}
+      {% raise "remove_belongs_to expected a symbol like ':user', instead got: '#{association_name}'" %}
```
paulcsmith: Nice!

src/avram/migrator/alter_table_statement.cr
```diff
@@ -127,6 +127,25 @@ class Avram::Migrator::AlterTableStatement
...
+      default: nil
+    )
+    .build_add_statement_for_alter
+
+    rows << Avram::Migrator::Columns::StringColumn.new(
+      name: {{ "#{type_declaration.id}_type" }},
+      nilable: !{{ nilable }},
+      default: nil
+    )
+    .build_add_statement_for_alter
```
paulcsmith: I believe you can simplify these to `add {{ type_declaration.id}_type : StringColumn{% if nilable %}?{% end %}`

That way you don't need to manually use the column builder classes

src/avram/migrator/index_statement_helpers.cr
paulcsmith: Woo hoo!!

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
   end
 
+  macro polymorphic_belongs_to(type_declaration, autogenerated = false)
+
+    {% nilable = false %}
+    {% for type in type_declaration.type.types %}
+      {% if type.class_name == "Path" %}
+        {% nilable = true if type.names[0] == "Nil" %}
+      {% elsif type.class_name == "Union" %}
+        {% nilable = true if type.types.any? { |item| item.names[0] == "Nil" } %}
```
paulcsmith: Could you replace all the `[0]` with `first`

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
     {% COLUMNS << {name: type_declaration.var, type: data_type, nilable: nilable.id, autogenerated: autogenerated} %}
   end
 
+  macro polymorphic_belongs_to(type_declaration, autogenerated = false)
+
+    {% nilable = false %}
+    {% for type in type_declaration.type.types %}
+      {% if type.class_name == "Path" %}
+        {% nilable = true if type.names[0] == "Nil" %}
+      {% elsif type.class_name == "Union" %}
```
paulcsmith: You can check for type like this `type.is_a?(Path)` and `type.is_a?(Union)`. This is what we do in other places so I think it'd be good to do that here as well

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+
+    {% COLUMNS << {name: type_declaration.var + "_id", type: Int64, nilable: nilable.id, autogenerated: autogenerated} %}
+    {% COLUMNS << {name: type_declaration.var + "_type", type: String, nilable: nilable.id, autogenerated: autogenerated} %}
+
+    macro finished
+      class SaveOperation
+        before_save :validates_polymorphic_association
+
+        private def validates_polymorphic_association
+          previous_def
```
paulcsmith: Instead of naming this and using `previous_def` you can now do `before_save` with a block:

```crystal
before_save do
  # the code
end
```

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+
+          {% if nilable %}
+          validate_inclusion_of {{ type_declaration.var }}_type, in: {{type_declaration.type.types.map(&.stringify).push("").push(nil)}}
+          {% else %}
+          validate_required {{ type_declaration.var }}_type
+          validate_required {{ type_declaration.var }}_id
+          validate_inclusion_of {{ type_declaration.var }}_type, in: {{type_declaration.type.types.map(&.stringify)}}
```
paulcsmith: That is pretty clever 👍

src/avram/save_operation_template.cr
```diff
@@ -31,6 +31,8 @@ class Avram::SaveOperationTemplate
         :{{ primary_key_name.id }}
       end
 
+      private def validates_polymorphic_association; end
```
paulcsmith: This can be removed if using the suggestion about using `before_save` with a block

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+      end
+    end
+
+    def {{type_declaration.var}} : {{type_declaration.type}}{% if nilable %}?{% end %}
+    {% if nilable %}
+      return nil if {{type_declaration.var}}_id.nil?
+    {% end %}
+
+    {% if type_declaration.type.is_a?(Union) %}
+      {% first = true %}
```
paulcsmith: Can you explain what this is for? And/or maybe add some comments on how it is used?

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+      {% for type in type_declaration.type.types %}
+        {% if first == true && type.id != "Nil" %}
+      if {{type_declaration.var}}_type == "{{type}}"
+        {% if nilable %}
+        if lookup_id = {{type_declaration.var}}_id
+          {{type}}::BaseQuery.find(lookup_id)
+        end
+        {% else %}
+          {{type}}::BaseQuery.find({{type_declaration.var}}_id)
+        {% end %}
```
paulcsmith: I think this macro can be simplified with `try`

```crystal
{{ type_declaration.var }}_id.try do |lookup_id|
  {{type}}::BaseQuery.find(lookup_id)
end
```

I believe that will make it so you don't need to check for `nilable`

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+      {% end %}
+      else
+        raise "Unknown {{type_declaration.var}}: #{{{type_declaration.var}}_type}"
+      end
+      {% data_type = "#{type_declaration.type.types.first}".id %}
+      {% nilable = true %}
+    {% else %}
+      {% data_type = "#{type_declaration.type}".id %}
+      {% nilable = false %}
```
paulcsmith: It looks like this is not being used. Is it?

src/avram/model.cr
```diff
@@ -184,6 +184,79 @@ class Avram::Model
...
+        {% else %}
+          {{type}}::BaseQuery.find({{type_declaration.var}}_id)
+        {% end %}
+        {% end %}
+      {% end %}
+      else
+        raise "Unknown {{type_declaration.var}}: #{{{type_declaration.var}}_type}"
+      end
+      {% data_type = "#{type_declaration.type.types.first}".id %}
+      {% nilable = true %}
```
paulcsmith: This looks like it isn't being used either

paulcsmith:
Thank you so much for working on this! Left a few comments

paulcsmith:
Would be nice to also add some helpers to the query object and form so you can do `commentable(post)` but that can be added later I think

src/avram/migrator/alter_table_statement.cr
fernandes: This way, as this is not `nilable` it's asking for a `default_value`, but for me makes no sense to ask for a default value for an association heheh

the error message

> Try one of these...
> 
>   ▸ add optional_commentable_id : Int64, default: "Something"
>   ▸ add optional_commentable_id : Int64, fill_existing_with: "Something"
>   ▸ add optional_commentable_id : Int64, fill_existing_with: :nothing
> 
>         add_polymorphic_belongs_to :optional_commentable, optional: true

```crystal
add {{ type_declaration.id }}_id : {{ foreign_key_type }}{% if nilable %}?{% end %}
add {{ type_declaration.id }}_type : String{% if nilable %}?{% end %}
```

any idea?

src/avram/model.cr
fernandes: what if pass a `Nil` ? like

`String | Int32 | Nil`

this way `Nil` is a Path, that's why I'm checking... or we can choose do not support `Nil`, only `?` hahaha

wdyt?

src/avram/model.cr
fernandes: awesome!!! 🎉

src/avram/model.cr
fernandes: 🙇

src/avram/model.cr
fernandes: ohhhhh that's so nice.. good trick! simplified a lot hahahah

src/avram/model.cr
fernandes: Yeah, this is because the macro generates this code:

```
  27.     def commentable : Employee | Company
  28.     
  29. 
  30.     
  31.       
  32.       
  33.         
  34.       if commentable_type == "Employee"
  35.         commentable_id.try do |lookup_id|
  36.           Employee::BaseQuery.find(lookup_id)
  37.         end
  38.         
  39.       
  40.       
  41.         
  42.       elsif commentable_type == "Company"
  43.         commentable_id.try do |lookup_id|
  44.           Company::BaseQuery.find(lookup_id)
  45.         end
  46.         
  47.       
  48.       else
  49.         raise "Unknown commentable: #{commentable_type}"
  50.       end
  51.       
  52.       
  53.     
  54.     end
```

So for the first element, it's an `if`, for all subsequent, it is an `elseif` that why I use this `first` variable, any simpler way of doing this? should I add a comment?

src/avram/model.cr
fernandes: sorry for the dead code hehe

fernandes:
> I may have missed it, but is there a spec testing that you can retrieve the polymorphic belongs_to from a model?
> 
> ```
> PostQuery.new.first.comment
> ```

on the first test:

```crystal
company.comments.map(&.id).each do |id|
  [comment1.id, comment2.id].should contain(id)
end
employee.comments.map(&.id).should eq([comment3.id])
```

I get the already created objects that were used to build the relationships and check they have the `comments` associated to them, makes sense or should I change the test somehow?

fernandes:
> Would be nice to also add some helpers to the query object and form so you can do `commentable(post)` but that can be added later I think

yeah, I thought the PR was getting bigger and bigger hahaha but after we merge this one, I open a new one with this method

src/avram/migrator/alter_table_statement.cr
paulcsmith: I would pass `fill_existing_with: :nothing`

src/avram/migrator/create_table_statement.cr
```diff
@@ -123,6 +123,25 @@ class Avram::Migrator::CreateTableStatement
...
+      default: nil
+    )
+    .build_add_statement_for_create
+
+    rows << Avram::Migrator::Columns::StringColumn.new(
+      name: {{ "#{type_declaration.id}_type" }},
+      nilable: !{{ nilable }},
+      default: nil
+    )
+    .build_add_statement_for_create
```
paulcsmith: These could also probably use `add` like mentioned above

src/avram/model.cr
paulcsmith: I believe you can use `case` for this to simplify this so it generates something like:

```crystal
case commentbale_type
when "Company"
  # do stuff
when "Employee"
  # do stuff
else
  # raise
end
```

That way it works for any number of items

fernandes: you're totally right, way better heheh

fernandes:
@paulcsmith do you mean on `preloading_spec.cr` ?

paulcsmith:
Yep! Just to make sure preloading works as expected

fernandes:
Just pushed a fix to make the test more reliable (don't depend on Postgres _random_ default order hehehe)

paulcsmith:
Closing in favor of #165 

Thanks so much for working on this. Even though we won't merge it in it was super helpful to see it in action so we could figure out a way to do this with less code.

PR #151, opened 2019-07-15
Adding support for Arrays

### LINKED ISSUE #147: Support for Array fields
Postgres supports [array data types](https://www.postgresql.org/docs/10/arrays.html). These come with special querying syntax, and implementation. I think we can start with single dimensional arrays first which would look like

```crystal
column items : Array(String)
column ticks : Array(Int32)
```

A multidimensional may get tricky since postgres seems to allow for any number of nested arrays as long as they're all the same type. So I think `text[][][][][]` would be allow 😝  I have no idea how that would map to crystal cleanly, but also super edge case that could come later.

jwoertink:
fixes #147 
This PR adds support for postgres Arrays so you can define columns like

```crystal
column things : Array(String)
```

Some things to note: 

Crystal doesn't currently support nested constants in generics [ref](https://forum.crystal-lang.org/t/using-generic-with-inner-constants/953). This means that with the current code, we get macros generating things like `Array(String)::Lucky::ColumnType` which is a syntax error.

### DIFF
--- a/db/migrations/20190712145900_create_buckets.cr
+++ b/db/migrations/20190712145900_create_buckets.cr
@@ -2,14 +2,14 @@ class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
   def migrate
     create table_for(Bucket) do
       primary_key id : Int64
-      #add bools : Array(Bool)
-      #add prices : Array(Float64)
-      #add small_numbers : Array(Int16)
-      #add numbers : Array(Int32)
-      #add big_numbers : Array(Int64)
+      add bools : Array(Bool)
+      add prices : Array(Float64)
+      add small_numbers : Array(Int16)
+      add numbers : Array(Int32)
+      add big_numbers : Array(Int64)
       add names : Array(String)
-      #add dates : Array(Time)
-      #add ids : Array(UUID)
+      add dates : Array(Time)
+      add ids : Array(UUID)
     end
   end
 
--- a/db/migrations/20190712145900_create_buckets.cr
+++ b/db/migrations/20190712145900_create_buckets.cr
@@ -1,0 +1,19 @@
+class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
+  def migrate
+    create table_for(Bucket) do
+      primary_key id : Int64
+      add bools : Array(Bool)
+      add prices : Array(Float64)
+      add small_numbers : Array(Int16)
+      add numbers : Array(Int32)
+      add big_numbers : Array(Int64)
+      add names : Array(String)
+      add dates : Array(Time)
+      add ids : Array(UUID)
+    end
+  end
+
+  def rollback
+    drop :buckets
+  end
+end
--- a/db/migrations/20190712145900_create_buckets.cr
+++ b/db/migrations/20190712145900_create_buckets.cr
@@ -3,13 +3,13 @@ class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
     create table_for(Bucket) do
       primary_key id : Int64
       add bools : Array(Bool)
-      add prices : Array(Float64)
-      add small_numbers : Array(Int16)
-      add numbers : Array(Int32)
-      add big_numbers : Array(Int64)
-      add names : Array(String)
-      add dates : Array(Time)
-      add ids : Array(UUID)
+      #add prices : Array(Float64)
+      #add small_numbers : Array(Int16)
+      #add numbers : Array(Int32)
+      #add big_numbers : Array(Int64)
+      #add names : Array(String)
+      #add dates : Array(Time)
+      #add ids : Array(UUID)
     end
   end
 
--- a/db/migrations/20190712145900_create_buckets.cr
+++ b/db/migrations/20190712145900_create_buckets.cr
@@ -2,14 +2,13 @@ class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
   def migrate
     create table_for(Bucket) do
       primary_key id : Int64
+      add_timestamps
       add bools : Array(Bool)
       add prices : Array(Float64)
       add small_numbers : Array(Int16)
       add numbers : Array(Int32)
       add big_numbers : Array(Int64)
       add names : Array(String)
-      add dates : Array(Time)
-      add ids : Array(UUID)
     end
   end
 
--- a/db/migrations/20190712145900_create_buckets.cr
+++ b/db/migrations/20190712145900_create_buckets.cr
@@ -2,12 +2,12 @@ class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
   def migrate
     create table_for(Bucket) do
       primary_key id : Int64
-      add bools : Array(Bool)
+      #add bools : Array(Bool)
       #add prices : Array(Float64)
       #add small_numbers : Array(Int16)
       #add numbers : Array(Int32)
       #add big_numbers : Array(Int64)
-      #add names : Array(String)
+      add names : Array(String)
       #add dates : Array(Time)
       #add ids : Array(UUID)
     end
--- a/db/migrations/20190712145900_create_buckets.cr
+++ b/db/migrations/20190712145900_create_buckets.cr
@@ -4,7 +4,6 @@ class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
       primary_key id : Int64
       add_timestamps
       add bools : Array(Bool)
-      add prices : Array(Float64)
       add small_numbers : Array(Int16)
       add numbers : Array(Int32)
       add big_numbers : Array(Int64)
--- a/spec/array_column_spec.cr
+++ b/spec/array_column_spec.cr
@@ -1,0 +1,18 @@
+require "./spec_helper"
+
+private class BucketQuery < Bucket::BaseQuery
+end
+
+describe "Array Columns" do
+  it "fails when passing a single value to an array query" do
+    bucket = BucketBox.new.numbers([1, 2, 3]).create
+    expect_raises(PQ::PQError) do
+      BucketQuery.new.numbers(1).select_count
+    end
+  end
+
+  it "returns no results when passing in a proper query that doesn't match" do
+    bucket = BucketBox.new.numbers([1, 2, 3]).create
+    BucketQuery.new.numbers([1]).select_count.should eq 0
+  end
+end
--- a/spec/migrator/create_table_statement_spec.cr
+++ b/spec/migrator/create_table_statement_spec.cr
@@ -102,7 +102,7 @@ describe Avram::Migrator::CreateTableStatement do
       joined_at timestamptz NOT NULL DEFAULT NOW(),
       future_time timestamptz NOT NULL DEFAULT '#{Time.local.to_utc}',
       friend_count smallint NOT NULL DEFAULT '1',
-      friends text[] NOT NULL DEFAULT {'Paul'});
+      friends text[] NOT NULL DEFAULT '["Paul"]');
     SQL
   end
 
--- a/spec/migrator/create_table_statement_spec.cr
+++ b/spec/migrator/create_table_statement_spec.cr
@@ -86,6 +86,7 @@ describe Avram::Migrator::CreateTableStatement do
       add joined_at : Time, default: :now
       add future_time : Time, default: Time.local
       add friend_count : Int16, default: 1
+      add friends : Array(String), default: ["Paul"]
     end
 
     built.statements.size.should eq 1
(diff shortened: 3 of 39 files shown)


src/avram/model.cr
```diff
@@ -151,7 +151,11 @@ class Avram::Model
             type: PG::Numeric,
             convertor: Float64Convertor,
           {% else %}
+            {% if column[:type].is_a?(Generic) %}
+            type: {{column[:type]}},
```
jwoertink: This would pass `type: Array(String)` to DB.

src/avram/model.cr
```diff
@@ -168,7 +172,11 @@ class Avram::Model
   macro setup_getters(columns, *args, **named_args)
     {% for column in columns %}
       def {{column[:name]}}
-        {{ column[:type] }}::Lucky.from_db! @{{column[:name]}}
+        {% if column[:type].is_a?(Generic) %}
+          {{column[:type].type_vars.first}}::Lucky.from_db! @{{column[:name]}}, array: true
```
jwoertink: This would mean every `from_db!` method would now need an array option... This becomes very messy 😢

src/avram/migrator/create_table_statement.cr
```diff
@@ -78,15 +78,22 @@ class Avram::Migrator::CreateTableStatement
...
       {% type = type_declaration.type %}
       {% nilable = false %}
+      {% array = false %}
     {% end %}
 
     rows << Avram::Migrator::Columns::{{ type }}Column.new(
       name: {{ type_declaration.var.stringify }},
       nilable: {{ nilable }},
       default: {{ default }},
+      array: {{ array }}
```
jwoertink: This tells the column to be `text[]` instead of just `text` so postgres knows it's an array of strings.

src/avram/base_query_template.cr
```diff
@@ -25,7 +25,7 @@ class Avram::BaseQueryTemplate
       macro generate_criteria_method(query_class, name, type)
         def \{{ name }}
           column_name = "#{@@table_name}.\{{ name }}"
-          \{{ type }}::Lucky::Criteria(\{{ query_class }}, \{{ type }}).new(self, "#{@@table_name}.\{{ name }}")
+          \{{ type }}::Lucky::Criteria(\{{ query_class }}, \{{ type }}).new(self, column_name)
```
jwoertink: The `column_name` variable was never actually being used. We could also just get rid of it too

src/avram/save_operation.cr
```diff
@@ -82,7 +82,16 @@ abstract class Avram::SaveOperation(T)
     private def extract_changes_from_params
       permitted_params.each do |key, value|
         {% for attribute in attributes %}
-          set_{{ attribute[:name] }}_from_param value if key == {{ attribute[:name].stringify }}
+          set_{{ attribute[:name] }}_from_param(
+          {% if attribute[:type].is_a?(Generic) %}
+            # THIS IS A HACK.
+            # Need to figure out what `value` is, and how we can make it
+            # an array
+            [value]
```
jwoertink: This is always `Array(String)`. It eventually is passed to parse like `Bool::Lucky.parse`.

src/avram/type.cr
```diff
@@ -11,6 +11,18 @@ module Avram::Type
...
   end
 
+  def parse(values : Array(String))
+    values = values.map do |value|
+      parsed_result = parse(value)
+      if parsed_result.is_a? Avram::Type::SuccessfulCast
+        parsed_result.value
+      else
+        nil
+      end
```
jwoertink: This is due to that if a parse fails, `FailedCast` doesn't have a `value` method. Maybe if even one fails, the whole thing should? How would that look though?

paulcsmith: I think if one fails, they all should. Not sure how it would look though. Will need to dig in more later!

spec/type_extensions/array_spec.cr
```diff
@@ -0,0 +1,39 @@
...
+
+  it "parses Array(UUID) from Array(String)" do
+    result = UUID::Lucky.parse(["f51a73d5-bb75-465a-8f8b-d213658dfc63"])
+    result.value.should eq([UUID.new("f51a73d5-bb75-465a-8f8b-d213658dfc63")])
+  end
+
+  it "parses Array(Time) from Array(String)" do
+    result = Time::Lucky.parse(["2019-07-16 14:42:12"])
+    time = Time.parse_utc("2019-07-16 14:42:12", "%Y-%m-%d %H:%M:%S")
+    result.value.should eq([time])
```
jwoertink: This spec fails: (not sure why)

```
       Expected: [2019-07-16 14:42:12.0 UTC]
            got: [2019-07-16 00:00:00.0 UTC]

     # spec/type_extensions/array_spec.cr:37
```

src/avram/save_operation.cr
```diff
@@ -248,17 +261,16 @@ abstract class Avram::SaveOperation(T)
...
 
-  private def cast_value(value : Object)
-    case value
-    when JSON::Any
-      value.to_json
-    else
-      value.to_s
+    {% for column in columns %}
+    private def cast_value(value : {{ column[:type] }})
+      value.not_nil!.class.adapter.to_db(value.as({{ column[:type] }}))
```
jwoertink: Pass the `value` to it's `Lucky.to_db` for parsing. In most cases it's just calling `to_s` on the value, but in the case of the Arrays, it passes to `PQ::Param` to properly encode so `[true]` becomes `{t}`, and so on. 

We cast the value here with `value.as({{ column[:type] }})` otherwise the compiler sees a Union and throws no method overload. Also, I had to add the `not_nil!` because the compiler kept throwing undefined method adapter for Nil.class. In theory, it should never be nil because of the other `cast_value` overload, but maybe that would have to be defined later?

src/avram/charms/uuid_extensions.cr
```diff
@@ -1,6 +1,6 @@
 struct UUID
-  def blank?
```
jwoertink: moved this to the blank_extensions file

src/avram/charms/bool_extensions.cr
```diff
@@ -1,13 +1,13 @@
 struct Bool
-  def blank?
```
jwoertink: moved this to the blank_extensions file

src/avram/save_operation.cr
```diff
@@ -123,9 +123,17 @@ abstract class Avram::SaveOperation(T) < Avram::Operation
       end
 
       def set_{{ attribute[:name] }}_from_param(_value)
-        parse_result = {{ attribute[:type] }}::Lucky.parse(_value)
+        {% if attribute[:type].is_a?(Generic) %}
+          # HACK:
+          # We have to pass an array her, but what's `_value`?
+          # If `_value` is `"true"`, then we pass `["true"]` which can be parsed, but
+          # what if we need to pass in multiple values? I don't know that params handles that currently
```
paulcsmith: I think this is spot on. Right now there is no way for params to return anything other than a single value. I'd say that right now that is ok and can be handled later. People can manually parse JSON responses, or if it is URL encoded I'd say it's probably best to use a `String` attribute and manually parse it into an array.

Typically most sites that use these have to do some kind of param parsing/manipulation before save anyway so I think this is ok and can be improved later

src/avram/base_query_template.cr
paulcsmith: I like that it uses the name so it's clear what it is. Good catch on actually using it though haha

db/migrations/20190712145900_create_buckets.cr
```diff
@@ -0,0 +1,17 @@
+class CreateBuckets::V20190712145900 < Avram::Migrator::Migration::V1
```
paulcsmith: Can you give this a descriptive name like `CreateBucketsWithArrayTypes` so it's clear from the filename it is testing array types?

spec/migrator/create_table_statement_spec.cr
```diff
@@ -86,6 +86,7 @@ describe Avram::Migrator::CreateTableStatement do
       add joined_at : Time, default: :now
       add future_time : Time, default: Time.local
       add friend_count : Int16, default: 1
+      add friends : Array(String), default: ["Paul"]
```
paulcsmith: 😊

src/avram/base_query_template.cr
```diff
@@ -34,11 +34,19 @@ class Avram::BaseQueryTemplate
           {{ column[:name] }}.eq(value)
         end
 
-        generate_criteria_method(BaseQuery, {{ column[:name] }}, {{ column[:type] }})
+        {% if column[:type].is_a?(Generic) %}
```
paulcsmith: Maybe leave a comment that this is checking if the type is an Array?

src/avram/type.cr
```diff
@@ -32,5 +42,8 @@ module Avram::Type
   end
 
   class FailedCast
+    def value
```
paulcsmith: Why does `FailedCast` need a value? Could you link me to the code that needs it? Just making sure I understand :)

src/avram/migrator/create_table_statement.cr
```diff
@@ -78,9 +78,15 @@ class Avram::Migrator::CreateTableStatement
...
+      {% type = type_declaration.type.type_vars.first %}
+      {% nilable = false %}
+      {% array = true %}
     {% else %}
       {% type = type_declaration.type %}
       {% nilable = false %}
+      {% array = false %}
     {% end %}
 
     rows << Avram::Migrator::Columns::{{ type }}Column.new(
```
paulcsmith: I had an idea to clean up the defaults. Right now we have to remember to add `Array(TheType)` to each column. It also incorrectly would allow array defaults for non-array types and vice versa. 

Instead we could make the Column types generic and pass the default type to it. An example will be clear (I hope :P)

```crystal
# What we have now
rows << Avram::Migrator::Columns::{{ type }}Column.new(

# Proposed change 
rows << Avram::Migrator::Columns::{{ type }}Column({% if array %}Array({% end %}{{ allowed_default_type }}{% if array %}){% end %}).new(
```

Then in the column classes:

```crystal
module Avram::Migrator::Columns
   class BoolColumn(T) < Base
     @default : T | Nil = nil
```

So now the default type matches the array type explicitly, and also removes some duplication. Does that make sense?

src/avram/save_operation.cr
paulcsmith: This makes sense. I think it would be good to include this as a comment in the code. Usually I'm anti-comments but in this case I think it would be handy

src/avram/migrator/create_table_statement.cr
jwoertink: Yeah that makes sense. Is that a change you want me to include in this PR?

paulcsmith: Yeah I think that'd be good to have so we make sure the types for defaults line up correctly

src/avram/type.cr
jwoertink: this isn't totally necessary, but without it there's a compilation error. If I remove it, then the `parse!` method above on line 24 would have to add in a catch for when `parse(value)` doesn't return a `SuccessfulCast`. Alternatively, the array parse method on line 20 above could just not return a `FailedCast`... but that adds in some other weird stuff. I'm down for either way though if you have a better suggestion that makes it cleaner.

paulcsmith:
I think handling the FailedCast explicitly might make the parse method clear. I think as it is it isn’t clear something went wrong

jwoertink:
> What happens if you try to query an array type with a single value?

This doesn't throw a compile time error (but I agree it probably should), but you do get a postgres error for `malformed array literal`. I can add in a catch for that

```crystal
rescue PQ::PQError => e
   if e.message =~ /malformed array literal/
     raise Avram::MalformedArrayError.new
  end
```

Or we leave as is to get this PR out, then update with PRs later to clean up some stuff. I'd imagine there's gonna be tons of clean ups after the next crystal.

spec/array_column_spec.cr
```diff
@@ -0,0 +1,18 @@
+require "./spec_helper"
+
+private class BucketQuery < Bucket::BaseQuery
+end
+
+describe "Array Columns" do
+  it "fails when passing a single value to an array query" do
+    bucket = BucketBox.new.numbers([1, 2, 3]).create
```
paulcsmith: Could you create an issue to tackle this later? I think it’ll require reworking the criteria methods a bit and will probably be a pretty big refactor

src/avram/migrator/columns/float_column.cr
```diff
@@ -1,10 +1,10 @@
 require "./base"
 
 module Avram::Migrator::Columns
-  class FloatColumn < Base
+  class FloatColumn(T) < Base
```
paulcsmith: I think this can be abstract now

PR #159, opened 2019-07-23
Clean up validations, add docs, cleanup specs

paulcsmith:


spec/attribute_spec.cr
```diff
@@ -17,4 +25,15 @@ describe "Avram::Attribute" do
...
+
+  it "can accept custom callable messages" do
+    attribute = Avram::Attribute.new(name: :attr_name, param: nil, value: "fake", param_key: "na")
+    attribute.add_error ->(attribute_name : String, attribute_value : String) { "#{attribute_name} message from Proc" }
+    attribute.add_error CallableMessage.new
+
+    attribute.errors.should eq([
+      "attr_name message from Proc",
+      "Error: attr_name with value of 'fake' is invalid",
+    ])
```
paulcsmith: Let's just test this in one spot instead of for every single validation

spec/validations_spec.cr
```diff
@@ -1,289 +1,183 @@
...
+end
 
-  column String, name
-  column String, name_confirmation
-  column String, city
-  column String, state
-  column Int32?, age
-  column Bool?, terms
+private def attribute(value)
+  Avram::Attribute.new(value: value, param: nil, param_key: "fake", name: :fake)
```
paulcsmith: Create fake attributes more easily

src/avram/attribute.cr
```diff
@@ -1,4 +1,5 @@
 class Avram::Attribute(T)
+  alias ErrorMessage = String | Proc(String, String, String) | Avram::CallableErrorMessage
```
paulcsmith: Extract a type alias  that we can use in `Avram::Validations` to give better error message and make it clearer what you can use for a custom message in validations

src/avram/validations.cr
```diff
@@ -1,39 +1,96 @@
-require "./validations/**"
+require "./validations/callable_error_message"
```
paulcsmith: Be clear about what we need

src/avram/validations.cr
```diff
@@ -1,39 +1,96 @@
...
 
+# A number of methods for validating Avram::Attributes
+#
+# This module is included in `Avram::Operation` and `Avram::SaveOperation`
 module Avram::Validations
-  private def validate_required(*attributes, message = "is required")
+  extend self
```
paulcsmith: Allow using validations without including them

src/avram/validations.cr
```diff
@@ -1,39 +1,96 @@
...
+  #
+  # You can pass in one or more attributes at a time. The attribute will be
+  # marked as invalid if the value is `nil`, or "blank" (empty strings or strings with just whitespace)
+  #
+  # `false` is not considered invalid.
+  #
+  # ```
+  # validate_required name, age, email
+  # ```
+  def validate_required(*attributes, message : Avram::Attribute::ErrorMessage = "is required")
```
paulcsmith: There was no need for these to be `private` so I made them `public`

src/avram/validations.cr
```diff
@@ -1,39 +1,96 @@
...
+  # Example:
+  #
+  # ```
+  # validate_confirmation_of password, with: password_confirmation
+  # ```
+  #
+  # If `password_confirmation` does not match, it will be marked invalid.
+  def validate_confirmation_of(
+    attribute : Avram::Attribute(T),
+    with confirmation_attribute : Avram::Attribute(T),
```
paulcsmith: Make sure the Attribute types are the same. So you can't accidentally compare  a Bool to a String

src/avram/validations.cr
```diff
@@ -1,39 +1,96 @@
...
+  # Validates that the attribute value is in a list of allowed values
+  #
+  # ```
+  # validate_inclusion_of state, in: ["NY", "MA"]
+  # ```
+  #
+  # This will mark `state` as invalid unless the value is `"NY"`, or `"MA"`.
+  def validate_inclusion_of(
+    attribute : Avram::Attribute(T),
+    in allowed_values : Enumerable(T),
```
paulcsmith: Make sure the Enumerable only has types that include the Attribute could be. E.g. an Int32 should not have `allowed_values` that are `Strings`

Issue #15, opened 2019-02-02
Add `debug_info` to `LuckyRecord::Field`


paulcsmith:
We ran into https://github.com/luckyframework/lucky_record/pull/272#pullrequestreview-141172842 and finding the issue was a bit tricky because it said the Time we were passing was "invalid" but it wasn't clear why. 

I think we should have a `debug_info` instance var to LuckyRecord::Field that is an array of strings. This way you could add error messages when parsing that are useful to developers. For example, it would have added: `Invalid ("2018-08-08T16:00:00.000Z"): "Time format did not include time zone and no default location provided"`

@edwardloveall What do you think? This way when someone `p the_field` it will show you more helpful information

PR #165, opened 2019-07-24
Polymorphic helpers

### LINKED ISSUE #145: Polymorphic associations

paulcsmith:
Closes #145

### DIFF
--- a/db/migrations/20190723233130_create_polymorphic_belongs_to.cr
+++ b/db/migrations/20190723233130_create_polymorphic_belongs_to.cr
@@ -1,0 +1,28 @@
+class CreatePolymorphicBelongsTo::V20190723233130 < Avram::Migrator::Migration::V1
+  def migrate
+    create table_for(PolymorphicTask) do
+      primary_key id : Int64
+      add_timestamps
+      add title : String
+    end
+
+    create table_for(PolymorphicTaskList) do
+      primary_key id : Int64
+      add_timestamps
+      add title : String
+    end
+
+    create table_for(PolymorphicEvent) do
+      primary_key id : Int64
+      add_timestamps
+      add_belongs_to task : PolymorphicTask?, on_delete: :cascade
+      add_belongs_to task_list : PolymorphicTaskList?, on_delete: :cascade
+    end
+  end
+
+  def rollback
+    drop table_for(PolymorphicEvent)
+    drop table_for(PolymorphicTask)
+    drop table_for(PolymorphicTaskList)
+  end
+end
--- a/spec/polymorphic_spec.cr
+++ b/spec/polymorphic_spec.cr
@@ -1,0 +1,123 @@
+require "./spec_helper"
+
+include LazyLoadHelpers
+
+private class PolymorphicTask < BaseModel
+  table do
+    column title : String
+  end
+end
+
+private class PolymorphicTaskList < BaseModel
+  table do
+    column title : String
+  end
+end
+
+private class PolymorphicEvent < BaseModel
+  table do
+    belongs_to task : PolymorphicTask?
+    belongs_to task_list : PolymorphicTaskList?
+    polymorphic :eventable, associations: [:task, :task_list]
+  end
+end
+
+private class OptionalPolymorphicEvent < BaseModel
+  table :polymorphic_events do
+    belongs_to task : PolymorphicTask?
+    belongs_to task_list : PolymorphicTaskList?
+    polymorphic :optional_eventable, optional: true, associations: [:task, :task_list]
+  end
+end
+
+describe "polymorphic belongs to" do
+  it "sets up a method for accessing associated record" do
+    task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+    event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+    event.eventable.should eq(task)
+
+    task_list = PolymorphicTaskList::SaveOperation.create!(title: "Use Lucky")
+    event = PolymorphicEvent::SaveOperation.create!(task_list_id: task_list.id)
+    event.eventable.should eq(task_list)
+  end
+
+  it "can require preloading" do
+    with_lazy_load(enabled: false) do
+      expect_raises Avram::LazyLoadError do
+        task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+        event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+        event.eventable # should raise because it was not preloaded
+      end
+    end
+  end
+
+  it "has ! method to allow lazy loading" do
+    with_lazy_load(enabled: false) do
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      event.eventable!.should eq(task)
+    end
+  end
+
+  it "can preload the polymorphic associations" do
+    with_lazy_load(enabled: false) do
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      event = PolymorphicEvent::BaseQuery.new.preload_eventable.find(event.id)
+      event.eventable.should eq(task)
+
+      # Check that it preloads both belongs_to, not just the first
+      task_list = PolymorphicTaskList::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_list_id: task_list.id)
+      event = PolymorphicEvent::BaseQuery.new.preload_eventable.find(event.id)
+      event.eventable.should eq(task_list)
+    end
+  end
+
+  describe "required (default)" do
+    it "validates that exactly one polymorphic belongs to is allowed" do
+      operation = PolymorphicEvent::SaveOperation.create(task_id: 1, task_list_id: 1) do |operation, _event|
+        operation.valid?.should be_false
+        operation.task_list_id.errors.should eq(["must be blank"])
+      end
+
+      operation = PolymorphicEvent::SaveOperation.create do |operation, _event|
+        operation.valid?.should be_false
+        operation.task_id.errors.should eq(["at least one 'eventable' must be filled"])
+      end
+    end
+
+    it "type should not include nil" do
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      typeof(event.eventable).should eq(PolymorphicTask | PolymorphicTaskList)
+      typeof(event.eventable!).should eq(PolymorphicTask | PolymorphicTaskList)
+    end
+  end
+
+  describe "optional" do
+    it "allows not setting a polymorphic association" do
+      operation = OptionalPolymorphicEvent::SaveOperation.create do |operation, _event|
+        operation.valid?.should be_true
+      end
+    end
+
+    it "validates that at most one polymorphic belongs to is allowed" do
+      operation = OptionalPolymorphicEvent::SaveOperation.create(task_id: 1, task_list_id: 1) do |operation, _event|
+        operation.valid?.should be_false
+        operation.task_list_id.errors.should eq(["must be blank"])
+      end
+    end
+
+    it "can return 'nil'" do
+      event = OptionalPolymorphicEvent::SaveOperation.create!(task_id: nil, task_list_id: nil)
+      typeof(event.optional_eventable).should eq(PolymorphicTask | PolymorphicTaskList | Nil)
+      typeof(event.optional_eventable!).should eq(PolymorphicTask | PolymorphicTaskList | Nil)
+      event.optional_eventable!.should be_nil
+    end
+  end
+
+  # They must be nullable since only one can be filled in at a time.
+  # And remind to make migration optional too
+  pending "ensure defined polymorphic associations are nullable"
+end
--- a/spec/validations_spec.cr
+++ b/spec/validations_spec.cr
@@ -18,6 +18,60 @@ private def attribute(value)
 end
 
 describe Avram::Validations do
+  describe "validate_at_most_one_filled" do
+    it "marks filled attribute as invalid if more than one is filled" do
+      filled_attribute = attribute("filled")
+      filled_attribute_2 = attribute("filled")
+      blank_attribute = attribute("")
+      Avram::Validations.validate_at_most_one_filled(filled_attribute, filled_attribute_2, blank_attribute)
+      filled_attribute.valid?.should be_true
+      blank_attribute.valid?.should be_true
+      filled_attribute_2.errors.should eq(["must be blank"])
+    end
+
+    it "does not mark any fields as invalid if just one is filled" do
+      filled_attribute = attribute("filled")
+      blank_attribute = attribute("")
+
+      Avram::Validations.validate_at_most_one_filled(filled_attribute, blank_attribute)
+
+      filled_attribute.valid?.should be_true
+      blank_attribute.valid?.should be_true
+    end
+  end
+
+  describe "validate_exactly_one_filled" do
+    it "marks filled attribute as invalid if more than one is filled" do
+      filled_attribute = attribute("filled")
+      filled_attribute_2 = attribute("filled")
+      blank_attribute = attribute("")
+      Avram::Validations.validate_exactly_one_filled(filled_attribute, filled_attribute_2, blank_attribute)
+      filled_attribute.valid?.should be_true
+      blank_attribute.valid?.should be_true
+      filled_attribute_2.errors.should eq(["must be blank"])
+    end
+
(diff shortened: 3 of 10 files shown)


paulcsmith:
Still needs a bit of work to support custom foreign keys for `belongs_to` but this is close enough that I think a review would be great!

db/migrations/20190723233130_create_polymorphic_belongs_to.cr
```diff
@@ -0,0 +1,28 @@
...
+      primary_key id : Int64
+      add_timestamps
+      add title : String
+    end
+
+    create table_for(PolymorphicEvent) do
+      primary_key id : Int64
+      add_timestamps
+      add_belongs_to task : PolymorphicTask?, on_delete: :cascade
+      add_belongs_to task_list : PolymorphicTaskList?, on_delete: :cascade
```
jwoertink: Should there be a `has_many events` for the Task and List?

spec/polymorphic_spec.cr
```diff
@@ -0,0 +1,123 @@
...
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      event.eventable!.should eq(task)
+    end
+  end
+
+  it "can preload the polymorphic associations" do
+    with_lazy_load(enabled: false) do
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      event = PolymorphicEvent::BaseQuery.new.preload_eventable.find(event.id)
```
jwoertink: `preload_eventable` 👍

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
...
+  # Sets up helper methods for validating and loading polymorphic associatiions
+  #
+  # This will generate methods for validiating, preloading, and lazy loading
+  # polymorphic associations.
+  #
+  # Let's start with this example:
+  #
+  # > *Important note:* the `belongs_to` associations must be nilable since only
+  # > one polymorphic can be set at a time. Remember to also make the association
+  # > nilable in the migration, for example: `add_belongs_to photo : Photo?`
```
jwoertink: Does an exception get raised if you don't make it nillable? Is that even possible to catch at the time the polymorphic method is called?

db/migrations/20190723233130_create_polymorphic_belongs_to.cr
paulcsmith: There doesn’t have to be if you don’t use it. In real life you probably would have that but isn’t necessary for the test since we test has many elsewhere

src/avram/polymorphic.cr
paulcsmith: Not yet. It’s something that I’m trying to figure out how to check. You’ll probably notice it at runtime because you won’t be able to set just one polymorphic association without a database constraint error. I’d like to figure out how or if we can raise a better error. Either at compile time or at least at app start at runtime

paulcsmith:
The “photo || video” calls the belongs to method just like normal. So you preload your polymorphic association which will make queried to get those associations. Which may or may not be nil. If photo is nil then it’ll try video and so on. So yeah it makes queries using the belongs to association

spec/polymorphic_spec.cr
```diff
@@ -0,0 +1,123 @@
...
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      event.eventable!.should eq(task)
+    end
+  end
+
+  it "can preload the polymorphic associations" do
+    with_lazy_load(enabled: false) do
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
```
edwardloveall: Nit: probably don't need to save this variable.

spec/polymorphic_spec.cr
```diff
@@ -0,0 +1,123 @@
...
+      event = PolymorphicEvent::BaseQuery.new.preload_eventable.find(event.id)
+      event.eventable.should eq(task_list)
+    end
+  end
+
+  describe "required (default)" do
+    it "validates that exactly one polymorphic belongs to is allowed" do
+      operation = PolymorphicEvent::SaveOperation.create(task_id: 1, task_list_id: 1) do |operation, _event|
+        operation.valid?.should be_false
+        operation.task_list_id.errors.should eq(["must be blank"])
```
edwardloveall: Sounds tricky to do, but what if the error message was on the operation instead:

> When using polymorphic associations, only one model may be associated at once.
>
> Models associated currently:
> - Task
> - TaskList

spec/polymorphic_spec.cr
```diff
@@ -0,0 +1,123 @@
...
+      operation = PolymorphicEvent::SaveOperation.create do |operation, _event|
+        operation.valid?.should be_false
+        operation.task_id.errors.should eq(["at least one 'eventable' must be filled"])
+      end
+    end
+
+    it "type should not include nil" do
+      task = PolymorphicTask::SaveOperation.create!(title: "Use Lucky")
+      event = PolymorphicEvent::SaveOperation.create!(task_id: task.id)
+      typeof(event.eventable).should eq(PolymorphicTask | PolymorphicTaskList)
```
edwardloveall: You could also use `be_a` (I think)

```crystal
event.eventable.should be_a(PolymorphicTask | PolymorphicTaskList)
```

spec/validations_spec.cr
```diff
@@ -18,6 +18,60 @@ private def attribute(value)
 end
 
 describe Avram::Validations do
+  describe "validate_at_most_one_filled" do
+    it "marks filled attribute as invalid if more than one is filled" do
+      filled_attribute = attribute("filled")
+      filled_attribute_2 = attribute("filled")
+      blank_attribute = attribute("")
+      Avram::Validations.validate_at_most_one_filled(filled_attribute, filled_attribute_2, blank_attribute)
```
edwardloveall: This is _cooool_

spec/validations_spec.cr
```diff
@@ -18,6 +18,60 @@ private def attribute(value)
...
+      blank_attribute = attribute("")
+      Avram::Validations.validate_exactly_one_filled(filled_attribute, filled_attribute_2, blank_attribute)
+      filled_attribute.valid?.should be_true
+      blank_attribute.valid?.should be_true
+      filled_attribute_2.errors.should eq(["must be blank"])
+    end
+
+    it "marks first field as invalid if no attributes are filled" do
+      first_blank_attribute = attribute(nil)
+      second_blank_attribute = attribute("")
```
edwardloveall: Just kind of riffing off this in my head, it might be cool to be able to set what it counts as `blank` under the hood, so you could do things like:

```crystal
tos_accepted = attribute("0")
just_browsing = attribute("1")

Avram::Validations.validate_at_most_one_filled(...
```

spec/validations_spec.cr
```diff
@@ -18,6 +18,60 @@ private def attribute(value)
...
+
+      Avram::Validations.validate_exactly_one_filled(first_blank_attribute, second_blank_attribute)
+
+      first_blank_attribute.errors.should eq(["at least one must be filled"])
+      second_blank_attribute.valid?.should be_true
+    end
+
+    it "does not mark any fields as invalid if just one is filled" do
```
edwardloveall: Nit: what about using a positive case instead of negative case for the spec?

> Fields are valid is only one is filled

src/avram/model.cr
```diff
@@ -33,7 +37,11 @@ class Avram::Model
     ASSOCIATIONS = [] of {name: Symbol, foreign_key: Symbol, type: Object, through: Object, relationship_type: Symbol}
   end
 
-  def_equals id
+  def_equals id, model_name
```
edwardloveall: I didn't know about `def_equals`, neat!

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
+# Helpers for defining polymorphic associations
+module Avram::Polymorphic
+  # Sets up helper methods for validating and loading polymorphic associatiions
+  #
+  # This will generate methods for validiating, preloading, and lazy loading
+  # polymorphic associations.
+  #
+  # Let's start with this example:
```
edwardloveall: I think I'd put this below the `Important note:`

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
...
+  #   end
+  # end
+  #
+  # class Photo < BaseModel
+  #   table do
+  #     has_many comments : Comment
+  #   end
+  # end
+  #
+  # class Photo < BaseModel
```
edwardloveall:
```suggestion
  # class Video < BaseModel
```

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
...
+  #   table do
+  #     has_many comments : Comment
+  #   end
+  # end
+  # ```
+  #
+  # `has_many` calls work just like normal and so do `belongs_to`.
```
edwardloveall: For readability, I'd put both association types together up front:

```suggestion
  # `has_many` (and `belongs_to`) work just like normal.
```

Feel free to take or leave the parenthetical. I overuse them 😄

spec/polymorphic_spec.cr
edwardloveall: Yeah the more I think about this the harder it seems. I think I'm just looking for a hint of where to look if I get this error. What about a field error:

```crystal
errors.should eq(["can't be set if `task_id` is also set"])
```

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
...
+  # In this example `commentable` and `commentable!`. Calling these will return
+  # either the `video` or the `photo`. The `!` version will allow lazy loading.
+  # The non `!` version requires preloading with `preload_{polymorphic_name}.
+  #
+  # If `optional` is set to true these methods will return `nil` if nothing
+  # is associated.
+  #
+  # ```
+  # Comment::BaseQuery.new.each do |comment|
+  #   comment.commentable! # Returns the photo or video
```
edwardloveall: I _think_ we'd want them to preload explicitly more often than not, yeah? If so, I'd change this example to use `preload_commentable` and the non-bang method

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
...
+  # If you know which association you want you can just call it directly like
+  # you would with any other `belongs_to`:
+  #
+  # ```
+  # photo.comments.each do |comment|
+  #   # We know it is for a photo, so we can call it directly
+  #   # Note that Crystal will think this might be `nil` since the `belongs_to`
+  #   # is nilable. So you'll need to call `not_nil!` or handle the `nil` case
+  #   # with an `if`
+  #   comment.photo!.not_niL!
```
edwardloveall:
```suggestion
  #   comment.photo!.not_nil!
```

src/avram/polymorphic.cr
```diff
@@ -0,0 +1,144 @@
...
+  #
+  # # Saving/updating polymorphic associations
+  #
+  # Polymorphic associations are saved just like your normally would with a
+  # `belongs_to` association:
+  #
+  # ```
+  # photo = PhotoQuery.first
+  # # Save a comment for this photo
+  # Comment::SaveOperation.create!(photo_id: photo.id)
```
edwardloveall: The setting of the polymorphic model surprised me when I saw it in the tests, so it might be worth showing the example of the other model being set.

src/avram/polymorphic.cr
paulcsmith: Looks good to me!

src/avram/polymorphic.cr
paulcsmith: I blame it on the MacBook Pro keyboard

spec/polymorphic_spec.cr
paulcsmith: It is used in the next line. But maybe I misunderstood?

spec/polymorphic_spec.cr
paulcsmith: I will give it a try!

spec/validations_spec.cr
paulcsmith: Glad you like it!

spec/validations_spec.cr
paulcsmith: This actually is what happens. Since attribute.value is used it will call blank on the parsed type. So if you have a Boolean it’ll use blank on that. Does that make sense?

spec/validations_spec.cr
paulcsmith: I dig it!

src/avram/polymorphic.cr
paulcsmith: Good idea

src/avram/polymorphic.cr
paulcsmith: Whoops!

src/avram/polymorphic.cr
paulcsmith: Yeah that’s a good point. I was going for concise but that’s probably a better idea.

src/avram/polymorphic.cr
paulcsmith: Can you clarify what surprised you and what you’d like to see to make this clearer? I’m not sure I understood

edwardloveall: Ah sorry. I think I wouldn't have assumed that you could associate a Comment with a Photo by setting its ID when it's polymorphic. I think I would have assumed `Comment::SaveOperation.create!(commentable: photo)`. Because you can set either the photo id or the video_id also, it'd love to see another line below:

```crystal
Comment::SaveOperation.create!(video_id: video.id)
```

spec/validations_spec.cr
edwardloveall: It does, but I think I explained my idea poorly. I'm saying that it might parse the value, get `false`, assume that `false.blank?` is "not filled" but you may not want that. However, trying to think of an example I can't actually think of anything that can't be solved better with a custom validation. Here's my contrived example anyway:

```crystal
product_image = attribute("/path/to/default.png") # no imaged uploaded, blank? => false
no_image_checkbox = attribute("0") # evaluates to `false`, blank? => true

validate_exactly_one_filled(product_image, no_image_checkbox) # passes, but shouldn't because there's no image uploaded and the `no image checkbox` wasn't checked.
```

So I'm saying you could define whatever predicate that lets a field pass or fail this validation for the `product_image` field to be `!= "/path/to/default.png"`.

spec/polymorphic_spec.cr
edwardloveall: Oh sorry, missed that. I think it threw me that it was defined twice. Nevermind.

spec/validations_spec.cr
paulcsmith: I think I’m that case you’d want to create some kind of custom type. Like a custom Boolean or a FilePath type so you can change what blank means. That way it changes what blank? Is everywhere. Like most things, creating custom types for columns needs to be documented 😂

src/avram/polymorphic.cr
paulcsmith: Oh I see. So make it clear that’s how you associate them by showing both associations being set, right?

edwardloveall: Yup!

spec/validations_spec.cr
edwardloveall: Gotcha, that makes sense and is a better solution!

spec/polymorphic_spec.cr
paulcsmith: I like this idea a lot but these error messages might be shown to users if you serialize all the errors in a JSON response for example. I think what we need is something like #15 so we can add  more helpful errors like this and display them in a nice way to developers

spec/polymorphic_spec.cr
paulcsmith: Ok so `be_a` is for a single type as it operates on the runtime type and not compile time type. An example probably works best to describe this:

```crystal
def mixed_array
  [1, "foo"]
end

typeof(mixed_array) # Will be Array(String | Int32). typeof works at compile time. Does not actually call the method.
mixed_array.first # Will be String | Int32 at compile time, but at runtime it is one of the two, not *both*
mixed_array.first.should be_a(Int32) # We can only pass one type
```

edwardloveall: Interesting. Thanks for trying!

Issue #166, opened 2019-07-24
Add a new task for checking postgres connection


jwoertink:
This task would only check that a connection can be made using your defined credentials. It would eventually be added to the setup script on the cli to test that a connection can be made prior to running createdb or migrate.

PR #167, opened 2019-07-24
Add lucky db.verify_connection task

jwoertink:
Phase 2 of https://github.com/luckyframework/lucky_cli/pull/376
Fixes https://github.com/luckyframework/avram/issues/166

This PR adds a new task to verify that avram can actually connect to the DB based on your credentials.

Screen Shot 2019-07-24 at 2 33 41 PM

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -5,7 +5,6 @@ require "./support/**"
 require "../config/database"
 
 Db::VerifyConnection.new.call
-
 Db::Create.new(quiet: true).call
 Db::Migrate.new(quiet: true).call
 
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -4,6 +4,8 @@ require "./support/base_model"
 require "./support/**"
 require "../config/database"
 
+Db::VerifyConnection.new.call
+
 Db::Create.new(quiet: true).call
 Db::Migrate.new(quiet: true).call
 
--- a/src/avram/tasks/db/verify_connection.cr
+++ b/src/avram/tasks/db/verify_connection.cr
@@ -7,12 +7,19 @@ class Db::VerifyConnection < LuckyCli::Task
       end
     rescue PQ::ConnectionError | DB::ConnectionRefused
       raise <<-ERROR
-      Unable to connect to postgres.
+      Unable to connect to Postgres for database '#{Avram.settings.database_to_migrate}'.
 
-      Try this...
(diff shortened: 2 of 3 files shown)


spec/spec_helper.cr
```diff
@@ -4,6 +4,8 @@ require "./support/base_model"
 require "./support/**"
 require "../config/database"
 
+Db::VerifyConnection.new.call
+
```
paulcsmith: I think you can get rid of this newline

src/avram/tasks/db/verify_connection.cr
```diff
@@ -0,0 +1,20 @@
+class Db::VerifyConnection < LuckyCli::Task
+  summary "Verify connection to postgres"
+
+  def call
+    begin
+      DB.open(Avram::Migrator::Runner.database_url) do |db|
+      end
+    rescue PQ::ConnectionError | DB::ConnectionRefused
+      raise <<-ERROR
+      Unable to connect to postgres.
```
paulcsmith: I think it'd be good to also print which database it is and/or the url that was used so it is clear what is being used

jwoertink: I have that on the line below it under the second help option. Does that work? If I print the whole URL, that may include the password, does that matter?

paulcsmith: I don't think so since script/setup should only be used in dev, but I suppose it is possible  that someone uses it in prod on accident. Maybe better to be safe than sorry. 

How about we just print the database it is using: https://github.com/luckyframework/avram/blob/3a6e5df4aac9239c71d733163f3cc0948433aeb8/src/avram.cr#L20 `Avram.settings.database_to_migrate`. And maybe we print out each section instead of a psql command, and we just filter out the password. So it'd look something like this:

```
Unable to connect to Postgres for database 'AppDatabase'

Connection info:

  * username: postgres
  * host: whatever
  * password: check config/database.cr

Try this...

  >  Check your database config in config/database.cr and make sure it is correct
  > Then run lucky db.verify_connection to make sure it can connect
```

Thoughts? I believe you can get connection info here https://github.com/luckyframework/avram/blob/3a6e5df4aac9239c71d733163f3cc0948433aeb8/src/avram/migrator/runner.cr#L15-L33

paulcsmith: And maybe include "make sure Postgres is running" 🤣

PR #169, opened 2019-07-25
Adding new delete_all query method to remove records based on query

### LINKED ISSUE #111: Add `delete_all` that deletes all based on the query
Pre-requisite: #110 

```crystal
UserQuery.new.delete_all # Delete the table
UserQuery.new.age.gt(80).delete_all 
```

jwoertink:
Fixes #111 

Based on how rails works, when you run the `delete_all` method, it'll return an integer, so it's going to be the last thing you run in your query. You can chain all your wheres like normal though.

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -590,19 +590,20 @@ describe Avram::Query do
   end
 
   describe "delete_all" do
-    it "uses the glove" do
+    it "deletes user records that are young" do
       UserBox.new.name("Tony").age(48).create
       UserBox.new.name("Peter").age(15).create
       UserBox.new.name("Bruce").age(49).create
       UserBox.new.name("Wanda").age(17).create
 
       ChainedQuery.new.select_count.should eq 4
+      # use the glove to remove half of them
       result = ChainedQuery.new.young.delete_all
       result.should eq 2
       ChainedQuery.new.select_count.should eq 2
     end
 
-    it "finishes what it started" do
+    it "delete all records since no where clause is specified" do
       UserBox.new.name("Steve").age(90).create
       UserBox.new.name("Nick").age(66).create
 
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -588,4 +588,27 @@ describe Avram::Query do
       UserQuery.new.select_count.should eq 0
     end
   end
+
+  describe "delete_all" do
+    it "uses the glove" do
+      UserBox.new.name("Tony").age(48).create
+      UserBox.new.name("Peter").age(15).create
+      UserBox.new.name("Bruce").age(49).create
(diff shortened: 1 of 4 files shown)


spec/query_spec.cr
```diff
@@ -588,4 +588,27 @@ describe Avram::Query do
       UserQuery.new.select_count.should eq 0
     end
   end
+
+  describe "delete_all" do
+    it "uses the glove" do
```
paulcsmith: I’m sure this is a reference I’m not cool enough to understand 😆 could you also add a real description before or after?

src/avram/queryable.cr
```diff
@@ -55,6 +55,13 @@ module Avram::Queryable(T)
     self
   end
 
+  def delete_all : Int64
```
paulcsmith: Could you add some docs for this?

spec/query_spec.cr
jwoertink: 😱

jwoertink:
hmm.... I can see how it may be a bit confusing, though, I do feel like `delete_all` is pretty standard convention.

Ecto: https://hexdocs.pm/ecto/Ecto.Repo.html#c:delete_all/2
Rails: https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all

Though, some do break that
Gorm: http://gorm.io/docs/delete.html#Batch-Delete  uses `Delete`
Laravel: http://gorm.io/docs/delete.html#Batch-Delete uses `->delete()`
Granite: https://github.com/amberframework/granite/blob/c4bc435fc9a24a132ff349367ba2b324d22d8948/docs/crud.md#delete not sure if they have a delete based on WHERE query, but maybe `clear`?
Django: https://docs.djangoproject.com/en/dev/ref/class-based-views/#deleteview also uses `delete`...

so maybe we just call it `delete`?

paulcsmith:
Good call. I think delete makes sense. Other orms use it. It’s what the SQL call uses. I like it!

paulcsmith:
Also I think we should add delete_all and do a compile time raise that tells people to use delete. 

That way if people use it they’ll know what to do

src/avram/queryable.cr
```diff
@@ -55,6 +55,15 @@ module Avram::Queryable(T)
     self
   end
 
+  # returns the number of records removed as `Int64`.
```
paulcsmith: Could you make the first line a summary with a new line after. “This will delete the records using the queries where clauses, or all records if no where’s are added”

Or something like that

I like that it says what it returns but maybe that can go after the part that explains what it does. So tit reads in the order that it executed

src/avram/queryable.cr
paulcsmith: Could you change this to delete? And also the example on line 65

PR #172, opened 2019-07-30
Make querying associations more explicit

### LINKED ISSUE #171: Passing custom queries to association join queries
It seems to be that this doesn't work:

```crystal
class PostQuery < Post::BaseQuery
  def released
    released_at.not.eq(nil)
  end
end

class UserQuery < User::BaseQuery
  def released_posts
    join_posts.where_posts(&.released)
  end
end

# undefined method 'released' for Post::BaseQuery
# join_posts.where_posts(&.released)
#                          ^~~~~~~~~~~
```

Since the instance passed in [the block](https://github.com/luckyframework/avram/blob/master/src/avram/base_query_template.cr#L89) is the `Post::BaseQuery`, and now `PostQuery`. 

ref: https://gitter.im/luckyframework/Lobby?at=5d3f71fac356af25e0500328

paulcsmith:
Closes #171

Another issue has been opened for ensuring joins and wheres are unique before generating a statement https://github.com/luckyframework/avram/issues/175. That should make this a lot more reliable

### DIFF
--- a/spec/query_associations_spec.cr
+++ b/spec/query_associations_spec.cr
@@ -1,5 +1,12 @@
 require "./spec_helper"
 
+# Ensure it works with inherited query classes
+class CommentQuery < Comment::BaseQuery
+  def body_eq(value)
+    body.eq(value)
+  end
+end
+
 describe "Query associations" do
   it "can query associations" do
     post_with_matching_comment = PostBox.create
@@ -16,10 +23,12 @@ describe "Query associations" do
       .post_id(post_without_matching_comment.id)
       .create
 
-    posts = Post::BaseQuery.new.join_comments.where_comments(&.body.eq("matching"))
+    posts = Post::BaseQuery.new
+      .where_comments(CommentQuery.new.body_eq("matching"))
     posts.results.should eq([post_with_matching_comment])
 
-    posts = Post::BaseQuery.new.join_comments.where_comments(&.body("matching"))
+    posts = Post::BaseQuery.new
+      .where_comments(Comment::BaseQuery.new.body("matching"))
     posts.results.should eq([post_with_matching_comment])
   end
 
@@ -38,7 +47,9 @@ describe "Query associations" do
       .post_id(post_without_matching_comment.id)
       .create
 
-    posts = Post::BaseQuery.new.inner_join_comments.where_comments(&.body.eq("matching"))
+    posts = Post::BaseQuery.new
+      .inner_join_comments
+      .where_comments(Comment::BaseQuery.new.body.eq("matching"), auto_inner_join: false)
     posts.to_sql[0].should contain "INNER JOIN"
     posts.results.should eq([post_with_matching_comment])
   end
@@ -58,7 +69,9 @@ describe "Query associations" do
       .post_id(post_without_matching_comment.id)
       .create
 
-    posts = Post::BaseQuery.new.left_join_comments.where_comments(&.body.eq("matching"))
+    posts = Post::BaseQuery.new
+      .left_join_comments
+      .where_comments(Comment::BaseQuery.new.body.eq("matching"), auto_inner_join: false)
     posts.to_sql[0].should contain "LEFT JOIN"
     posts.results.should eq([post_with_matching_comment])
   end
@@ -78,7 +91,9 @@ describe "Query associations" do
       .post_id(post_without_matching_comment.id)
       .create
 
-    posts = Post::BaseQuery.new.right_join_comments.where_comments(&.body.eq("matching"))
+    posts = Post::BaseQuery.new
+      .right_join_comments
+      .where_comments(Comment::BaseQuery.new.body.eq("matching"), auto_inner_join: false)
     posts.to_sql[0].should contain "RIGHT JOIN"
     posts.results.should eq([post_with_matching_comment])
   end
@@ -98,7 +113,9 @@ describe "Query associations" do
       .post_id(post_without_matching_comment.id)
       .create
 
-    posts = Post::BaseQuery.new.full_join_comments.where_comments(&.body.eq("matching"))
+    posts = Post::BaseQuery.new
+      .full_join_comments
+      .where_comments(Comment::BaseQuery.new.body.eq("matching"), auto_inner_join: false)
     posts.to_sql[0].should contain "FULL JOIN"
     posts.results.should eq([post_with_matching_comment])
   end
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -68,7 +68,7 @@ class Avram::BaseQueryTemplate
               )
             {% elsif assoc[:through] %}
               {{ join_type.downcase.id }}_join_{{ assoc[:through].id }}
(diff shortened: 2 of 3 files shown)


paulcsmith:
I'll  add the s pec for just in case. I believe   deeply nested joins work since that is what we're doing for  has_many  through :D

spec/query_associations_spec.cr
```diff
@@ -38,7 +47,9 @@ describe "Query associations" do
       .post_id(post_without_matching_comment.id)
       .create
 
-    posts = Post::BaseQuery.new.inner_join_comments.where_comments(&.body.eq("matching"))
+    posts = Post::BaseQuery.new
+      .inner_join_comments
+      .where_comments(Comment::BaseQuery.new.body.eq("matching"), join: false)
```
edwardloveall: This feels confusing to me. First we join comments then we `join: false`? I'm not sure what this is doing.

paulcsmith: We made it so  the where automatically does an inner join. If you want a different join  you'd  want  to disable that behavior. Maybe there is some other better way to name it?

`auto_join: false` or maybe `skip_join: true`?

edwardloveall: Do you have an easy way to paste what this whole statement is doing in SQL? Maybe that would help me understand what it's doing. I'm not getting how we're joining comments in the query, then later doing a `WHERE` and then specifying joining behavior again (which doesn't seem like it should have anything to do with `WHERE`).

paulcsmith: `where_comments(SomeQuery, automatically_add_join: false)`? Maybe not 100% self explanatory, but easier to google I think

src/avram/base_query_template.cr
```diff
@@ -85,20 +85,30 @@ class Avram::BaseQueryTemplate
         {% end %}
 
 
-        def where_{{ assoc[:table_name] }}
-          {{ assoc[:type] }}::BaseQuery.new_with_existing_query(query).tap do |assoc_query|
-            yield assoc_query
-          end
+        def where_{{ assoc[:table_name] }}(assoc_query : ::{{ assoc[:type] }}::BaseQuery, join auto_join : Bool = true)
+          join_{{ assoc[:table_name] }} if auto_join
```
paulcsmith: @edwardloveall Here is the code we're talking about. Previously you had to do the join yourself or the where would not work: `some_query.join_comments.where_comments(..)`. This makes it so that it calls `join_comments` for you so you don't need to do it every time. The problem is that sometimes you want a different kind of  join (not  an inner join). So in those cases you want to skip  the method automatically adding a join or it will be  the  equivalent of this:  `some_query.join_comments.left_join_comments.where_comments(...)`

paulcsmith: We could revert this behavior and have you explicitly join each time, but I don't like that very much because 99% of the time you want a regular inner join, and you won't get a compile time error if you forget. Laravel does the join for you when querying associations which I think is pretty cool

spec/query_associations_spec.cr
paulcsmith:
```
SELECT posts.custom_id, posts.created_at, posts.updated_at, posts.title, posts.published_at FROM posts INNER JOIN comments ON posts.custom_id = comments.post_id WHERE comments.body = $1
```

But if you did left_join_comments it will do a left_join instead

edwardloveall: Okay, I checked out the branch and am playing around with it and am still confused. All four of these post the exact same SQL:

```crystal
Post::BaseQuery.new
  .where_comments(Comment::BaseQuery.new.body.eq("matching"))

Post::BaseQuery.new
  .inner_join_comments
  .where_comments(Comment::BaseQuery.new.body.eq("matching"))

Post::BaseQuery.new
  .inner_join_comments
  .where_comments(Comment::BaseQuery.new.body.eq("matching"), join: false)

Post::BaseQuery.new
  .where_comments(Comment::BaseQuery.new.body.eq("matching"), join: true
```

So I'm confused why we use both `inner_join_{{assoc}}` _and_ `join: {{Boolean}}`. I think I'm missing something deadly obvious 😅. Or maybe it's besides the point in this branch?

paulcsmith: Ok so part of this is that in SQL a join and inner join are the same  https://softwareengineering.stackexchange.com/questions/206035/join-vs-inner-join-and-full-outer-join

We have both though in case you want to clarify it. Maybe we should remove the inner join method though. It probably just adds to the confusion. 

So let’s ignore the inner join thing for now. The main reason you’d use “join: false” is if you’re doing a left, right, or full join (or some other custom join) that you want to use. In that case you don’t want Avram to add an additional inner join so you use “join:false” or whatever we decide to call that option

edwardloveall: Ah ha! So where_{{assoc}} does an automatic _inner_ join for you and you can tell it _not_ to do that because you want to do some other kind of join with `join: false`. Is that right?

edwardloveall: Would it be a reach for this PR to instead do `join: INNER` or `join: NONE` or `join: LEFT` or something?

paulcsmith: Yep you nailed it!

I think that the problem there is that now we’d introduce more ways to join stuff. You can use the join methods or this other kind of option. It’s also be trickier to make type safe as well

I think we definitely need a better name for this option though. Maybe “add_join: false” or automatically_add_join: false?

A bit long but I think this won’t be used much so clarity is probably more important than keeping it super short. 

Thoughts @jwoertink?

edwardloveall: For the record, I think what threw me into this confusion was the spec. It has the `inner_join_comments` and `join: false`. I think if it hadn't had the `inner_join_comments` I would have been much less confused.

`auto_join: false`?

paulcsmith: Did the left_join and  right_join ones  not  confuse you? https://github.com/luckyframework/avram/pull/172/files#diff-54c72b6e1799559414b793e3fa88bc3eR73

Just  wondering if  it  was specifically the inner_join (because I think we should remove it to reduce confusion)

paulcsmith: And I think `auto_join` works!

edwardloveall: Yeah, I think they are confusing also. It's the join twice that's weird. It might make me think that I'm canceling the left_join

edwardloveall: `auto_inner_join: false` might be even better if we can stand the length.

paulcsmith: Yeah I can definitely see that.  Like `join: false` is removing the join completely or something. `auto_inner_join` I *love*. Nice and  clear  that it is  one specific type of join we are not doing!

edwardloveall: Yay, we did it! ✋

paulcsmith:
@edwardloveall Do I get a 👍? :D

paulcsmith:
Woo hoo! Thanks for taking a look. I'm glad we talked through that join stuff. The new argument name is WAY better

PR #176, opened 2019-07-30
Update for Crystal v0.30.0

bcardiff:
Hi, this is a partial update to Crystal 0.30.0

I am a bit lost in the metamodel of avram regarding the last two commits where I am trying to make the "Add column getter return type" commit work. I didn't follow all the type information of column, attributes, etc. Probably I am missing something obvious. Any clue?

Since the return types will be required and `Authentic::PasswordAuthenticatable#encrypted_password` has return type I was trying to set the return types for every column.

Note: Some dependencies need to be updated also, my `bcardiff/use-fork` branch has those overrided.

### DIFF
--- a/src/avram/model.cr
+++ b/src/avram/model.cr
@@ -158,11 +158,13 @@ class Avram::Model
 
   macro setup_getters(columns, *args, **named_args)
     {% for column in columns %}
-      def {{column[:name]}}
-        {% if column[:type].is_a?(Generic) %}
-          {{column[:type].type_vars.first}}::Lucky.from_db! @{{column[:name]}}
+      {% db_type = column[:type].is_a?(Generic) ? column[:type].type_vars.first : column[:type] %}
+      def {{column[:name]}} : {% if column[:nilable] %}::Union({{db_type}}, ::Nil){% else %}{{column[:type]}}{% end %}
+        %from_db = {{ db_type }}::Lucky.from_db!(@{{column[:name]}})
+        {% if column[:nilable] %}
+          %from_db.as?({{db_type}})
         {% else %}
-          {{ column[:type] }}::Lucky.from_db! @{{column[:name]}}
+          %from_db.as({{column[:type]}})
         {% end %}
       end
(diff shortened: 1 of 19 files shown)


jwoertink:
Dude! Thanks for submitting this. We appreciate it. I totally get how it gets a bit hairy in there 😂 . So all methods will require a return type now?

bcardiff:
The implementation of abstract methods with return type will need to honor the return type explicitly. It can be a subclass though.

But since a User model somewhere includes `Authentic::PasswordAuthenticatable` ... either the method there need to remove the return type, or the return types of all getters should be expanded. The later sounds reasonable, but I would need some input, or someone to take if from here. 

I am testing as much as possible before the release, but it got... _hairy_ 😅

src/avram/model.cr
```diff
@@ -158,11 +158,13 @@ class Avram::Model
 
   macro setup_getters(columns, *args, **named_args)
     {% for column in columns %}
-      def {{column[:name]}}
-        {% if column[:type].is_a?(Generic) %}
-          {{column[:type].type_vars.first}}::Lucky.from_db! @{{column[:name]}}
+      {% db_type = column[:type].is_a?(Generic) ? column[:type].type_vars.first : column[:type] %}
+      def {{column[:name]}} : {% if column[:nilable] %}::Union({{db_type}}, ::Nil){% else %}{{db_type}}{% end %}
```
jwoertink: This one is a bit tricky. `column[:type]` could be `Array(String)`, but `db_type` would be `String`. So I think we could do 

```crystal
def {{column[:name]}} : {% if column[:nilable] %}::Union({{db_type}}, ::Nil){% else %}{{column[:type]}}{% end %}
```

But the `db_type` is still ok, because we need it on the next line so `Array(String)` maps to `String::Lucky.from_db`

paulcsmith:
Thank you so dang much! This is super helpful

paulcsmith:
Closing in favor of  #198 because  this needed  a rebase

PR #180, opened 2019-07-31
Add lucky db.create to verification check

paulcsmith:


src/avram/tasks/db/verify_connection.cr
```diff
@@ -21,6 +21,7 @@ class Db::VerifyConnection < LuckyCli::Task
 
         ▸ Make sure Postgres is running.
         ▸ Check connection settings in config/database.cr.
+        ▸ Make sure the database is created with 'lucky db.create'
```
jwoertink: This is a little tricky. At first it reads as if you want me to run that task to make sure the database exists. So then you do run it just in case, and you get 

```
[08:29AM] avram (master)$ lucky db.create
Unhandled exception: createdb: could not connect to database template1: could not translate host name "db" to address: nodename nor servname provided, or not known
 (Exception)
```

What about:

> Ensure the database exists. If not, use 'lucky db.create'.

paulcsmith: I like that!

jwoertink: I'm also cool with just leaving as you have it and feel it out. We can always update later if people get confused.

src/avram/tasks/db/verify_connection.cr
```diff
@@ -21,6 +21,7 @@ class Db::VerifyConnection < LuckyCli::Task
 
         ▸ Make sure Postgres is running.
         ▸ Check connection settings in config/database.cr.
+        ▸ If the database has not been created yet, run 'lucky db.create'
```
paulcsmith: How does that look @jwoertink?

jwoertink: ah yup! I think that reads a lot more clear. Thanks!

PR #187, opened 2019-07-31
Add is_nil and is_not_nil

### LINKED ISSUE #184: Query for things that are not nil looks weird.
If you want to query with `IS NOT NULL` your first instinct might be to do `thing.not.eq(nil)`, but that throws a compile-time error saying you should use `nilable_eq` instead. So now your query looks like `thing.not.nilable_eq(nil)`. I think we can do better about how that looks. 

Maybe  `is_nil` and `is_not_nil` methods so we get `thing.is_nil` or `thing.is_not.nil` for a type-safe `IS NULL` or `IS NOT NULL`. Another option could be to make `eq` allow a `nil` literal value, but still throw the error for things that may be nilable.

paulcsmith:
With fancy error handling

Screen Shot 2019-07-31 at 5 00 24 PM

Also clarified the  other error  message:

Screen Shot 2019-07-31 at 5 06 32 PM

Closes #184

### DIFF
--- a/src/avram/criteria.cr
+++ b/src/avram/criteria.cr
@@ -125,7 +125,8 @@ class Avram::Criteria(T, V)
     add_clause(Avram::Where::In.new(column, values))
   end
 
-  def distinct_on
+  # :nodoc:
+  def __distinct_on : Avram::QueryBuilder
     rows.query.distinct_on(column)
   end
 
--- a/src/avram/criteria.cr
+++ b/src/avram/criteria.cr
@@ -6,37 +6,77 @@ class Avram::Criteria(T, V)
     @negate_next_criteria = false
   end
 
-  def desc_order
+  def desc_order : T
     rows.query.order_by(column, :desc)
     rows
   end
 
-  def asc_order
+  def asc_order : T
     rows.query.order_by(column, :asc)
     rows
   end
 
-  # :nodoc:
-  def eq(_value : Nil)
-    {{ raise "Use `nilable_eq` instead of `eq` if trying to check against a nilable value." }}
-  end
-
-  def eq(value)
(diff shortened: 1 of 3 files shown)


src/avram/criteria.cr
```diff
@@ -16,29 +16,70 @@ class Avram::Criteria(T, V)
...
     add_clause(Avram::Where::Equal.new(column, V::Lucky.to_db!(value)))
     rows
   end
 
   def nilable_eq(value)
+    check_just_nil!(typeof(value))
+
     if value.nil?
-      add_clause(Avram::Where::Null.new(column))
-      rows
```
jwoertink: what was `rows` doing here, and is it still needed?

paulcsmith: I'm not sure if it is still needed. I'm getting a spec failure that may be related. Investiigating that now

paulcsmith: Turns out this was being subtly done by the `add_clause` method which returned `rows`. Instead I've removed that and explicitly return `rows` now. That's what allows chaining criteria

paulcsmith:
Ok should be good to go now! Let me know what you think

src/avram/criteria.cr
paulcsmith: Cancel that. This breaks all the other criteria. This will be a pain to always add so I'll change it back lol

src/avram/criteria.cr
```diff
@@ -6,39 +6,79 @@ class Avram::Criteria(T, V)
     @negate_next_criteria = false
   end
 
-  def desc_order
+  def desc_order : T
```
paulcsmith: I decided to add return types while I'm in here

PR #190, opened 2019-08-01
merge in orders for queries

### LINKED ISSUE #189: Ordering by a column on a joined table
Along with https://github.com/luckyframework/avram/pull/174 we also need to merge orders.

This would allow for something like

```crystal
UserQuery.new.where_tasks(TaskQuery.new.created_at.desc_order)
```

```sql
SELECT * FROM users
INNER JOIN tasks ON users.id = tasks.user_id
ORDER BY tasks.created_at DESC
```

jwoertink:
Fixes #189 

This allows us to order on a column in a joined table:

```crystal
# ORDER BY comments.created_at ASC
Post::BaseQuery.new.where_comments(Comment::BaseQuery.new.created_at.asc_order)
```

### DIFF
--- a/spec/query_builder/merge_spec.cr
+++ b/spec/query_builder/merge_spec.cr
@@ -16,6 +16,16 @@ describe "Avram::QueryBuilder#merge" do
     query_1.statement.should eq "SELECT * FROM users INNER JOIN posts ON users.id = posts.user_id INNER JOIN tasks ON users.id = tasks.user_id WHERE age = $1 AND age = $2 AND name = 'Mary' AND name = 'Greg'"
     query_1.args.should eq ["42", "20"]
   end
+
+  it "merges orders" do
+    query_1 = new_query
+      .order_by(:id, :asc)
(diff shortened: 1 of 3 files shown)


paulcsmith:
I think merging the order in a method called “where” is a bit unexpected. I’m worried that someone will set a default order on an association query or just not realize it’ll be merged in and then get unexpected results. 

What do you think about a new method called “merge_order_by” on the query builder and then a new method called “order_by_{{assoc}}(SomeQuery.new.newest_first)?

Or maybe I’m overthinking this and we just need to document that it merges orders! And have some kind of .clear_order that can be called on the query so you can clear it if you don’t want it in the query

paulcsmith:
Could you see what ActiveRecord#merge does? I wonder if it merges the order

paulcsmith:
Actually I think it makes sense to merge the order like you have in this PR but we need a “reset_order” or “clear_order” method so there is an escape hatch if you need it

PR #192, opened 2019-08-01
Reset order

### LINKED ISSUE #191: Add reset_order to Queryable and QueryBuilder

paulcsmith:
Closes #191 

This also makes sure order bys are unique


paulcsmith:
Wow that was fast :D

jwoertink:
I was already digging around and just happen to catch it right when you pushed lol

PR #201, opened 2019-08-05
Print "Done" when connection is verified

### LINKED ISSUE #182: db.verify_connection should say "Success" if things worked
https://github.com/luckyframework/avram/blob/master/src/avram/tasks/db/verify_connection.cr

Right now it prints nothing, which makes it unclear if it failed or not. We may want a `quiet` option like migrations and db.create have that can be used in spec_helper

paulcsmith:
Closes #182


jwoertink:
Looks good. Thanks

PR #209, opened 2019-08-13
Add change_type

### LINKED ISSUE #206: Helper method in migrations to change the collumn type
This came up because we're switching `Int32` ids to `Int64`, `int4` and `int8` in postgres nomenclature respectively. We could have folks set the type of their old primary keys to `Int32`, but then all tables going forward by default would have a different primary key type `Int64`. In addition, new tables that have foreign keys that point to these old tables would have a mixed `Int32` foreign keys and an `Int64` primary key. No good.

The proposed solution is to have people run a migration that changes all keys to `Int64`. Currently, that would have to look something like:

```crystal
class ConvertInt32KeysToInt64::V20190810171432 < Avram::Migrator::Migration::V1
  def migrate
    execute "ALTER TABLE users ALTER COLUMN id SET DATA TYPE bigint"
    execute "ALTER TABLE posts ALTER COLUMN id SET DATA TYPE bigint"
    execute "ALTER TABLE posts ALTER COLUMN user_id SET DATA TYPE bigint"
  end

  def rollback
    execute "ALTER TABLE users ALTER COLUMN id SET DATA TYPE int"
    execute "ALTER TABLE posts ALTER COLUMN id SET DATA TYPE int"
    execute "ALTER TABLE posts ALTER COLUMN user_id SET DATA TYPE int"
  end
end
```

But we think we could make something better. Maybe something like:

```crystal
class ConvertInt32KeysToInt64::V20190810171432 < Avram::Migrator::Migration::V1
  def migrate
    change_column_type :posts, :id, Int64
    change_column_type :posts, :user_id, Int64
    change_column_type :users, :id, Int64
  end

  def rollback
    change_column_type :posts, :id, Int32
    change_column_type :posts, :user_id, Int32
    change_column_type :users, :id, Int32
  end
end
```

paulcsmith:
Closes #206

### DIFF
--- a/db/migrations/20190723233131_test_change_type.cr
+++ b/db/migrations/20190723233131_test_change_type.cr
@@ -1,0 +1,35 @@
+class TestChangeType::V20190723233131 < Avram::Migrator::Migration::V1
+  # Test that we can save and query after type changes
+  class TempUserInt64 < Avram::Model
+    skip_default_columns
+
+    def self.database
+      TestDatabase
+    end
+
+    table :users do
+      primary_key id : Int64
+      column name : String
+      column age : Int32
+      column joined_at : Time
+      timestamps
+    end
+  end
+
+  def migrate
+    TempUserInt64::SaveOperation.create!(name: "foo", age: 0, joined_at: Time.now)
+
+    alter table_for(User) do
+      change_type id : Int32
+    end
+
+    alter table_for(User) do
+      change_type id : Int64
+    end
+
+    TempUserInt64::BaseQuery.first # should not raise
+  end
+
+  def rollback
+  end
+end
--- a/spec/migrator/alter_table_statement_spec.cr
+++ b/spec/migrator/alter_table_statement_spec.cr
@@ -48,6 +48,22 @@ describe Avram::Migrator::AlterTableStatement do
     built.statements[6].should eq "ALTER TABLE users ALTER COLUMN updated_at SET NOT NULL;"
   end
 
+  it "does not build statements if nothing is altered" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build { }
+    built.statements.size.should eq 0
+  end
+
+  it "can change column types" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      change_type id : Int64
+      change_type age : Float64, precision: 1, scale: 2
+    end
+
+    built.statements.size.should eq 2
+    built.statements[0].should eq "ALTER TABLE users ALTER COLUMN id SET DATA TYPE bigint;"
+    built.statements[1].should eq "ALTER TABLE users ALTER COLUMN age SET DATA TYPE decimal(1,2);"
+  end
+
   describe "associations" do
     it "can create associations" do
       built = Avram::Migrator::AlterTableStatement.new(:comments).build do
--- a/src/avram/migrator/alter_table_statement.cr
+++ b/src/avram/migrator/alter_table_statement.cr
@@ -6,8 +6,26 @@ class Avram::Migrator::AlterTableStatement
   getter rows = [] of String
   getter dropped_rows = [] of String
   getter fill_existing_with_statements = [] of String
+  getter change_type_statements = [] of String
 
   def initialize(@table_name : Symbol)
+  end
+
+  macro change_type(type_declaration, **type_options)
+    {% if !type_declaration.is_a?(TypeDeclaration) %}
+      {% type_declaration.raise "Must pass a type declaration to 'change_type'. Example: change_type age : Int32" %}
+    {% end %}
+    %column = ::Avram::Migrator::Columns::{{ type_declaration.type }}Column({{ type_declaration.type }}).new(
+      name: {{ type_declaration.var.stringify }},
+      nilable: false,
+      default: nil,
+      {{ **type_options }}
+    )
+    add_change_type_statement %column
+  end
+
+  def add_change_type_statement(column : ::Avram::Migrator::Columns::Base)
+    change_type_statements << column.build_change_type_statement(@table_name)
   end
 
   # Accepts a block to alter a table using the `add` method. The generated sql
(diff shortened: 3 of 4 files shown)


db/migrations/20190723233131_test_change_type.cr
```diff
@@ -0,0 +1,14 @@
+class TestChangeType::V20190723233130 < Avram::Migrator::Migration::V1
+  def migrate
+    alter table_for(User) do
+      change_type id : Int32
+    end
+
+    alter table_for(User) do
+      change_type id : Int64
```
paulcsmith: Integration test these by changing it to Int32 and back

spec/migrator/alter_table_statement_spec.cr
```diff
@@ -48,6 +48,22 @@ describe Avram::Migrator::AlterTableStatement do
...
 
+  it "does not build statements if nothing is altered" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build { }
+    built.statements.size.should eq 0
+  end
+
+  it "can change column types" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      change_type id : Int64
+      change_type age : Int32
```
paulcsmith: Decided to just make it part of `alter` since we do that for indexes and `fill_existing_with`

src/avram/migrator/alter_table_statement.cr
```diff
@@ -6,10 +6,28 @@ class Avram::Migrator::AlterTableStatement
...
   end
 
+  macro change_type(type_declaration, **type_options)
+    {% if !type_declaration.is_a?(TypeDeclaration) %}
+      {% type_declaration.raise "Must pass a type declaration to 'change_type'. Example: change_type age : Int32" %}
+    {% end %}
+    %column = ::Avram::Migrator::Columns::{{ type_declaration.type }}Column({{ type_declaration.type }}).new(
+      name: {{ type_declaration.var.stringify }},
+      nilable: false,
+      default: nil,
```
paulcsmith: This is probably a smell, but I can try to tackle the refactor later

src/avram/migrator/alter_table_statement.cr
```diff
@@ -36,14 +54,19 @@ class Avram::Migrator::AlterTableStatement
...
+    alter_statements + change_type_statements + index_statements + fill_existing_with_statements
   end
 
-  def alter_statement
-    String.build do |statement|
-      statement << "ALTER TABLE #{@table_name}"
-      statement << "\n"
-      statement << (rows + dropped_rows).join(",\n")
+  def alter_statements : Array(String)
+    if (rows + dropped_rows).empty?
```
paulcsmith: If there  are no alter statements then we should generate an empty `ALTER` or it will error

spec/migrator/alter_table_statement_spec.cr
```diff
@@ -48,6 +48,22 @@ describe Avram::Migrator::AlterTableStatement do
...
 
+  it "does not build statements if nothing is altered" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build { }
+    built.statements.size.should eq 0
+  end
+
+  it "can change column types" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      change_type id : Int64
+      change_type age : Float64, precision: 1, scale: 2
```
paulcsmith: Because we use the Column statements, we get the options for free

src/avram/migrator/alter_table_statement.cr
```diff
@@ -6,10 +6,28 @@ class Avram::Migrator::AlterTableStatement
...
   getter dropped_rows = [] of String
   getter fill_existing_with_statements = [] of String
+  getter change_type_statements = [] of String
 
   def initialize(@table_name : Symbol)
   end
 
+  macro change_type(type_declaration, **type_options)
+    {% if !type_declaration.is_a?(TypeDeclaration) %}
+      {% type_declaration.raise "Must pass a type declaration to 'change_type'. Example: change_type age : Int32" %}
```
edwardloveall: How about

```suggestion
      {% type_declaration.raise "Must pass a type declaration to 'change_type'. Example: change_type user_id : Int64" %}
```

Because it's likely be a more common example of what people are doing

src/avram/migrator/alter_table_statement.cr
```diff
@@ -6,10 +6,28 @@ class Avram::Migrator::AlterTableStatement
...
+    {% end %}
+    %column = ::Avram::Migrator::Columns::{{ type_declaration.type }}Column({{ type_declaration.type }}).new(
```
edwardloveall: Yeah it would have taken me weeks to figure this one out

src/avram/migrator/alter_table_statement.cr
paulcsmith: I dig it!

paulcsmith:
I think we’ll be ok without that spec since Postgres will raise if it can’t do the type change. For example if you try to change a string to an int it will raise and ask you to tell it how to do the conversion. 

So if we add specs we’d kind of be testing Postgres I think. Thoughts?

paulcsmith:
And thanks for the fast review!!

src/avram/migrator/alter_table_statement.cr
paulcsmith: I’ll try to see if I can make this type of change easier. I’ve got some ideas I’ll try out later!

edwardloveall:
That’s true. I think the test would be us monitoring if postgres changes. I don’t think it’s an unproductive test to write but I see why we wouldn’t. I think if I knew how to do it and it was easy I’d write it. Otherwise not.

paulcsmith:
@edwardloveall I'll give it a shot! I'm not 100% sure how to do it but maybe something will come to mind

paulcsmith:
@jwoertink We let them deal with it :P There isn't much we can do to handle all those cases I think so they'd have to figure out how to handle it

db/migrations/20190723233131_test_change_type.cr
```diff
@@ -0,0 +1,35 @@
...
+
+    alter table_for(User) do
+      change_type id : Int32
+    end
+
+    alter table_for(User) do
+      change_type id : Int64
+    end
+
+    TempUserInt64::BaseQuery.first # should not raise
```
paulcsmith: I think this should cover it. We change it and then back and then run a query that would raise if the row is missing or id was not changed correctly

paulcsmith:
Ok I think this covers everything! I'll merge once this is green and then we can update the UPGRADE_NOTES so show how to use this to update primary and foreign keys

PR #214, opened 2019-08-15
add ability to clone a query. 

### LINKED ISSUE #208: Unable to fork queries / need a dup method.
Let's say you have this:

```crystal
# grab a list of tasks
tasks = TaskQuery.new.assigned_at.asc_order

# Now you want to grab 2 specific tasks
recently_completed_task = tasks.completed_at.is_not_nil.desc_order.first
current_working_task = tasks.started_at.lte(Time.utc).first
```

You created your initial query, then forked it in to 2 separate queries. This will throw the error:

> bind message supplies 3 parameters, but prepared statement "" requires 2 (PQ::PQError)

If you got rid of the `first` method call on the completed tasks, you're still left with the fact that the second query will still contain `WHERE completed_at IS NOT NULL` which in this case you wouldn't want.

I tried adding a `dup` like `tasks.dup.completed_at.is_not_nil`, but that didn't work. I think adding a `dup` method to Query would solve this though. 

Related: #194 #195

jwoertink:
Fixes #208

This adds a `clone` method to queries so you can do things like fork a query and not have it mutate the original query.

```crystal
original_query = UserQuery.new.admin(true)
new_query = original_query.clone.last_name.desc_order

original_query # does not order by last name
```

I first called this method `dup` because that made most sense, but it seems there's already a `dup` method available that doesn't really work how you'd think.

### DIFF
--- a/src/avram/queryable.cr
+++ b/src/avram/queryable.cr
@@ -44,6 +44,13 @@ module Avram::Queryable(T)
       .select(@@schema_class.column_names)
   end
 
+  def clone : self
+    original_query = query
+    instance = self.class.new
+    instance.query.merge(original_query)
+    instance
+  end
+
   def distinct : self
     query.distinct
     self
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -626,4 +626,40 @@ describe Avram::Query do
       query.to_sql[0].should contain "ORDER BY comments.created_at ASC"
     end
   end
+
+  describe "cloning queries" do
+    it "leaves the original query unaffected" do
+      original_query = ChainedQuery.new.young
+      new_query = original_query.clone.named("bruce wayne").joined_at.asc_order
+
+      original_query.to_sql.size.should eq 2
+      new_query.to_sql.size.should eq 3
+      new_query.to_sql[0].should contain "ORDER BY users.joined_at"
+      original_query.to_sql[0].should_not contain "ORDER BY"
+    end
+
+    it "returns separate results than the original query" do
+      UserBox.new.name("Purcell").age(22).create
+      UserBox.new.name("Purcell").age(84).create
+      UserBox.new.name("Griffiths").age(55).create
+      UserBox.new.name("Griffiths").age(75).create
+
+      original_query = ChainedQuery.new.named("Purcell")
+      new_query = original_query.clone.age.gt(30)
+
+      original_query.select_count.should eq 2
+      new_query.select_count.should eq 1
+    end
+
+    it "clones joined queries" do
+      post = PostBox.new.create
+      CommentBox.new.post_id(post.id).create
+
+      original_query = Post::BaseQuery.new.where_comments(Comment::BaseQuery.new.created_at.asc_order)
+      new_query = original_query.clone.select_count
+
+      original_query.first.should_not eq nil
+      new_query.should eq 1
+    end
+  end
 end


src/avram/queryable.cr
```diff
@@ -44,6 +44,13 @@ module Avram::Queryable(T)
       .select(@@schema_class.column_names)
   end
 
+  def clone : self
+    original_query = query
+    instance = self.class.new
+    instance.query.merge(original_query)
```
paulcsmith: I think instead of query.merge we should  have  query.clone that basically  does a merge but merges everything. Selects, orders, limits, etc. Thoughts?

jwoertink: That's a good idea.

src/avram/query_builder.cr
```diff
@@ -196,7 +212,13 @@ class Avram::QueryBuilder
...
   end
 
-  def select(selection : Array(Symbol))
+  def selects
+    @selections
+      .split(", ")
+      .map { |column| column.split(".").last }
+  end
+
+  def select(selection : Array(Symbol | String))
```
jwoertink: I didn't like this change, but Crystal has no way to turn a String back in to a Symbol, so we have to be able to pass in `Array(String)`.

src/avram/query_builder.cr
```diff
@@ -110,11 +118,19 @@ class Avram::QueryBuilder
     @distinct || @distinct_on
   end
 
+  def limit
+    @limit
+  end
```
jwoertink: This (as well as the offset) could probably just be `getter` macros, but for consistency sake, I just added them as methods.

spec/query_builder_spec.cr
```diff
@@ -241,6 +241,27 @@ describe Avram::QueryBuilder do
       query.ordered?.should eq false
     end
   end
+
+  describe "clone" do
+    it "copies over all parts of a query" do
+      old_query = new_query
+        .select([:name, :age])
+        .join(Avram::Join::Inner.new(:users, :posts))
+        .where(Avram::Where::Equal.new(:name, "Paul"))
```
paulcsmith: Does it also copy raw_wheres? One more thing...I'm thinking we should automatically clone the query every time so that people don't even need to think  about it. It's just always a fresh copy. I think this could be done here in Queryable: 

Here's what it is now:

https://github.com/luckyframework/avram/blob/178e3e3dd282355e4d7b7b287bf0a70625ed0bbd/src/avram/queryable.cr#L41-L45

It could be now:

```
  def query
    # Don't reuse query, clone it
    if @query
      @query  = @query.clone
    else # If no query, create it
      @query = Avram::QueryBuilder
      .new(table: @@table_name)
      .select(@@schema_class.column_names)
     end
  end
```

jwoertink: Yes, it does copy the raw_wheres because it calls `merge` first. I actually tried that bit from your branch you pushed up, but the specs wouldn't pass. I'll have to try again because I don't remember the error, but provided the specs are all correct, then I will get it to work.

paulcsmith: Ok awesome! And  yeah that was total untested code so maybe it won't work, but if we can that'd be awesome

jwoertink: ok, yeah, this isn't working the way I'd hope. It's the first branch that gets me.

```crystal
# clone requires an arg
@query = @query.clone(what_would_go_here?)

# using `dup` instead causes nil errors all over.
@query = @query.dup
```

In the first one, we can't just pass the same query in because it wouldn't do anything. We'd have to create a new instance for it to work. That's what the `clone` method is already doing, so it's just double the work. Maybe you can see something I'm missing?

PR #215, opened 2019-08-15
update Box.create_pair to now return an array of the instance. 

### LINKED ISSUE #199: `Box#create_pair` should return the records and allow overriding defaults
```crystal
admins = UserBox.create_pair do |user|
  user.admin(true)
end
```

jwoertink:
Fixes #199

Also added method overload for accepting a block. 

Before this PR, using `Box.create_pair` always returned `nil`, and there was no way to customize the boxes being created. Now it returns `[model, model]`. You can also pass a block in to customize it:

```crystal
users = UserBox.create_pair do |user_box|
  user_box.name(user_box.sequence("Jeremy"))
end

users.first.name #=> Jeremy-1
users.last.name #=> Jeremy-2
```

### DIFF
--- a/spec/box_spec.cr
+++ b/spec/box_spec.cr
@@ -1,19 +1,37 @@
 require "./spec_helper"
 
-describe "Sequences" do
-  it "increases a value every time it's called" do
-    BaseBox::SEQUENCES["name"] = 0
+describe Avram::Box do
+  describe "Sequences" do
+    it "increases a value every time it's called" do
+      BaseBox::SEQUENCES["name"] = 0
 
-    tag1 = TagBox.create
-    tag2 = TagBox.create
+      tag1 = TagBox.create
+      tag2 = TagBox.create
 
-    tag1.name.should eq("name-1")
-    tag2.name.should eq("name-2")
+      tag1.name.should eq("name-1")
+      tag2.name.should eq("name-2")
+    end
+
+    it "can be overridden" do
+      tag = TagBox.create(&.name("not-a-sequence"))
+
+      tag.name.should eq("not-a-sequence")
+    end
   end
 
-  it "can be overridden" do
-    tag = TagBox.create(&.name("not-a-sequence"))
+  describe "create_pair" do
+    it "creates 2 tags" do
+      tags = TagBox.create_pair
+      tags.size.should eq 2
+      tags.class.name.should eq "Array(Tag)"
+    end
 
-    tag.name.should eq("not-a-sequence")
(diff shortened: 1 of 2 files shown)


spec/box_spec.cr
```diff
@@ -1,19 +1,37 @@
...
+    it "creates 2 tags" do
+      tags = TagBox.create_pair
+      tags.size.should eq 2
+      tags.class.name.should eq "Array(Tag)"
+    end
 
-    tag.name.should eq("not-a-sequence")
+    it "yields the block to both boxes" do
+      tags = TagBox.create_pair do |box|
+        box.name(box.sequence("new-tag"))
```
paulcsmith: I wonder if  UserBox might be better here so you could also do `box.age(30)` and check that both records have 30?

src/avram/box.cr
```diff
@@ -36,8 +36,36 @@ abstract class Avram::Box
...
+  #
+  # ```crystal
+  # tags = TagBox.create_pair
+  # typeof(tags) # => Array(Tag)
+  # tags.size    # => 2
+  # ```
+  def self.create_pair
+    [1, 2].map do |n|
+      new.create
+    end
```
paulcsmith: I think this should call `create_pair { }` so that the implementation is DRYed up a bit. Thoughts?

src/avram/box.cr
```diff
@@ -36,8 +36,36 @@ abstract class Avram::Box
...
+  end
+
+  # Similar to `create_pair`, but accepts a block which yields the box instance.
```
paulcsmith: I think a newline in  here would be good so it separates the summary line from the rest of the docs

src/avram/box.cr
```diff
@@ -36,8 +36,36 @@ abstract class Avram::Box
...
+  #
+  # ```crystal
+  # TagBox.create_pair do |box|
+  #   # set both boxes name to "test"
+  #   box.name("test")
+  # end
+  # ```
   def self.create_pair
-    2.times { new.create }
+    [1, 2].map do |n|
```
paulcsmith: Clever! :D

src/avram/box.cr
jwoertink: I originally did that and ran in to a weird issue passing the block around. I'd rather go that route though, so I'll give it another shot and post what issue I run in to.

spec/box_spec.cr
jwoertink: why not both

PR #216, opened 2019-08-16
Added 2 new tasks for importing and exporting sql dump files. 

### LINKED ISSUE #168: Add new tasks for importing and exporting schema
Postgres comes with 2 handy tools `pg_dump` and `pg_restore`. These allow you to dump out your entire database to sql, or you can specify just the schema itself. Then you can drop a database, and restore using that dump.

I have a need for this where I don't have migrations because they were all created in rails, but I have the schema dump. I'd like to import that schema, then run specs against it. I imagine we can create 2 new tasks to handle this.
`lucky db.schema.import "/path/to/some.sql"` and `lucky db.schema.export` (generates lucky_export_2019000000.sql)

jwoertink:
Fixes #168 

This adds in 2 new tasks. 
* `lucky db.schema.restore path/to/sql_dump.sql`
* `lucky db.schema.dump`

These allow you to task a sql dump file and import your schema in the case you don't have migration files to run. Or you can create a dump file from your current schema that you can use to import later.


jwoertink:
oh yeah, what did you want to do about that? Should I just get specs running outside of Docker? Or should I update the Dockerfile to install the latest pg client stuff?

paulcsmith:
@jwoertink Whichever is easiest. If you wanna remove Docker and just use Crystal on Travis then I'm cool with that too!

spec/tasks/db_schema_dump_spec.cr
```diff
@@ -0,0 +1,12 @@
+require "../spec_helper"
+
+describe Db::Schema::Dump do
+  it "generates a new sql dump file" do
+    Db::Schema::Dump.new.call
+
+    filename = Dir.entries(Dir.current).find { |e| e =~ /lucky_avram_dev_dump/ }.to_s
+    File.exists?(filename).should eq true
+    File.read(filename).should contain "PostgreSQL database dump"
+    File.delete(filename)
```
paulcsmith: This could cause some issues where if File.read(filename) fails the  file would not be deleted. What if instead you add a new arg to `Dump.new` that accepts a filename. It would default to `db/structure.sql` but for tests could be configured to `tmp/structure.sql`. Could also accept an ARGV.first for a custom location. Then there could be a `Spec.before_each` that deletes all the files in `tmp`. That way we can be sure it is always clean. This would also let us ignore `/tmp` in gitignore instead of ignoring a file pattern in the project directory

I think the initialize method would look something like this:

```crystal
def initialize(@dump_to : String)
end

def initialize
  @ dump_to = ARGV.first? || raise "Must pass a path to dump the db structure to"
end
```

Something like that I think. Thoughts?

src/avram/migrator/runner.cr
```diff
@@ -95,6 +95,44 @@ class Avram::Migrator::Runner
...
+      ERROR
+    elsif message.includes?("Connection refused")
+      raise <<-ERROR
+      Unable to connect to db #{db_name}.
+
+      Try this...
+
+        ▸ Check your settings in 'config/database.cr'.
+        ▸ Run 'luck db.verify_connection' to ensure you can connect
+      ERROR
```
paulcsmith: Maybe run the verifiy connection task before dumping so you don't need these. I think that task handles most of this

src/avram/tasks/db/schema_restore.cr
```diff
@@ -0,0 +1,13 @@
+class Db::Schema::Restore < LuckyCli::Task
+  summary "Restore database from a sql dump file"
+
+  def initialize(@import_file : String? = nil, @quiet : Bool = false)
+  end
+
+  def call
+    import_file = @import_file || ARGV.first? || raise "An import file must be provided"
```
paulcsmith: Maybe "An import  file *path*" so it is clear it is the path and not file contents from STDIN or something

spec/spec_helper.cr
```diff
@@ -10,6 +11,7 @@ Db::Migrate.new(quiet: true).call
 
 Spec.before_each do
   TestDatabase.truncate
+  FileUtils.rm_r("tmp") rescue nil
```
paulcsmith: I don't think we should `rescue nil` because it might `rescue` a legit error. Could you rescue a specific error instead? Or maybe `rm_rf` so it deletes if it isn't there.

Also I'd suggest doing `Dir.mkdir_p("tmp")` so it is there for any test that needs it and doesn't need to manually be added

spec/tasks/db_schema_restore_spec.cr
```diff
@@ -4,7 +4,7 @@ SQL_DUMP_FILE = "spec/support/files/sample_backup.sql"
 
 describe Db::Schema::Restore do
   it "raises an error when no import file is supplied" do
-    expect_raises(Exception, "An import file must be provided") do
+    expect_raises(Exception, "A path to the import SQL file must be provided") do
```
paulcsmith: Nice

spec/spec_helper.cr
jwoertink: Yeah, turns out we already had a way to do this 😂

PR #218, opened 2019-08-22
Better error when Postgres not running

paulcsmith:
You get a weird error like this: `createdb: could not connect to database template1: could not connect to server: No such file or directory`

This PR helps people figure out what to do

### DIFF
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -70,9 +70,25 @@ class Avram::Migrator::Runner
       raise <<-ERROR
       #{message}
 
-        #{green_arrow} If you are on macOS  you can install postgres tools from #{macos_postgres_tools_link}
-        #{green_arrow} If you are on linux you can try running #{linux_postgres_installation_instructions}
-        #{green_arrow} If you are on CI or some servers, there may already be a database created so you don't need this command"
+      Try this...
+
+        ▸ #{green_arrow} If you are on macOS  you can install postgres tools from #{macos_postgres_tools_link}
+        ▸ #{green_arrow} If you are on linux you can try running #{linux_postgres_installation_instructions}
+        ▸ #{green_arrow} If you are on CI or some servers, there may already be a database created so you don't need this command"
+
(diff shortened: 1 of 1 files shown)


src/avram/migrator/runner.cr
```diff
@@ -70,9 +70,24 @@ class Avram::Migrator::Runner
...
-        #{green_arrow} If you are on linux you can try running #{linux_postgres_installation_instructions}
-        #{green_arrow} If you are on CI or some servers, there may already be a database created so you don't need this command"
+      Try this...
+
+        ▸ #{green_arrow} If you are on macOS  you can install postgres tools from #{macos_postgres_tools_link}
+        ▸ #{green_arrow} If you are on linux you can try running #{linux_postgres_installation_instructions}
+        ▸ #{green_arrow} If you are on CI or some servers, there may already be a database created so you don't need this command"
+
+      ERROR
+    elsif (message = e.message) && messages.includes?("could not connect to database template1")
```
jwoertink: I'm not sure what all the `template` mean, but my DB has `template0` through `template4`. Maybe we can just leave off the number just in case?

src/avram/migrator/runner.cr
```diff
@@ -70,9 +70,24 @@ class Avram::Migrator::Runner
...
+      raise <<-ERROR
+      Creating the database failed. It looks like Postgres is not running.
+
+      Message from Postgres:
+        #{message}
+
+      Try this...
+
+        ▸ Make sure Postgres is running
```
jwoertink: Should we also mention the `verify_connection` task here?

src/avram/migrator/runner.cr
paulcsmith: Good idea!

src/avram/migrator/runner.cr
paulcsmith: I don't think  so since this only requires a running db and `createdb`. Verify_connection wouldn't  ever work since the db  failed to be created

PR #221, opened 2019-08-28
ensure preloads get copied over on clone. 

### LINKED ISSUE #220: Query#clone doesn't clone preloads
There's a new `clone` method on Query objects so you can fork your queries. It fails when a query requires a preload.

```crystal
posts = PostQuery.new.preload_comments
post_maclone = posts.clone

# This one works fine
posts.first.comments

# This one throws error comments for Post must be preloaded with 'preload_comments'
post_maclone.first.comments
```

jwoertink:
Fixes #220

This adds a copy for the preloads on to the new cloned object.

Note: One issue is that the spec fails on this line

```
products.first.line_items.first.price.as(Price).in_cents.should eq(500)
```

With the error: 

> price for LineItem must be preloaded with 'preload_price' (Avram::LazyLoadError)

I'm not sure why since I am calling `preload_price`. Could this be an unrelated bug?

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -672,5 +672,20 @@ describe Avram::Query do
         cloned_posts.first.comments.should eq([comment])
       end
     end
+
+    it "clones nested preloads" do
+      Avram.temp_config(lazy_load_enabled: false) do
+        item = LineItemBox.create
+        price = PriceBox.create &.line_item_id(item.id).in_cents(500)
+        product = ProductBox.create
+        LineItemProductBox.create &.line_item_id(item.id).product_id(product.id)
+
+        products = Product::BaseQuery.new
+                      .preload_line_items(LineItem::BaseQuery.new.preload_price)
+        cloned_products = products.clone
+        products.first.line_items.first.price.as(Price).in_cents.should eq(500)
+        cloned_products.first.line_items.first.price.as(Price).in_cents.should eq(500)
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


paulcsmith:
Hmm. My guess is there are preloads on the other queries that also need to be copied over too

jwoertink:
Ok, I used the spec in another branch pre-clone and it passed. So verified that it's related. I'm copying over the preloads now. I need to see if I can copy over all the nested query stuff too.

jwoertink:
I was wrong again. I think there's a separate issue with those nested preloads. I ran a `git bisect` back to before avram 0.11 was released, and it was still an issue. I'm going to open a separate issue about it because my guess is that it never worked. 

As for this PR, it actually does fix the initial issue #220 which also solves an issue I have in my app for now.

spec/query_spec.cr
```diff
@@ -661,5 +661,31 @@ describe Avram::Query do
...
+        item = LineItemBox.create
+        price = PriceBox.create &.line_item_id(item.id).in_cents(500)
+        product = ProductBox.create
+        LineItemProductBox.create &.line_item_id(item.id).product_id(product.id)
+
+        products = Product::BaseQuery.new
+          .preload_line_items(LineItem::BaseQuery.new.preload_price)
+        cloned_products = products.clone
+        products.first.line_items.first.price.as(Price).in_cents.should eq(500)
+        cloned_products.first.line_items.first.price.as(Price).in_cents.should eq(500)
```
paulcsmith: I think you could shorten this up by doing something like `line_items.first.price.should_not be_nil`

PR #227, opened 2019-09-07
adding between query method. 

### LINKED ISSUE #219: Add between criteria method
I'd like to see a `between` method similar to `gte` and `lte`. It would essentially just add both in for an easier short hand.

```crystal
UserQuery.new.joined_at.between(1.week.ago, 1.day.ago)
# SELECT * FROM users WHERE joined_at BETWEEN (? AND ?)
# which is the same as doing
# SELECT * FROM users WHERE joined_at >= ? AND joined_at <= ?
UserQuery.new.joined_at.gte(1.week.ago).joined_at.lte(1.day.ago)
```

jwoertink:
fixes #219

This adds in a new `between` query method for numbers and time. 

```crystal
UserQuery.new.joined_at.between(2.days.ago, 1.hour.ago)
```

Originally I was going to add the actual `BETWEEN` sql call, but from what I've read it seems to just be an alias for `>= AND <=`. Using `BETWEEN` would have required altering a few other methods to allow for prepared statements to take more than 1 value placeholder.


spec/query_spec.cr
```diff
@@ -688,4 +688,32 @@ describe Avram::Query do
...
+      company = CompanyBox.create &.sales(50)
+      companies = Company::BaseQuery.new.sales.between(1, 100)
+      companies.query.statement.should eq "SELECT companies.id, companies.created_at, companies.updated_at, companies.sales, companies.earnings FROM companies WHERE companies.sales >= $1 AND companies.sales <= $2"
+      companies.query.args.should eq ["1", "100"]
+      companies.first.should eq company
+    end
+
+    it "queries between floats" do
+      company = CompanyBox.create &.earnings(300.45)
+      companies = Company::BaseQuery.new.earnings.between(123.45, 678.901)
```
paulcsmith: Could you add new lines above and below the line that creates the query so it is clear where the setup/exercise/verify steps start and end?

src/avram/charms/int32_extensions.cr
```diff
@@ -34,6 +34,10 @@ struct Int32
     end
 
     class Criteria(T, V) < Avram::Criteria(T, V)
+      def between(low_value : V, high_value : V)
```
paulcsmith: Thoughts on BetweenCriteria module that can be included in these?

jwoertink: Makes sense

PR #228, opened 2019-09-08
[WIP] Extend ordering with null sorting

### LINKED ISSUE #153: Add option for sorting nulls first/last
https://twitter.com/reinink/status/1151162015144173570

Add an optional arg to `desc_order`  and `asc_order` for sorting nulls

```crystal
UserQuery.new.age.desc_order(:nulls_first) # or :nulls_last
```

This should be an enum so it is type-safe. Something like `Enum::NullSorting::NullsFirst` or `Enum::NullSorting::NullsLast`. The compiler will match the symbol to the enum

jwoertink:
fixes #153

This adds the ability to order with `NULLS FIRST` or `NULLS LAST` by adding an argument to the `asc_order` and `desc_order` methods.

Something interesting I learned from postgres:
> By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.

I'm thinking maybe we use that and orders just always show the explicit null sorting....

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -625,6 +625,18 @@ describe Avram::Query do
       query = Post::BaseQuery.new.where_comments(Comment::BaseQuery.new.created_at.asc_order)
       query.to_sql[0].should contain "ORDER BY comments.created_at ASC"
     end
+
+    it "orders nulls first" do
+      query = Post::BaseQuery.new.published_at.asc_order(:nulls_first)
+
+      query.to_sql[0].should contain "ORDER BY posts.published_at ASC NULLS FIRST"
+    end
+
+    it "orders nulls last" do
+      query = Post::BaseQuery.new.published_at.asc_order(:nulls_last)
+
+      query.to_sql[0].should contain "ORDER BY posts.published_at ASC NULLS LAST"
+    end
   end
 
   describe "cloning queries" do
--- a/src/avram/criteria.cr
+++ b/src/avram/criteria.cr
@@ -2,17 +2,22 @@ class Avram::Criteria(T, V)
   property :rows, :column
   @negate_next_criteria : Bool
 
+  enum NullSorting
+    NullsFirst
+    NullsLast
+  end
+
   def initialize(@rows : T, @column : Symbol | String)
     @negate_next_criteria = false
   end
 
-  def desc_order : T
-    rows.query.order_by(column, :desc)
+  def desc_order(null_sorting : NullSorting? = nil) : T
+    rows.query.order_by(column, :desc, null_sorting)
     rows
   end
 
-  def asc_order : T
-    rows.query.order_by(column, :asc)
+  def asc_order(null_sorting : NullSorting? = nil) : T
+    rows.query.order_by(column, :asc, null_sorting)
     rows
   end
 
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -7,8 +7,8 @@ class Avram::QueryBuilder
   @raw_wheres = [] of Avram::Where::Raw
   @joins = [] of Avram::Join::SqlClause
   @orders = {
-    asc:  [] of Symbol | String,
-    desc: [] of Symbol | String,
+    asc:  [] of Tuple(ColumnName, Symbol?),
+    desc: [] of Tuple(ColumnName, Symbol?),
   }
   @selections : String = "*"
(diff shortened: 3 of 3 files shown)


src/avram/query_builder.cr
```diff
@@ -135,38 +128,35 @@ class Avram::QueryBuilder
...
     self
   end
 
   def order_sql
     if ordered?
-      "ORDER BY " + orders.map do |direction, columns|
-        next if columns.empty?
-        "#{columns.join(" #{direction.to_s.upcase}, ")} #{direction.to_s.upcase}"
-      end.reject(&.nil?).join(", ")
+      String.build do |str|
```
paulcsmith: I think the SQL can be built inside the `OrderBy` to clean things up a bit

src/avram/query_builder.cr
```diff
@@ -226,9 +216,7 @@ class Avram::QueryBuilder
   end
 
   def ordered?
-    @orders.values.any? do |columns|
-      columns.any?
-    end
+    @orders.size > 0
```
paulcsmith: I *think* you can do `@orders.any?`. But if not this is good!

src/avram/queryable.cr
```diff
@@ -99,7 +99,8 @@ module Avram::Queryable(T)
   end
 
   def order_by(column, direction) : self
-    query.order_by(column, direction)
+    direction = Avram::OrderBy::Direction.parse(direction.to_s)
```
paulcsmith: Could we change this to pass direction as-is to `OrderBy`? I think `order_by` isn't typically used directly so maybe that'll work?

Also how does `parse` work? Is that just something you get with Crystal enums?

jwoertink: We can't because enums only take symbol literals. So passing it down actually throws an exception. `parse` is just on enum members. It takes a string and returns the enum or throws an exception. So this would catch if someone didn't pass the correct symbol

jwoertink: ref: https://gitter.im/crystal-lang/crystal?at=5d76690d6e889c4bbdc77c63

src/avram/query_builder.cr
jwoertink: they both work. I just ran a small benchmark, it turns out if `@orders` is actually empty, then `@orders.size` is faster. If there's stuff in it, then `@orders.any` is faster. However, the speed difference is negligible, so I think I'll roll with any

src/avram/queryable.cr
paulcsmith: Gotcha! And that is cool about parse!

jwoertink: Yeah, we can catch that. 

> Unknown enum Avram::OrderBy::Direction value: sideways (ArgumentError)

paulcsmith: Actually I think that is pretty good. I think we can keep that and just append a message: `Unknown enum Avram::OrderBy::Direction value: sideways` + `"Accepted values are: asc, desc" or something. What do you think of that?

jwoertink: I'm cool with that! 👍

spec/query_spec.cr
```diff
@@ -625,6 +625,24 @@ describe Avram::Query do
...
+
+    it "orders nulls last" do
+      query = Post::BaseQuery.new.published_at.asc_order(:nulls_last)
+
+      query.to_sql[0].should contain "ORDER BY posts.published_at ASC NULLS LAST"
+    end
+
+    it "returns a nice error when trying to order by a weird direction" do
+      expect_raises(Exception, /Accepted values are: :asc, :desc/) do
+        Post::BaseQuery.new.order_by(:published_at, :sideways)
```
paulcsmith: We should support sideways ordering in a future PR

src/avram/criteria.cr
```diff
@@ -6,13 +6,13 @@ class Avram::Criteria(T, V)
...
-  def desc_order : T
-    rows.query.order_by(column, :desc)
+  def desc_order(null_sorting : Avram::OrderBy::NullSorting = :default) : T
+    rows.query.order_by(Avram::OrderBy.new(column, :desc, null_sorting))
     rows
   end
 
-  def asc_order : T
-    rows.query.order_by(column, :asc)
+  def asc_order(null_sorting : Avram::OrderBy::NullSorting = :default) : T
```
paulcsmith: Love it!

PR #232, opened 2019-09-10
Return all attributes

### LINKED ISSUE #231: SaveOperation#attributes does not include column_attributes if used with non-database attributes
https://gitter.im/luckyframework/Lobby?at=5d77e7dea08e2b4bd2c0ac78

When using attributes with column attributes, the column attributes are clobbered and `attributes` only returns the non-column attributes.

paulcsmith:
Fixes #231

### DIFF
--- a/src/avram/define_attribute.cr
+++ b/src/avram/define_attribute.cr
@@ -63,7 +63,7 @@ module Avram::DefineAttribute
     ensure_base_attributes_method_is_present
 
     def attributes
-      previous_def + [{{ name }}]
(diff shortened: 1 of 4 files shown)


src/avram/define_attribute.cr
```diff
@@ -63,7 +63,7 @@ module Avram::DefineAttribute
     ensure_base_attributes_method_is_present
 
     def attributes
-      previous_def + [{{ name }}]
+      ([{{ name }}] + previous_def + super).uniq
```
paulcsmith: Gotta get 'em all

PR #234, opened 2019-09-13
Add in GROUP BY querying

### LINKED ISSUE #193: Add missing GROUP BY stuff
Need both a low level and type-safe way to handle groups. Grouping in postgres is SUPER hard when compared to mysql. So this one may get a little tricky. For example:

```sql
SELECT users.id, users.name, users.created_at, users.updated_at
FROM users
INNER JOIN tasks on users.id = tasks.user_id
WHERE tasks.completed = true
GROUP BY tasks.date
```

Assuming you wrote something like:

```crystal
UserQuery.new.where_tasks(TaskQuery.new.completed(true).group_date)
```

This will throw a postgres error:

> ERROR:  column "users.id" must appear in the GROUP BY clause or be used in an aggregate function

So you'd also have to do something like:

```crystal
UserQuery.new.where_tasks(TaskQuery.new.completed(true).group_date).id_group
```

So that way the `GROUP BY` becomes `GROUP BY tasks.date, users.id`.

jwoertink:
Fixes #193 

* [x] - Low level group_by SQL generation
* [x] - High level Query access to low level group
* [x] - type-safe group_by access
* [ ] - A simple way to get counts from a group like rails `User.group(:status).count #=> {"active" => 100, "inactive" => 4}`

We have `select_count` currently, but maybe that needs to become aware if there's some grouping?

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -746,14 +746,10 @@ describe Avram::Query do
 
       users = UserQuery.new.group(&.age).group(&.id)
       users.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users GROUP BY users.age, users.id"
-      users.map(&.name).should contain "Dwight"
+      users.map(&.name).should eq ["Dwight", "Michael", "Jim"]
     end
 
     it "raises an error when grouped incorrectly" do
-      user = UserBox.create &.age(25).name("Erin")
-      user = UserBox.create &.age(25).name("Angela")
-      user = UserBox.create &.age(21).name("Pam")
-
       users = UserQuery.new.group(&.age)
 
       expect_raises(PQ::PQError, /column "users\.id" must appear in the GROUP BY/) do
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -737,4 +737,28 @@ describe Avram::Query do
       companies.first.should eq company
     end
   end
+
+  describe "#group" do
+    it "groups" do
+      user = UserBox.create &.age(25).name("Michael")
+      user = UserBox.create &.age(25).name("Dwight")
+      user = UserBox.create &.age(21).name("Jim")
+
+      users = UserQuery.new.group(&.age).group(&.id)
+      users.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users GROUP BY users.age, users.id"
+      users.map(&.name).should contain "Dwight"
+    end
+
+    it "raises an error when grouped incorrectly" do
+      user = UserBox.create &.age(25).name("Erin")
+      user = UserBox.create &.age(25).name("Angela")
+      user = UserBox.create &.age(21).name("Pam")
+
+      users = UserQuery.new.group(&.age)
+
+      expect_raises(PQ::PQError, /column "users\.id" must appear in the GROUP BY/) do
+        users.map(&.name).should contain "Pam"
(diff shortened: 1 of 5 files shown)


spec/query_spec.cr
```diff
@@ -737,4 +737,28 @@ describe Avram::Query do
...
+
+      users = UserQuery.new.group(&.age).group(&.id)
+      users.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users GROUP BY users.age, users.id"
+      users.map(&.name).should contain "Dwight"
+    end
+
+    it "raises an error when grouped incorrectly" do
+      user = UserBox.create &.age(25).name("Erin")
+      user = UserBox.create &.age(25).name("Angela")
+      user = UserBox.create &.age(21).name("Pam")
```
paulcsmith: I don’t think you need to create any users since the query will raise

spec/query_spec.cr
```diff
@@ -737,4 +737,28 @@ describe Avram::Query do
...
+
+  describe "#group" do
+    it "groups" do
+      user = UserBox.create &.age(25).name("Michael")
+      user = UserBox.create &.age(25).name("Dwight")
+      user = UserBox.create &.age(21).name("Jim")
+
+      users = UserQuery.new.group(&.age).group(&.id)
+      users.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users GROUP BY users.age, users.id"
+      users.map(&.name).should contain "Dwight"
```
paulcsmith: I think there should also be an “eq” test to make sure the result is exactly what is expected

src/avram/queryable.cr
```diff
@@ -106,6 +106,12 @@ module Avram::Queryable(T)
     raise "#{e.message}. Accepted values are: :asc, :desc"
   end
 
+  def group(&block) : self
+    criteria = yield self
+    criteria.__group
```
paulcsmith: Nice. I like the __ to make sure people don’t use the wrong version

paulcsmith:
Very small comments. Otherwise looks awesome!

For the count I think we should have a different method otherwise the return type will be weird. How about “select_counts_by_group” or “select_group_counts”?

paulcsmith:
Can be added in a separate PR though. This is useful even without a count method!

spec/query_spec.cr
```diff
@@ -746,14 +746,10 @@ describe Avram::Query do
 
       users = UserQuery.new.group(&.age).group(&.id)
       users.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users GROUP BY users.age, users.id"
-      users.map(&.name).should contain "Dwight"
+      users.map(&.name).should eq ["Dwight", "Michael", "Jim"]
```
paulcsmith: Oh so does it just return an array? What does group by do in this case. I thought it'd be a hash or something

jwoertink: It returns an array here of `.map()`, but `UserQuery.new.group(&.age).group(&.id)` just returns the `BaseQuery` object like normal. It is confusing though because technically in this query, the group is useless. 

I see this more as building blocks for the next parts to come as well as the underlying access. But if you have suggestions, or whatever, I'm all ears. Postgres grouping hurts my head lol

paulcsmith:
Yeah let’s merge it! I think this is a good starting point

PR #243, opened 2019-09-23
Update to crystal-db ~> 0.7.0 and Crystal 0.31.0

bcardiff:
Commits are fairly descriptive on their intention.
Changes are compatible with Crystal 0.30.0 and 0.31.0
The `Handle overflow` commit point some smells here or there that might be worth checking.

Most of the changes regarding crystal-db 0.7.0 are due to https://github.com/crystal-lang/crystal-db/pull/110.

### DIFF
--- a/src/avram/charms/int16_extensions.cr
+++ b/src/avram/charms/int16_extensions.cr
@@ -27,6 +27,8 @@ struct Int16
 
     def parse(value : Int32)
       SuccessfulCast(Int16).new value.to_i16
+    rescue OverflowError
+      FailedCast.new
     end
 
     def to_db(value : Int16)
--- a/src/avram/charms/int32_extensions.cr
+++ b/src/avram/charms/int32_extensions.cr
@@ -21,6 +21,8 @@ struct Int32
       SuccessfulCast(Int32).new(value)
     end
 
+    # Isn't a `def parse(value : Int64)` missing?
+
     def parse(values : Array(Int32))
       SuccessfulCast(Array(Int32)).new values
     end
--- a/src/avram/pool_statement_logging.cr
+++ b/src/avram/pool_statement_logging.cr
@@ -5,14 +5,9 @@ module DB
       statement_with_retry &.exec
     end
 
-    def exec(*args) : ExecResult
-      log_query(args)
-      statement_with_retry &.exec(*args)
-    end
-
-    def exec(args : Array) : ExecResult
-      log_query(args)
-      statement_with_retry &.exec(args)
+    def exec(*args_, args : Array? = nil) : ExecResult
+      log_query(*args_, args: args)
+      statement_with_retry &.exec(*args_, args: args)
     end
 
     def query : ResultSet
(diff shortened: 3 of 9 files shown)


src/avram/charms/int32_extensions.cr
```diff
@@ -21,6 +21,8 @@ struct Int32
       SuccessfulCast(Int32).new(value)
     end
 
+    # Isn't a `def parse(value : Int64)` missing?
```
paulcsmith: Yeah that probably makes sense!  Nice catch!

src/avram/charms/int16_extensions.cr
```diff
@@ -27,6 +27,8 @@ struct Int16
 
     def parse(value : Int32)
       SuccessfulCast(Int16).new value.to_i16
+    rescue OverflowError
+      FailedCast.new
```
paulcsmith: Nice!!

src/avram/pool_statement_logging.cr
```diff
@@ -5,39 +5,31 @@ module DB
...
+    def scalar(*args_, args : Array? = nil)
+      log_query(*args_, args: args)
+      statement_with_retry &.scalar(*args_, args: args)
     end
 
-    private def log_query(args = [] of String)
+    private def log_query(*args_, args : Array? = nil)
       Avram.settings.query_log_level.try do |level|
-        Avram.logger.log(level, {query: @query, args: args})
+        logging_args = EnumerableConcat.build(args_, args)
```
jwoertink: Where does `EnumerableConcat` come from? I don't see any docs on it, and it doesn't come up [in search](https://github.com/crystal-lang/crystal/search?q=EnumerableConcat&unscoped_q=EnumerableConcat)

src/avram/query_builder.cr
```diff
@@ -74,7 +74,7 @@ class Avram::QueryBuilder
       else
         value.to_s
       end
-    end
+    end.to_a
```
jwoertink: This is called at the end of `.map` which should return an array already. @bcardiff does this help to ensure the type is always an array?

src/avram/pool_statement_logging.cr
bcardiff: It's a helper class using in crystal-db to implement the `*args/args:Array`. I actually didn't expect to expose it so it's a `:nodoc:` https://github.com/crystal-lang/crystal-db/blob/master/src/db/enumerable_concat.cr .

The whole idea is to ensure a single Enumerable for arguments without allocating extra instances unless needed. And is due to the lack of logging that it made it's way up to avram.

src/avram/query_builder.cr
bcardiff: Yes. Since the `args:` named keyword is restricted to arrays since from a user perspective if it was a tuple the arguments could be splatted directly.

src/avram/pool_statement_logging.cr
jwoertink: ah, cool. Thanks!

src/avram/query_builder.cr
jwoertink: oh, that makes sense.

bcardiff: I might be overlooking some designs of avram though. From what I understood all the query building goes through the same process. Despite knowing by construction that the query will have only one argument: `User.find(1)`. I think that pushed the dynamic aspect of the arguments to all the queries. Maybe things can be done differently.

paulcsmith:
Rebased and closed with #250. Thanks @bcardiff

PR #259, opened 2019-10-11
adding a reset_limit and reset_offset. 

### LINKED ISSUE #248: Need a way to reset limit / offset
We already have a [reset_order](https://github.com/luckyframework/avram/blob/master/src/avram/queryable.cr#L60) method. It would be nice to have `reset_limit` and `reset_offset` methods which would let us reset our pagination query stuff. This would be helpful if you have something like:

```crystal
query = PostQuery.new.released.paginate(page: 1, per_page: 12)
# Where the `reset_pagination` method is something you wrote to call `reset_limit` and `reset_offset`
total_posts = query.reset_pagination.select_count
```

WIthout doing this, Avram will throw an exception when using `select_count` https://github.com/luckyframework/avram/blob/master/src/avram/query_builder.cr#L209

jwoertink:
Fixes #248

This adds `reset_limit` and `reset_offset` which can be combined to make your own `reset_pagination` if you need. This is useful for when you have a paginated collection, but you need to get a total count without the limit and offset.

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -757,4 +757,18 @@ describe Avram::Query do
       end
     end
   end
+
+  describe "#reset_limit" do
+    it "resets the limit to nil so we can use aggregate functions again" do
+      users = UserQuery.new.limit(10)
+      users.reset_limit.select_count.should eq 0
+    end
+  end
+
+  describe "#reset_offset" do
+    it "resets the offset to nil so we can use aggregate functions again" do
+      users = UserQuery.new.offset(10)
+      users.reset_offset.select_count.should eq 0
+    end
+  end
 end
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -759,16 +759,18 @@ describe Avram::Query do
   end
 
   describe "#reset_limit" do
-    it "resets the limit to nil so we can use aggregate functions again" do
+    it "resets the limit to nil" do
       users = UserQuery.new.limit(10)
-      users.reset_limit.select_count.should eq 0
+      users.query.limit.should eq 10
+      users.reset_limit.query.limit.should eq nil
     end
   end
 
   describe "#reset_offset" do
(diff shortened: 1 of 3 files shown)


spec/query_spec.cr
```diff
@@ -757,4 +757,18 @@ describe Avram::Query do
...
+    it "resets the limit to nil so we can use aggregate functions again" do
+      users = UserQuery.new.limit(10)
+      users.reset_limit.select_count.should eq 0
+    end
+  end
+
+  describe "#reset_offset" do
+    it "resets the offset to nil so we can use aggregate functions again" do
+      users = UserQuery.new.offset(10)
+      users.reset_offset.select_count.should eq 0
```
paulcsmith: I’m not sure this is testing the reset offset or limit method since you could remove it and I believe the test would still pass. I think you may need to inspect the sql. And you may also want to do something like creating 2 users and check that bother users are returned

spec/query_spec.cr
```diff
@@ -757,4 +757,18 @@ describe Avram::Query do
       end
     end
   end
+
+  describe "#reset_limit" do
+    it "resets the limit to nil so we can use aggregate functions again" do
```
paulcsmith: I think it’d be good to remove “so we can use aggregate functions” because it can be used in other ways. 

If you want to test that select count still works then I think a separate spec might be clearer “limits and offsets can be reset so aggregates still function” but I think that may nit be necessary. It’s up to you

spec/query_spec.cr
jwoertink: `select_count` is an aggregate method, so if the query has an offset or a limit, it throws an error 
> Can't use aggregates (count, min, etc.) with limit or offset.

Removing the `reset_offset` makes this spec fail.

spec/query_spec.cr
jwoertink: True. Really we just need to test the the limit and offset have a value before the reset, and no value after. I'll update the specs.

PR #261, opened 2019-10-11
This adds an extra before_save to SaveOperation subclasses

### LINKED ISSUE #258: set_uuid does not run with other before_save calls defined
I was trying to create a SaveOperation for a model that has a UUID as a primary key. When running that operation, I was getting errors about the primary key being nil. After digging around, I found that previous_def does not include parent classes.

You can see here, that callbacks calls `previous_def` but that doesn't ever run the before_save defined in the SaveOperationTemplate.

https://github.com/luckyframework/avram/blob/master/src/avram/callbacks.cr#L18
https://github.com/luckyframework/avram/blob/master/src/avram/save_operation_template.cr#L10

I'm using this as a temporary fix. Including that in my operations works. But this is really only good for an app using UUID's exclusively.

```
module UUIDOperations
  macro included
    before_save set_uuid
  end

  def set_uuid
    id.value ||= UUID.random
  end
end
```

jwoertink:
Fixes #258

The main issue is that `before_save` calls `previous_def` which doesn't bubble up in crystal. So a SaveOperation with no additional `before_save` callbacks ran just fine. Once you added one to the subclass, then the parent's callback was never ran. 

I also renamed the method. If someone wanted to generate a UUID other than v4, they could technically override this method with a different UUID generation.

### DIFF
--- a/spec/save_operation_spec.cr
+++ b/spec/save_operation_spec.cr
@@ -32,6 +32,11 @@ end
 
 private class SaveLineItem < LineItem::SaveOperation
   permit_columns :name
+  before_save prepare
+
+  def prepare
+    1
+  end
 end
 
 private class ValueColumnModel < BaseModel
--- a/src/avram/save_operation_template.cr
+++ b/src/avram/save_operation_template.cr
@@ -8,12 +8,20 @@ class Avram::SaveOperationTemplate
 
     class ::{{ type }}::SaveOperation < Avram::SaveOperation({{ type }})
       {% if primary_key_type.id == UUID.id %}
-        before_save set_uuid
-
-        def set_uuid
-          id.value ||= UUID.random()
-        end
+        # `before_save` calls `previous_def` which will not
+        # run this call on subclasses.
+        before_save set_primary_key_uuid
       {% end %}
+
+      macro inherited
+        {% if primary_key_type.id == UUID.id %}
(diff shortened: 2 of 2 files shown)


spec/save_operation_spec.cr
```diff
@@ -32,6 +32,11 @@ end
 
 private class SaveLineItem < LineItem::SaveOperation
   permit_columns :name
+  before_save prepare
+
+  def prepare
+    1
```
jwoertink: This doesn't have to do anything. It was just to test that before the change, these specs failed when this was here. Now it passes.

src/avram/save_operation_template.cr
```diff
@@ -8,13 +8,21 @@ class Avram::SaveOperationTemplate
...
     class ::{{ type }}::SaveOperation < Avram::SaveOperation({{ type }})
       {% if primary_key_type.id == UUID.id %}
-        before_save set_uuid
-
-        def set_uuid
-          id.value ||= UUID.random()
-        end
+        # `before_save` calls `previous_def` which will not
+        # run this call on subclasses.
+        before_save set_primary_key_uuid
```
jwoertink: I kept this here because this class isn't abstract, so you can still call `LineItem::SaveOperation.create` which would fail without this.

paulcsmith:
Closing in favor of #266

PR #263, opened 2019-10-11
Fixing query with has_one 

### LINKED ISSUE #223: Issues with has_one association
If you query a `has_one` association from the `belongs_to` side, things work ok.

```crystal
PriceQuery.new.where_line_items(LineItemQuery.new)
```

However, if you query from the `has_one` side, you'll get an error from the wrong foreign key

```crystal
LineItemQuery.new.where_price(PriceQuery.new)
# Error: undefined local variable or method 'line_item_id' for LineItem::BaseQuery
```

This is using the latest Avram, and using the models in the avram spec. 

One other potential issue I didn't look in to is on [this line](https://github.com/luckyframework/avram/blob/master/src/avram/associations/has_one.cr#L14), it doesn't taken in to account for name spaces like the [has_many](https://github.com/luckyframework/avram/blob/master/src/avram/associations/has_many.cr#L9) does.

jwoertink:
Fixes #223

This fixes 2 main issues with the `has_one`:

* Using `has_one` when your model is namespaced
* Doing a join query with has_one from both sides of the association

### DIFF
--- a/src/avram/associations/has_one.cr
+++ b/src/avram/associations/has_one.cr
@@ -11,7 +11,7 @@ module Avram::Associations::HasOne
     {% end %}
 
     {% unless foreign_key %}
-      {% foreign_key = "#{@type.name.underscore}_id".id %}
+      {% foreign_key = "#{@type.name.underscore.split("::").last.id}_id".id %}
     {% end %}
 
     {% foreign_key = foreign_key.id %}
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -66,6 +66,15 @@ class Avram::BaseQueryTemplate
                   foreign_key: primary_key_name
                 )
               )
+            {% elsif assoc[:relationship_type] == :has_one %}
+              join(
+                Avram::Join::{{ join_type.id }}.new(
+                  from: @@table_name,
+                  to: :{{ assoc[:table_name] }},
+                  foreign_key: :{{ assoc[:foreign_key] }},
+                  primary_key: primary_key_name
+                )
+              )
             {% elsif assoc[:through] %}
               {{ join_type.downcase.id }}_join_{{ assoc[:through].id }}
               __yield_where_{{ assoc[:through].id }} do |join_query|
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -70,7 +70,7 @@ class Avram::BaseQueryTemplate
               join(
                 Avram::Join::{{ join_type.id }}.new(
                   from: @@table_name,
-                  to: :{{ assoc[:table_name] }},
+                  to: {{ assoc[:type] }}::TABLE_NAME,
                   foreign_key: :{{ assoc[:foreign_key] }},
                   primary_key: primary_key_name
                 )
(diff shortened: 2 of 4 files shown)


src/avram/base_query_template.cr
```diff
@@ -66,6 +66,15 @@ class Avram::BaseQueryTemplate
                   foreign_key: primary_key_name
                 )
               )
+            {% elsif assoc[:relationship_type] == :has_one %}
+              join(
+                Avram::Join::{{ join_type.id }}.new(
+                  from: @@table_name,
+                  to: :{{ assoc[:table_name] }},
```
jwoertink: This is singular, but needs to be plural for this to work. What's the best way to do that?

jwoertink:
@paulcsmith if you have a few min, I need to get back on this one. Maybe a second set of eyes would be great. Thanks!

paulcsmith:
@jwoertink Could you rebase and push? I'll try to take a look this evening! Thanks for the ping :D

src/avram/base_query_template.cr
jwoertink: Just for more info on this: The current failing spec mentions that relation `price` does not exist.  This is where `:price` gets set, but it should probably be more like `table_for(Price)`, or `"price".pluralize`?... This is where it gets a little tricky

paulcsmith: Do you know why it works in all the other spots? Maybe the table_name that gets set for `has_one` needs to be pluralized when set? 

Either way, I believe a way to get at this is to do `assoc[:type]::TABLE_NAME`. I'm not sure, but I think that would work

paulcsmith: You could go the `table_for({{ assoc[:type] }})` route but it would break if someone set a custom table name

paulcsmith: Here's where we set the `TABLE_NAME` constant for models: https://github.com/luckyframework/avram/blob/bbe0d2d1eab82ff9e9501c2d191af43bd94e2dc4/src/avram/model.cr#L132

paulcsmith: Oh and this is prob where you'd want to change it: https://github.com/luckyframework/avram/blob/bbe0d2d1eab82ff9e9501c2d191af43bd94e2dc4/src/avram/associations/has_one.cr#L20 so that way it gets the right table name from the start

jwoertink:
Alright, this is a long time coming 😅  It took me a min to remember what all is going on here, but it's all looking good now.

src/avram/associations/has_one.cr
```diff
@@ -11,7 +11,7 @@ module Avram::Associations::HasOne
     {% end %}
 
     {% unless foreign_key %}
-      {% foreign_key = "#{@type.name.underscore}_id".id %}
+      {% foreign_key = "#{@type.name.underscore.split("::").last.id}_id".id %}
```
jwoertink: This was fixed under the [has_many](https://github.com/luckyframework/avram/blame/1a9018d9d37015c775404326ed1f33d9246554fd/src/avram/associations/has_many.cr#L9) a long time ago, so this now fixes it for the has_one

paulcsmith:
It still says WIP so holding off on merging just in case you wanted to do something else. Feel free to merge when you’re ready

jwoertink:
oh, oops! Ok. I'll take one last look over it, but I'm pretty sure it's ready to go

PR #264, opened 2019-10-12
Adds new to_prepared_sql method 

### LINKED ISSUE #239: New to_sql type method
Right now we have a `to_sql` method which returns an array containing the statement, and the args. This works great for simple queries, but when you get in to more complex queries, it starts to get a bit more difficult to test things out. 

For example, if you wanted to run this query in `psql` manually, how would you?

```
[
  "SELECT things.id, things.created_at, things.updated_at, things.name, things.group FROM things WHERE things.group = ANY ($1) ORDER BY things.name", 
  ["598c7bc1-c283-4121-be6a-192f2aa6658b", "9cbfd9bc-ad52-4865-9d54-2517ea059ef1"]
]
```

It would be great to have a `to_prepared_sql` method that would actually return a string you could and paste in to psql directly allowing you to see the results directly, and do lots of other debugging.

```sql
SELECT things.id, things.created_at, things.updated_at, things.name, things.group FROM things WHERE things.group = ANY ('{"598c7bc1-c283-4121-be6a-192f2aa6658b", "9cbfd9bc-ad52-4865-9d54-2517ea059ef1"}') ORDER BY things.name
```

jwoertink:
Fixes #239

This adds a new `to_prepared_sql` method to query objects. When debugging, or logging, you might use `to_sql` which returns an array where the first element is the SQL string, and the second is an array of the WHERE args. In some cases when debugging, you may want to copy this SQL and run it directly in psql to see the result. With larger and more complex SQL (especially with arrays and json), this becomes very difficult. You have to know what a postgres json or array object looks like, and how to manually convert it.

This new `to_prepared_sql` method takes your args, converts them to PG, then injects them in to the SQL string so you have a straight copy/paste string that can be run directly in postgres.

### DIFF
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -24,11 +24,11 @@ class Avram::QueryBuilder
   # Prepares the SQL statement by combining the `args` and `statement`
   # in to a single `String`
   def to_prepared_sql : String
-    params = args.map { |arg| String.new(PQ::Param.encode(arg).slice) }
+    params = args.map { |arg| "'#{String.new(PQ::Param.encode(arg).slice)}'" }
     i = 0
     sql = statement
-    sql.scan(/\$\d/) do |match|
-      sql = sql.gsub(match[0], params[i])
+    sql.scan(/\$\d+/) do |match|
+      sql = sql.sub(match[0], params[i])
       i += 1
     end
     sql
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -21,6 +21,19 @@ class Avram::QueryBuilder
     [statement] + args
   end
 
+  # Prepares the SQL statement by combining the `args` and `statement`
+  # in to a single `String`
+  def to_prepared_sql : String
+    params = args.map { |arg| String.new(PQ::Param.encode(arg).slice) }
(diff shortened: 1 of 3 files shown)


paulcsmith:
Feel free to rebase and merge this whenever your ready

jwoertink:
Taking a look at this again, I found 2 small issues. I'm gonna fix those first, and then I'll push it up and merge.

src/avram/query_builder.cr
```diff
@@ -21,6 +21,19 @@ class Avram::QueryBuilder
     [statement] + args
   end
 
+  # Prepares the SQL statement by combining the `args` and `statement`
+  # in to a single `String`
+  def to_prepared_sql : String
+    params = args.map { |arg| "'#{String.new(PQ::Param.encode(arg).slice)}'" }
```
jwoertink: This needs the single quotes because strings replaced in the SQL would be `name = Don` instead of `name = 'Don'`. Thankfully ALL postgres values can contain the single quotes and still work fine. 

Now, if we can figure out how to properly cast the value, then we can do `value.class.adapter.to_db(value)`. Right now that throws a compile error because the compiler thinks `value` is a Union type 😕

PR #266, opened 2019-10-13
updating SaveOperation callbacks to also call the super callbacks

### LINKED ISSUE #258: set_uuid does not run with other before_save calls defined

### LINKED ISSUE #265: Issues with SaveOperation callbacks
There seems to be an issue with SaveOperation callbacks. When you add a new custom callback, the built-in ones like default validate required don't get fired off.

Also related: https://github.com/luckyframework/avram/issues/258

jwoertink:
Fixes #265
Fixes #258 
Supersedes #261 

Since `previous_def` doesn't bubble up, we add in a call to `super` when no other methods are there. This will bubble up and catches the built-in validation callbacks, and setting the id for UUID primary key.

### DIFF
--- a/spec/save_operation_callbacks_spec.cr
+++ b/spec/save_operation_callbacks_spec.cr
@@ -52,12 +52,25 @@ private class CallbacksSaveOperation < Post::SaveOperation
   end
 end
 
-private class SaveLineItem < LineItem::SaveOperation
+private class SaveLineItemBase < LineItem::SaveOperation
   permit_columns :name
-  before_save prepare
+  getter locked : Bool = false
 
-  def prepare
-    true
+  before_save lock
+
+  def lock
+    @locked = true
+  end
+end
+
+private class SaveLineItemSub < SaveLineItemBase
+  permit_columns :name
+  getter loaded : Bool = false
+
+  before_save load
+
+  def load
+    @loaded = true
   end
 end
 
@@ -111,9 +124,11 @@ describe "Avram::SaveOperation callbacks" do
     operation.callbacks_that_ran.should eq(["before_save", "before_save_again"])
   end
 
-  it "runs before_save in parent class when a custom before_save is added" do
+  it "runs before_save in parent class and before_save in child class" do
     params = {"name" => "A fancy hat"}
-    SaveLineItem.create params do |operation, record|
+    SaveLineItemSub.create params do |operation, record|
+      operation.locked.should be_true
+      operation.loaded.should be_true
       operation.saved?.should be_true
       record.should be_a(LineItem)
     end
--- a/spec/save_operation_callbacks_spec.cr
+++ b/spec/save_operation_callbacks_spec.cr
@@ -6,8 +6,6 @@ private class CallbacksSaveOperation < Post::SaveOperation
   @callbacks_that_ran = [] of String
   getter callbacks_that_ran
 
-  before_save setup_required_attributes
-
   before_save :run_before_save
   before_save { run_before_save_again }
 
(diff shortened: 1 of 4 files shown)


spec/save_operation_callbacks_spec.cr
```diff
@@ -45,15 +43,21 @@ private class CallbacksSaveOperation < Post::SaveOperation
...
 
-  private def setup_required_attributes
-    title.value = "Title"
+  def prepare
+    true
   end
 end
 
-describe "Avram::SaveOperation callbacks" do
+describe "Avram::SaveOperation callbacks", focus: true do
```
paulcsmith: Oops!

spec/save_operation_callbacks_spec.cr
```diff
@@ -64,6 +68,7 @@ describe "Avram::SaveOperation callbacks" do
   it "runs all callbacks when saving successfully" do
     post = PostBox.create
     operation = CallbacksSaveOperation.new(post)
+    operation.set_title_from_param("A Title")
```
paulcsmith: I think this was being done by setup_required_attributes previously. If you add that back in then I think you won’t need this. 

Or maybe create another operation if you need one with and one without the setup required attributes

jwoertink: I had to remove that to test the required. I needed a way for it to fail when that value isn't added. I could create an option to pass in I guess.

spec/save_operation_callbacks_spec.cr
```diff
@@ -47,7 +48,16 @@ private class CallbacksSaveOperation < Post::SaveOperation
...
+    title.value = "Title" unless @skip_set_required
+  end
+end
+
+private class SaveLineItem < LineItem::SaveOperation
+  permit_columns :name
+  before_save prepare
+
+  def prepare
+    true
```
paulcsmith: I'm not sure I understand what this is testing. Maybe there could be an an instance var that changes from false to true that you can assert in the specs? `property prepared = false` then set that to true to ensure it runs?

jwoertink: Before this PR, calling this `before_save prepare` wouldn't call the `before_save set_uuid`. Without this `before_save prepare` it does call that parent one. So I just needed a before_save at this level. It doesn't really matter what it does. So before this PR, it would fail, but now it doesn't.

I'm open to options to make it more readable, but I just need a before_save that does anything here for the test.

jwoertink:
Not sure why those specs failed. Looking in to it. I thought they passed locally, but I'll double check.

src/avram/save_operation.cr
```diff
@@ -264,26 +266,6 @@ abstract class Avram::SaveOperation(T) < Avram::Operation
...
-    #
-    # This must run after the other before_* callbacks otherwise it might mark
-    # a field as missing and then later another callback sets the field.
-    #
-    # For example:
-    #
-    #   before_save { email.value = "default@example.com" }
-    #
-    # If we had already validate 'email' it would say 'email is missing' when
-    # in reality We set it in another callback.
```
paulcsmith: I think moving this bit of comments to the new spot would still be helpful so people know why it is there and not in a regular `before_save`

spec/save_operation_callbacks_spec.cr
paulcsmith: I think what worries me is that you could remove the `before_save prepare` and the specs would all still pass. So I think this prepare should set some variable  we could test, or add a comment explaining what it is testing.  What do you think?

jwoertink: Ah, I see what you're saying. Ok, I can do that

jwoertink: Ok, so the spec as it was no longer made sense once the require validations moved. Now I've added a new subclass to show that the `before_save` actually trickle down from the top with using some properties.

paulcsmith:
Worse, Docker was done so anything using Docker images was having issues :P Looks like it is all better now. Will merge once it passes (I rebased the branch so it has to re-run CI)

paulcsmith:
@jwoertink Could you give this a spin on your projects? If everything seems ok I think we can tag and release a  v0.12.1 with  all these nice bug fixes and enhancements :D

jwoertink:
Yup. I'll try it out and see how it goes. Thanks!

PR #271, opened 2019-11-01
Small dockerfile fixes

### LINKED ISSUE #272: Dockerfile issues
I fixed these in #271 -- but am posting here: 

1. Docker commands create "layers" -- these layers are immutable. Cleanup must happen in the same layer. Fixed that. 
2. `ADD` is for archive files, -- semantically it behaves like `COPY` but it's bad practice to use it in that manner. Fixed that. 
3. Optimized other stuff as I saw it.

robbyoconnor:
1. Docker layers once created are immutable.    Clean up, therefore, must happen in the same layer. 
2. ADD is for archives, though semantically works like COPY, it's the wrong tool.  

This PR fixes both of those plus moves stuff around. Avoid invalidating the cache as long as possible.

Closes #272

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,15 @@
 FROM crystallang/crystal:0.31.1
+WORKDIR /data
 
+# install base dependencies
 RUN apt-get update && \
-  apt-get install -y libgconf-2-4 build-essential curl libreadline-dev libevent-dev libssl-dev libxml2-dev libyaml-dev libgmp-dev git wget
+  apt-get install -y libgconf-2-4 curl libreadline-dev && \
+  # postgres 11 installation
+  curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
+  echo "deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main" | tee /etc/apt/sources.list.d/postgres.list && \
+  apt-get update && \
+  apt-get install -y postgresql-11 && \
+  rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Lucky cli
 RUN git clone https://github.com/luckyframework/lucky_cli --branch v0.17.0 --depth 1 /usr/local/lucky_cli && \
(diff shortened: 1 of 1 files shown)


jwoertink:
Hey @robbyoconnor  thanks for the PR. I don't really understand it though. Can you open up an issue to talk about why this PR is necessary? I've been able to use the Dockerfile as is for a while now just fine, but maybe there's something I'm missing like on a specific machine?

robbyoconnor:
> Hey @robbyoconnor thanks for the PR. I don't really understand it though. Can you open up an issue to talk about why this PR is necessary? I've been able to use the Dockerfile as is for a while now just fine, but maybe there's something I'm missing like on a specific machine?

Hey so I explained it -- basically you *THINK* you are cleaning up but towards the bottom of the dockerfile but you're not. This is just optimizations -- I'm fixing the build, don't worry.

robbyoconnor:
Sorry for randomly just coming in and fixing it -- I tend to do that when I see Dockerfiles that aren't doing what people think they are. The gist is this: that each command is a layer, each layer once run cannot be fixed or cleaned up and whatever crud left behind from that layer is there forever. 
The way I did things cleanup happens in the same layer...(each command is a layer, some layers don't add anything but metadata to the resulting image).

jwoertink:
Ah, yeah. I was just looking in to that after seeing this. I'm all about making things more efficient! Once the build passes, I think this will be a good change. Most of our projects have Docker, so feel free to look at those 😂

robbyoconnor:
Yeah -- sorry for the missteps -- fixing this -- it should be a better overall size as well.

robbyoconnor:
I don't envision this failing but let's wait for CI to pass but otherwise this is ready for showtime @jwoertink

robbyoconnor:
I'll tackle the other repos after this. I hope not having an attached issue is okay... I generally know what needs to change.

robbyoconnor:
and we're green -- I cleaned up a bit -- have a peek -- nothing changes semantically -- it just actually cleans up the package repo leftovers 

@jwoertink

jwoertink:
The attached issues are generally more for the community as a whole. So that way everyone knows what's going on. There could be someone else setting up Docker for their Lucky app, and they come across the issue searching for docker and lucky to find a better way they could do things. (for example). In this case it's fine though. I had someone else vet the changes as well, so it all looks good!

robbyoconnor:
I just removed a single empty line-break before the `COPY`

PR #274, opened 2019-11-04
allow passing bind_args as array to Where::Raw

repomaa:
sometimes you might parse args from dynamic input so this enables you to do that.


repomaa:
i forgot to add an overload for `Queryable#where`. That's added now.

jwoertink:
Oh, good catch. I'll let this sit a bit just in case anyone else thinks of something that might be missing on this.

repomaa:
I can also use `bind_vars` instead of `args` as variable name. I was just orientating from the way it's called in crystal-db.

jwoertink:
Oh yeah, for sure. `args` is more intuitive than `bind_vars`.

PR #277, opened 2019-11-18
fix for createdb and other runners with socket connection

### LINKED ISSUE #273: Avram::Migrator::Runner assumes hostname is present and does not work when using UNIX domain socket
If postgres is running via a UNIX domain socket instead of binding to a tcp port `Avram::Migrator::Runner.create_db` fails with the error:

```
createdb: could not connect to database template1: could not translate host name "-p" to address: Name or service not known
```

The `DATABASE_URL` in this case looks something like this: `postgres://wmoore@/read_rust_development`, as per [the PostgreSQL docs](https://www.postgresql.org/docs/9.4/libpq-connect.html#AEN41719).

> The default behavior when host is not specified is to connect to a Unix-domain socket in /tmp (or whatever socket directory was specified when PostgreSQL was built).

MatthiasWinkelmann:
This fixes #273, i. e. the  database operations run via the cli. 

Sockets not in standard locations are unlikely to work. That would require parsing the database connection URI's query string for "host" parameters, since that is where socket paths containing slashes are stored.

### DIFF
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -17,11 +17,12 @@ class Avram::Migrator::Runner
   end
 
   def self.db_host
-    URI.parse(database_url).host || "localhost"
+    host = URI.parse(database_url).host
+    host unless host.blank?
   end
 
   def self.db_port
-    URI.parse(database_url).port || "5432"
+    URI.parse(database_url).port
   end
 
   def self.db_user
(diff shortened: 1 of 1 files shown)


paulcsmith:
@jwoertink if/when you have time could you implement the change you mentioned and give this a try?

jwoertink:
closing in favor of #333 

Thanks for your work on this @MatthiasWinkelmann

PR #285, opened 2019-12-20
Fixing clone for distincts

### LINKED ISSUE #284: Clone doesn't bring over distinct selects
We aren't bringing over the distinct when doing a clone

```crystal
query = PostQuery.new.distinct_on(&.title).published_at.lte(1.day.ago).clone
query2 = PostQuery.new.published_at.lte(1.day.ago).clone 

query.to_sql == query2.to_sql #=> true
```

jwoertink:
Fixes #284

I had to take off the private and break out the methods so that way I had access to the distincts on the query being merged in.

### DIFF
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -1,6 +1,7 @@
 class Avram::QueryBuilder
   alias ColumnName = Symbol | String
   getter table
+  getter distinct_on : String | Symbol | Nil = nil
   @limit : Int32?
   @offset : Int32?
   @wheres = [] of Avram::Where::SqlClause
@@ -12,7 +13,6 @@ class Avram::QueryBuilder
   @prepared_statement_placeholder = 0
   @distinct : Bool = false
   @delete : Bool = false
-  @distinct_on : String | Symbol | Nil = nil
 
   def initialize(@table : Symbol)
   end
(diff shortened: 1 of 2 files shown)


src/avram/query_builder.cr
```diff
@@ -112,8 +114,16 @@ class Avram::QueryBuilder
...
-    @distinct || @distinct_on
+  def distinct?
+    @distinct || has_distinct_on?
+  end
+
+  def has_distinct_on?
+    !!@distinct_on
+  end
+
+  def raw_distinct_on
```
jwoertink: I wasn't totally sold on the name. We could just make it a property, or add a getter, but that wouldn't really be consistent with the rest of the properties. Maybe we leave this as is, and then next year do a huge cleanup refactor 🤔

src/avram/query_builder.cr
```diff
@@ -125,8 +127,16 @@ class Avram::QueryBuilder
     self
   end
 
-  private def distinct?
-    @distinct || @distinct_on
+  def distinct?
+    @distinct || has_distinct_on?
+  end
+
+  def has_distinct_on?
```
paulcsmith: I’m not sure I understand this method. I think it was clearer when using a conditional directly with `@distinct_on` But maybe I’m misunderstanding

src/avram/query_builder.cr
paulcsmith: I’m not sure I get what is different between `raw_disrinct_on` and @distinct_on. Could you explain that?

jwoertink: I can access `raw_distinct_on` from a separate object.  This is so I don't do something like`other_query.@distinct_on`. I could just make a `getter` for that and call `other_query.distinct_on` if that works better.

src/avram/query_builder.cr
jwoertink: I needed this separate because there may be times where `distinct?` would return true, but you don't actually have `distinct_on`. So I need to know if you have both separately. This way it's like "merge distinct if you have distinct", but then "merge distinct_on if you have distinct_on".

paulcsmith:
Oh yeah I like that! I think a getter would be great

PR #295, opened 2020-02-06
Add Attribute#changed? with to: and from: options

wout:
Adds:
- `#changed?` with optional `from:` and `to:` arguments
- `#changes`
- `#original_value`

### DIFF
--- a/spec/attribute_spec.cr
+++ b/spec/attribute_spec.cr
@@ -8,7 +8,7 @@ private class CallableMessage
   end
 end
 
-describe "Avram::Attribute", focus: true do
+describe "Avram::Attribute" do
   describe "#value" do
     it "returns nil on empty strings" do
       empty_string = Avram::Attribute.new(name: :blank, param: nil, value: " ", param_key: "test_form")
--- a/spec/attribute_spec.cr
+++ b/spec/attribute_spec.cr
@@ -9,20 +9,28 @@ private class CallableMessage
 end
 
 describe "Avram::Attribute", focus: true do
-  it "Attribute#value returns nil on empty strings" do
-    empty_string = Avram::Attribute.new(name: :blank, param: nil, value: " ", param_key: "test_form")
-    empty_string.value.should be_nil
+  describe "#value" do
+    it "returns nil on empty strings" do
+      empty_string = Avram::Attribute.new(name: :blank, param: nil, value: " ", param_key: "test_form")
+      empty_string.value.should be_nil
+    end
 
-    empty_array = Avram::Attribute.new(name: :empty_array, param: nil, value: [] of String, param_key: "test_form")
-    empty_array.value.should_not be_nil
+    it "does not return nil on empty arrays" do
+      empty_array = Avram::Attribute.new(name: :empty_array, param: nil, value: [] of String, param_key: "test_form")
+      empty_array.value.should_not be_nil
+    end
   end
 
-  it "Attribute#original_value returns nil on empty strings" do
-    empty_string = Avram::Attribute.new(name: :blank, param: nil, value: " ", param_key: "test_form")
-    empty_string.original_value.should be_nil
+  describe "#original_value" do
+    it "returns nil on empty strings" do
+      empty_string = Avram::Attribute.new(name: :blank, param: nil, value: " ", param_key: "test_form")
+      empty_string.original_value.should be_nil
+    end
 
-    empty_array = Avram::Attribute.new(name: :empty_array, param: nil, value: [] of String, param_key: "test_form")
-    empty_array.original_value.should_not be_nil
+    it "does not return nil on empty arrays" do
+      empty_array = Avram::Attribute.new(name: :empty_array, param: nil, value: [] of String, param_key: "test_form")
+      empty_array.original_value.should_not be_nil
+    end
   end
 
   it "can reset errors" do
@@ -45,7 +53,7 @@ describe "Avram::Attribute", focus: true do
     ])
   end
 
-  describe "Attribute#changed?" do
+  describe "#changed?" do
     it "tests if the value has changed since initialization" do
       attribute = Avram::Attribute.new(name: :state, param: nil, value: "caterpillar", param_key: "test_form")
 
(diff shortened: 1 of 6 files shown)


wout:
I kept `focus: true` on the Attribute specs and the two specs with issues. Here is a proposal with passing specs:

https://github.com/wout/avram/tree/attribute-changed-with-explicit-nil/

But this might not be the ideal solution as it treats setting a `nil` value as change, even if the original value is `nil` too.

spec/attribute_spec.cr
```diff
@@ -36,4 +44,38 @@ describe "Avram::Attribute" do
...
+      attribute = Avram::Attribute.new(name: :color, param: nil, value: "red", param_key: "test_form")
+
+      attribute.value = "green"
+      attribute.changed?(from: "blue").should be_false
+      attribute.changed?(from: "red").should be_true
+      attribute.changed?(to: "red").should be_false
+      attribute.changed?(to: "green").should be_true
+      attribute.changed?(from: "green", to: "blue").should be_false
+      attribute.changed?(from: "red", to: "green").should be_true
+    end
```
paulcsmith: Could you add a spec where the original value is a strong and it is changed “to” nil? I think that might fail since the “to” argument will be nil

src/avram/attribute.cr
```diff
@@ -59,11 +66,13 @@ class Avram::Attribute(T)
     errors.empty?
   end
 
-  def value=(@value)
-    @changed = true
+  def changed?(**arguments)
```
paulcsmith: The issue with using named args is that you could also pass random arguments and they’d be silently ignored. 

It also makes the documentation a bit less clear since it isn’t clear what args it accepts. 

I think what we could do here instead is make 4 different changed? Methods with different args using method overloading. So one with no args, one with to, one with from, and another with both. 

The type for to/from could also be restricted to T so that you can only pass in values that make sense

Does this make sense?

src/avram/attribute.cr
```diff
@@ -59,11 +66,13 @@ class Avram::Attribute(T)
...
+    from = arguments[:from]? ? arguments[:from]? == @original_value : true
+    to = arguments[:to]? ? arguments[:to]? == @value : true
+    @value != @original_value && to && from
   end
 
-  def changed?
-    @changed
+  def changes
```
paulcsmith: I think that instead of returning an array of changes it would be cool to add a getter for “original_value” that way you can call the methods for what you want and not worry about arrays. Thoughts?

wout:
> I think if the original value is nil and you set the value to nil again that changed? Should be false. Does that answer the question?

With the original behaviour, setting any value would make `changed?` return `true`, regardless of its contents. That's why I created the second PR (attribute-changed-with-explicit-nil branch) where setting `nil`, when the original value was `nil`, results in `changed?` returning `true`, because the two mentioned specs (spec/polymorphic_spec.cr:112 and spec/json_column_spec.cr:20) would otherwise fail.

But it does not feel right to do it that way, just to make specs pass.Unless you disagree, of course. :smile:

src/avram/attribute.cr
wout: > I think what we could do here instead is make 4 different changed? Methods with different args using method overloading. So one with no args, one with to, one with from, and another with both.

That is exactly how I implemented it at first:

```cr
def changed?
  @value != @original_value
end

def changed?(from : T)
  changed? && from == @value
end

def changed?(to : T)
  changed? && to == @value
end

def changed?(from : T, to : T)
  changed?(from: from) && changed?(to: to)
end
```

But that does not work because `changed?(to : T)` will overwrite `changed?(from : T)` as they have the same argument type. In this scenario, we would have to ditch `changed?(to : T)` because that is the least interesting option. On the other hand, that would make the implementation seem incomplete.

Another solution I was thinking of would be:

```cr
def changed_from?(value : T)
  changed? && value == @value
end

def changed_to?(value : T)
  changed? && value == @value
end
```

But then what doe we do with the from/to situation? We could keep:

```cr
def changed?(from : T, to : T)
  changed_from?(from) && changed_to?(to)
end
```
Although, I think that's making for a confusing api since `changed?(from : T)` does not exist. Therefore, the only option would be to:

```cr
def changed_from_to?(from : T, to : T)
  changed_from?(from) && changed_to?(to)
end
```

There is no ideal solution, unless I am missing something. :smile:

wout: Right, just remembered I had two other solutions in mind, but they are not very sexy:


```cr
def changed?(value : T, to : Bool)
  changed? && value == (to ? @value : @original_value)
end

# changed?("value") => testing from
# changed?("value", true) => testing to
```
:mask: 
```cr
def changed?(way : Symbol, value : T)
  changed? && value == (way == :to ? @value : @original_value)
end

# changed?(:from, "value")
# changed?(:to, "value")

# And consequently:
# changed?(:from, "value", :to, "other_value")
```
Please forgive the uglyness. :blush:

src/avram/attribute.cr
wout: There is already a getter for `original_value`:

```cr
def original_value
  ensure_no_blank(@original_value)
end
```

So, we might as well ditch the whole `changes` method. It already felt superfluous. I probably tried to match Rails' behaviour too much here.

spec/attribute_spec.cr
wout: You are right, this won't work. But while playing with this, I ran into a bigger issue. Attributes are not nillable. I quickly tested this with the following spec on the original master branch, which fails:

```cr
it "#value= is nillable" do
  color = Avram::Attribute.new(name: :color, param: nil, value: "mint", param_key: "test_form")
  color.value = nil
  color.value.should be_nil
end
```
Which results in:

```cr
In src/avram/attribute.cr:62:14

 62 | def value=(@value)
                 ^-----
Error: instance variable '@value' of Avram::Attribute(String) must be String, not Nil
```

I already have a solution ready. But before I go ahead and add the fix, am I right this should be nillable from string (or any other type for that matter)?

paulcsmith: @wout it is nila or if you explicitly set the type to nil. So if you do `Avram::Attribute(String?).new...` then it should let you set nil :)

src/avram/attribute.cr
paulcsmith: Sounds good to me!

src/avram/attribute.cr
paulcsmith: Ah yes I see what you mean. There is a workaround that I used for save operations. I created a `Nothing` class that represents an unused argument in a save operation. So argument defaults are Nothing.new. That way someone can pass “nil” and we know they meant to pass that since if they passed no argument it’d be Nothing. 

https://github.com/luckyframework/avram/blob/f533f088c24d44c53baa48d8bf6d0c823c93fd36/src/avram/needy_initializer_and_save_methods.cr

https://github.com/luckyframework/avram/blob/f533f088c24d44c53baa48d8bf6d0c823c93fd36/src/avram/needy_initializer_and_save_methods.cr#L77

https://github.com/luckyframework/avram/blob/f533f088c24d44c53baa48d8bf6d0c823c93fd36/src/avram/needy_initializer_and_save_methods.cr#L92

So I think we could have one method with two args  for “from” and “to”. The arts would default to Nothing.new so that you could check for is_a?(Nothing) like in the link I posted. Does that make sense?

paulcsmith:
I didn’t notice the other PR! Oops. Yes I agree the old behavior did not make sense. Better to only return true if the value actually changed and change the specs that broke due to the new behavior.

src/avram/attribute.cr
wout: Yes, perfect. Works like a charm! Thanks for pointing me in the right direction. :)

wout:
Ok, so now everything passes, except for one spec, which I temporarily set to pending. That is a situation where a jsonb column has a default value and a record is created with a explicit `nil` value. Because, as we discussed before, setting `nil` on an attribute that was initialized with a `nil` value does not result in `changed?` being `true`, this test fails.

So for this to pass, we should either:
- load the default value from the database and use it as an initial value in an `Avram::Attribute` if the record has not yet been persisted (if at all possible in a performant way)
- remove this test (as the whole point of a default value on a column is avoiding empty/nil values)

wout:
Here you go. Now all specs run green.

PR #296, opened 2020-02-06
Attribute changed with explicit nil

wout:
This passes but is it the right way to go?


wout:
This is no longer relevant, so I'll close it.

Issue #293, opened 2020-01-28
Add allow nil for validations


bdtomlin:
It would be nice to be able to do something like:

```crystal
validate_size_of zip_code, is: 5, allow_nil: true
validate_size_of title, min: 3, allow_nil: true
validate_inclusion_of age, in: [30, 40, 50], allow_nil: true
```

This would help in the case where you want to validate optional fields.

Fyi, I'd be glad to handle this in the near future. I'm busy finishing a couple of projects right now but hope to get back to contributing soon.

paulcsmith:
I love the idea!

wout:
Is anyone already working on this? I actually need it. If not, I will have a go.

paulcsmith:
I’d say go for it! Thanks @wout

paulcsmith:
Closed by #299

bdtomlin:
@wout, sorry for my lack of response. Somehow I missed the conversation here. Thanks for handling this.

PR #299, opened 2020-02-08
Allow nil on size and inclusion validations.

wout:
As suggested in #293, this PR adds an `allow_nil` option for `validate_size_of` and `validate_inclusion_of`.

### DIFF
--- a/spec/validations_spec.cr
+++ b/spec/validations_spec.cr
@@ -199,20 +199,23 @@ describe Avram::Validations do
   describe "validate_inclusion_of" do
     it "validates" do
       allowed_name = attribute("Jamie")
-      Avram::Validations.validate_inclusion_of allowed_name, in: ["Jamie"]
+      Avram::Validations.validate_inclusion_of(allowed_name, in: ["Jamie"])
       allowed_name.valid?.should be_true
 
       forbidden_name = attribute("123123123")
-      Avram::Validations.validate_inclusion_of forbidden_name, in: ["Jamie"]
+      Avram::Validations.validate_inclusion_of(forbidden_name, in: ["Jamie"])
       forbidden_name.errors.should eq(["is invalid"])
     end
 
     it "can allow nil" do
       nil_name = attribute(nil)
-      Avram::Validations.validate_inclusion_of nil_name, in: ["Jamie"], allow_nil: true
+      Avram::Validations.validate_inclusion_of(nil_name, in: ["Jamie"], allow_nil: true)
       nil_name.valid?.should be_true
 
-      Avram::Validations.validate_inclusion_of nil_name, in: ["Jamie"]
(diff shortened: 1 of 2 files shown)


spec/validations_spec.cr
```diff
@@ -206,6 +206,15 @@ describe Avram::Validations do
       Avram::Validations.validate_inclusion_of forbidden_name, in: ["Jamie"]
       forbidden_name.errors.should eq(["is invalid"])
     end
+
+    it "can be allowed to be nil" do
```
paulcsmith:
```suggestion
    it "can allow nil" do
```

spec/validations_spec.cr
```diff
@@ -233,5 +242,20 @@ describe Avram::Validations do
         Avram::Validations.validate_size_of does_not_matter, min: 4, max: 1
       end
     end
+
+    it "can be allowed to be nil" do
```
paulcsmith:
```suggestion
    it "can allow nil" do
```

src/avram/validations.cr
```diff
@@ -94,12 +94,13 @@ module Avram::Validations
   #
   # This will mark `state` as invalid unless the value is `"NY"`, or `"MA"`.
   def validate_inclusion_of(
-    attribute : Avram::Attribute(T),
+    attribute : Avram::Attribute,
```
paulcsmith: Was removing this required? The nice thing about leaving it is that it enforces the allowed values to be the same type as the attribute. If we can change this back that’d be awesome!

wout: That would be a missed opportunity. Still learning a lot about the type system. :)

wout:
In this proposal, the original method keeps its original shape (more or less) and an overload is created specifically for the `nil` scenario.

spec/validations_spec.cr
```diff
@@ -199,13 +199,25 @@ describe Avram::Validations do
...
       allowed_name.valid?.should be_true
 
       forbidden_name = attribute("123123123")
-      Avram::Validations.validate_inclusion_of forbidden_name, in: ["Jamie"]
+      Avram::Validations.validate_inclusion_of(forbidden_name, in: ["Jamie"])
       forbidden_name.errors.should eq(["is invalid"])
     end
+
+    it "can allow nil" do
+      nil_name = attribute(nil)
```
paulcsmith: I think there is a way to keep the old type signature with just one method. The problem before was that setting the value to “nil” meant the type of the attribute was just nil. What you could do instead is create an attribute with a specific type of String? That way it will accept an array of strings for the allowed value. You may have to manually create the attribute or change the “attribute” helper to accept a type signature it can use for creating the attribute

Does that make sense?

wout: That does the trick. Thanks!

PR #302, opened 2020-02-17
v0.13.0

jwoertink:
This updates a few things:

* Avram to 0.13.0 with Crystal 0.33.0 support
* Use latest crystal-db
* Use latest crystal-pg

The new Docker container uses Ubuntu 18.04 (Bionic Beaver)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM crystallang/crystal:0.31.1
+FROM crystallang/crystal:0.33.0
 WORKDIR /data
 
 # install base dependencies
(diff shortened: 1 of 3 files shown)


paulcsmith:
If there are no breaking changes and it works in older versions of crystal could we do a patch release instead? So that this works with new lucky projects right away? That assume this really does work on crystal 0.32 which I think it does. Thoughts?

jwoertink:
Hmm.. This is going to include all of the latest updates as well. Like features and bug fixes. I'm pretty sure it all works on 0.32, but I didn't try it out. Maybe I should just do a branch from the last release and then do a patch release instead?

paulcsmith:
Yeah that’d a good idea. Branch of last release and cherry pick the fix should do it I think!

jwoertink:
Ok, killing off this PR in favor of the next one as a patch release. I can do this release later after a few more PR are merged or whatever

Issue #292, opened 2020-01-21
select_sum fails on Int64 columns


bdtomlin:
It appears that sum doesn't work with Int64 columns. It does work with Int32 which is the type used in Avram's tests here:

https://github.com/luckyframework/avram/blob/d1d6b63cf67e779a1d902d9d13aa446253c3f3d7/spec/query_spec.cr#L378-L394

This failure can be seen with this test:
```crystal
    it "returns the sum with Int64" do
      UserBox.create &.age(2)
      UserBox.create &.age(1)
      UserBox.create &.age(3)
      sum = UserQuery.new.id.select_sum
      sum.should eq 6
    end
```

and here is the error message:

```
Failures:

  1) Avram::Query #select_sum returns the sum with Int64

       cast from PG::Numeric to Int64 failed, at /Users/bryantomlin/dev/_src/avram/src/avram/criteria.cr:120:5:120 (TypeCastError)
         from src/avram/criteria.cr:120:5 in 'select_sum'
         from spec/query_spec.cr:391:7 in '->'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/example.cr:255:3 in 'internal_run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/example.cr:35:16 in 'run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/context.cr:296:23 in 'internal_run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/context.cr:291:7 in 'run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/context.cr:296:23 in 'internal_run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/context.cr:291:7 in 'run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/context.cr:48:23 in 'run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/spec/dsl.cr:270:7 in '->'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/kernel.cr:255:3 in 'run'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/crystal/main.cr:47:5 in 'main'
         from /Users/bryantomlin/.asdf/installs/crystal/0.32.1/src/crystal/main.cr:106:3 in 'main'
[...]

PR #304, opened 2020-02-17
Fix Criteria#select_sum

bdtomlin:
Currently select_sum only works for Int32 and Int16 columns. The fix allows select_sum for Int16, Int32, Int62, and Float64 columns.

This also fixes an issue when the result is nil which was failing before.

This change also ads a `select_sum!` method that returns 0 of the correct type instead of nil if there are no records.

See issue #292

### DIFF
--- a/db/migrations/20170127143149_create_users.cr
+++ b/db/migrations/20170127143149_create_users.cr
@@ -6,6 +6,7 @@ class CreateUsers::V20170127143149 < Avram::Migrator::Migration::V1
       add name : String
       add nickname : String?
       add age : Int32
+      add year_born : Int16?
       add joined_at : Time
       add average_score : Float64?
       add available_for_hire : Bool?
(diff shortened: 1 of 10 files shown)


paulcsmith:
I just had a thought, do we need to do something similar for selecting avg?

Also, issue for documenting the return types and select_sum! https://github.com/luckyframework/website/issues/212

bdtomlin:
@paulcsmith Yes, there is definitely a problem with select_average when there are no records. It looks like the pg shard always returns a PG::Numeric or Nil.

So it seems to me that select average should always return a `Float64?` since we don't want the decimals truncated right?

With that being the case I think we should probably just add this to Criteria:

```crystal
  def select_average : Float64?
    rows.query.select_average(column)
    rows.exec_scalar.as(PG::Numeric | Nil).try &.to_f64
  end

  def select_average! : Float64
    select_average || 0_f64
  end
```

Does that seem ok?

paulcsmith:
Yes that sounds good to me!

Issue #63, opened 2019-02-28
Add a query cache


paulcsmith:

stephendolan:
@matthewmcgarvey / @jwoertink Do y'all feel like this belongs in the framework before we can cut 1.0? I'm not familiar enough with Avram to know whether or not this is something that would be a massive breaking change, but I can say that the speed of Lucky and Postgres has made query caching less of something I'm worried about having access to in my day-to-day apps.

jwoertink:
In theory, if done right, we shouldn't need any breaking changes to make this happen. However, last time I tried to tackle this, it required some hacking to crystal-db. Since crystal-db is a dependency of crystal-pg, it made things a bit more complicated.

I'd say lets try and keep this on for 1.0. Granted speeds haven't really been an issue, I think this would really put us in the next category level. I may take another stab here soon.

PR #306, opened 2020-02-19
Preload query optimization

### LINKED ISSUE #305: Skipping preload when no there's no results
I'm not sure if this is even doable, but just in case, I'm gonna leave this here. 

If you do:

```crystal
PostQuery.new.preload_author.released_at.gte(2.hours.ago)
```

This will do a query for authors where the post_id is in whatever the ids returned from the `PostQuery` are.... However, if no posts are returned from that query, then you get this `SELECT * FROM authors WHERE id IN []` . It's an unnecessary query. So if we have a way to skip it when no results are returned, we could see a nice little performance boost (we being my app team, but probably others too 😅 )

jwoertink:
Fixes #305

This prevents the preload query from running when there's no results from the main query. I think this is all of the spots where this happens. Also, I'm not sure how to test that this query doesn't actually happen. Testing that there's no results works as normal because it would be no results anyway.

### DIFF
--- a/spec/preloading_spec.cr
+++ b/spec/preloading_spec.cr
@@ -14,6 +14,7 @@ module QueryCalledPatch
     end
   end
 end
+
 class Comment::BaseQuery
   include QueryCalledPatch
 end
--- a/spec/preloading_spec.cr
+++ b/spec/preloading_spec.cr
@@ -53,6 +53,7 @@ describe "Preloading" do
 
   it "preloads has_one with custom query and nested preload" do
     with_lazy_load(enabled: false) do
+      SignInCredential::BaseQuery.times_called = 0
       user = UserBox.create
       sign_in_credential = SignInCredentialBox.create &.user_id(user.id)
 
@@ -61,6 +62,7 @@ describe "Preloading" do
       ).first
 
       user.sign_in_credential.not_nil!.user.should eq user
+      SignInCredential::BaseQuery.times_called.should eq 1
     end
   end
 
@@ -78,12 +80,14 @@ describe "Preloading" do
 
   it "preloads has_many" do
     with_lazy_load(enabled: false) do
+      Comment::BaseQuery.times_called = 0
       post = PostBox.create
       comment = CommentBox.create &.post_id(post.id)
 
       posts = Post::BaseQuery.new.preload_comments
 
       posts.results.first.comments.should eq([comment])
+      Comment::BaseQuery.times_called.should eq 1
     end
   end
 
@@ -183,12 +187,14 @@ describe "Preloading" do
 
   it "preloads belongs_to" do
     with_lazy_load(enabled: false) do
+      Post::BaseQuery.times_called = 0
       post = PostBox.create
       comment = CommentBox.create &.post_id(post.id)
 
       comments = Comment::BaseQuery.new.preload_post
 
       comments.first.post.should eq(post)
+      Post::BaseQuery.times_called.should eq 1
     end
   end
 
--- a/spec/preloading_spec.cr
+++ b/spec/preloading_spec.cr
@@ -2,6 +2,30 @@ require "./spec_helper"
 
 include LazyLoadHelpers
 
+# NOTE: This is only for testing if this is called during a query
+# TODO: Remove once a proper mocking shard is built
+module QueryCalledPatch
+  macro included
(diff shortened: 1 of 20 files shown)


jwoertink:
I'll take a look at what that would look like and give it a shot.

jwoertink:
Getting back on this, looks like there's no mock shard out there at the moment, so I'm just rolling with the patch way for now. I'll make a note for us to remove it later once there's a better solution. 
The nice thing is, I removed my code, wrote the tests, and they fail. I added my code back in, and they pass! So this will be great to have in..

Now, I did run in to an unexpected result that I'm not so sure of... It seems that when you preload from the `belongs_to` side, that query is actually ran twice... I added my patch in, and now it only runs once; however, I expected it to not run at all. Maybe this is by design? Or maybe I just found another bug?


```crystal
comments = Comment::BaseQuery.new.preload_post
comments.results

Comment::BaseQuery.times_called == 1
Post::BaseQuery.times_called == 2
```

That's called from [here](https://github.com/luckyframework/avram/blob/master/src/avram/associations/belongs_to.cr#L55). Any idea why that might be called twice here @paulcsmith ?

paulcsmith:
@jwoertink did https://github.com/waterlink/mocks.cr not work? It's hasn't been updated for a long time so maybe it just doesn't work anymore :(

But yeah I think it's fine without mocks.

The double load definitely sounds like a bug. I'm guessing it is this:

https://github.com/luckyframework/avram/blob/0dc5d6f7b13627adfc0aca93814af7387a2b8e7e/src/avram/associations/belongs_to.cr#L58

https://github.com/luckyframework/avram/blob/0dc5d6f7b13627adfc0aca93814af7387a2b8e7e/src/avram/associations/belongs_to.cr#L64

Since we don't have a query cache that gets called twice. My guess is you'd want to do something like: `records = records.results` so it loads it once and then you work on an array. That might work. Also we should have a query cache per fiber. That would be frickin awesome, but that'll be a bit more work :)

jwoertink:
That shard is really out of date, and seems to also require using the Spec2 shard which also looks to be out of date. I asked in the crystal chat, and general consensus for the moment is to hand roll mocks 😒 Not the best, but this at least lets us get some fixes in until something better comes along.

related: https://github.com/luckyframework/avram/issues/63 😄 

I'll give your idea a shot and see how that goes. This might give us even more boosts!

jwoertink:
😂  ok, nevermind... It's not actually an issue. Just me not paying attention lol

spec/preloading_spec.cr
```diff
@@ -2,6 +2,31 @@ require "./spec_helper"
...
+  include QueryCalledPatch
+end
+
+class SignInCredential::BaseQuery
+  include QueryCalledPatch
+end
+
+class Post::BaseQuery
+  include QueryCalledPatch
+end
```
paulcsmith: Thoughts on this?

```suggestion
module QuerySpy
  macro included
    class_property times_called : Int32 = 0

    def database : Avram::Database.class
      self.class.times_called += 1
      previous_def
    end
  end
end

class Comment::BaseQuery
  include QuerySpy
end

class SignInCredential::BaseQuery
  include QuerySpy
end

class Post::BaseQuery
  include QuerySpy
end
```

This turned out super well. I love the module abstraction 👍

spec/preloading_spec.cr
```diff
@@ -245,4 +270,30 @@ describe "Preloading" do
...
     posts.results.first.comments.should eq([comment])
   end
+
+  describe "when there's no results in the parent query" do
+    it "skips running the preload query for has_many" do
+      Comment::BaseQuery.times_called = 0
+      posts = Post::BaseQuery.new.preload_comments
+      posts.results
+
+      Comment::BaseQuery.times_called.should eq 0
```
paulcsmith: One way to make sure this spy is working is to add a couple of these counts in other spots to see that it is calling it only once! This would be an awesome spce to have to make sure preloading is not accidentally calling things twice :D

paulcsmith:
This is awesome! :D

jwoertink:
Indeed! Thanks for the help on it!

paulcsmith:
You're welcome! And I agree on not using mocks.cr. Your solution with the module was super clean and this way we don't have another dependency 🎉

PR #307, opened 2020-02-19
Fix aggregates

bdtomlin:
Addresses issues with aggregate queries

See #292, #304, and #287

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -369,15 +369,16 @@ describe Avram::Query do
 
   describe "#select_average" do
     it "returns the average" do
-      UserBox.create &.total_score(2000)
-      UserBox.create &.total_score(1000)
-      UserBox.create &.total_score(3000)
-      average = UserQuery.new.total_score.select_average
-      average.should eq 2000_f64
+      UserBox.create &.age(2)
+      UserBox.create &.age(1)
+      UserBox.create &.age(3)
+      average = UserQuery.new.age.select_average
+      average.should eq 2
+      average.should be_a Float64
     end
 
-    it "returns nil if no records" do
-      average = UserQuery.new.total_score.select_average
+    it "returns nil if there are no records" do
+      average = UserQuery.new.age.select_average
       average.should be_nil
     end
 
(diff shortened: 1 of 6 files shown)


spec/query_spec.cr
```diff
@@ -474,27 +477,15 @@ describe Avram::Query do
...
-    end
-
-    it "works with chained where clauses" do
-      scores = [100.4, 123.22]
-      sum = scores.sum
-      out_of_range = 200
-      scores.each { |score| UserBox.create &.average_score(score) }
-      query_sum = UserQuery.new.average_score.lt(out_of_range).average_score.select_sum
-      query_sum.should eq sum
-    end
```
paulcsmith: I think this test would still be nice to have but maybe there is a reason for deleting it? Is it covered by another test?

spec/query_spec.cr
```diff
@@ -398,72 +414,59 @@ describe Avram::Query do
...
     end
   end
 
-  describe "#select_sum! for Int64 column" do
+  describe "#select_sum! for Int32 column" do
     it "returns 0 if there are no records" do
       query_sum = UserQuery.new.id.select_sum!
-      query_sum.should eq 0_i64
+      query_sum.should eq 0
+      query_sum.should be_a Int64
```
paulcsmith: Nice catch. More robust to check the type 👍

spec/query_spec.cr
bdtomlin: Yes, it's covered in a different test. I created a generic `describe "#select_sum" do` that handles the 2 non type specific tests here: https://github.com/luckyframework/avram/blob/69259484eac9995277a3384a6ac99d981ac09d26/spec/query_spec.cr#L402-L415

It seemed unnecessary to repeat for every type of integer and float. Let me know if you would like me to do something a little different.

paulcsmith: Ah I see! I think this was a good call 👍

paulcsmith:
Looks awesome. Loving these new fixes and improved/more accurate type signatures in aggregates

Issue #309, opened 2020-02-27
Creating a Box for a model with only default columns raises an error.


wout:
Let's say I have this model:

```crystal
class Account < BaseModel
  table do
  end
end
```
And the following box:

```crystal
class AccountBox < Avram::Box
  def initialize
  end
end
```

Then running this spec:

```crystal
require "../spec_helper"

describe Account do
  it "is createable" do
    account = AccountBox.create
  end
end
```

Will raise the following error:

```crystal
Nil assertion failed (NilAssertionError)
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/nil.cr:106:5 in 'not_nil!'
         from lib/avram/src/avram/save_operation.cr:302:7 in 'save!'
         from src/models/account.cr:2:3 in 'create!'
         from spec/models/account_spec.cr:7:7 in '->'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/example.cr:255:3 in 'internal_run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/example.cr:35:16 in 'run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/context.cr:296:23 in 'internal_run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/context.cr:291:7 in 'run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/context.cr:296:23 in 'internal_run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/context.cr:291:7 in 'run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/context.cr:48:23 in 'run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/spec/dsl.cr:270:7 in '->'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/kernel.cr:255:3 in 'run'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/crystal/main.cr:47:5 in 'main'
         from /home/wout/.crenv/versions/0.32.1/share/crystal/src/crystal/main.cr:106:3 in 'main'
         from __libc_start_main
         from _start
         from ???
```

But the issue extends beyond an empty model. If the model has just one additional, nilable column, and the box does not define a value for that column, the same error is thrown.

wout:
[...]

PR #310, opened 2020-02-27
Specs for empty boxes

wout:
Added a model with just the default columns, its box and a spec to verify the box is created properly. With the old code it failed, with the new code, which is already in master, all specs pass. So this confirms the fix for #309.


paulcsmith:
Thank you! I may rename the model but thank you for writing this!

Issue #204, opened 2019-08-06
Disallow raw hash in SaveOperation


paulcsmith:
Only allow Lucky::Paramable. Make helper for wrapping hash in a Lucky::Paramable if you really need to use a hash

mdwagner:
Hey @paulcsmith ,
I just picked up this issue and started working on a PR, but after doing some of the work, I'm seeing we're using the `Hash` version quite a bit (a lot of places in specs) and was wondering if there was a reason behind wanting this kind of change.

The code is currently auto-wrapping the `Hash` with `Avram::Params`, but with this change I feel like its unnecessary overhead for consumers of `SaveOperation`.

Maybe an alternative solution is to make a note in the docs/website about recommending `Avram::Params` over just using a `Hash`?

[My branch with some of the changes](https://github.com/mdwagner/avram/tree/disallow-raw-hash-so)

paulcsmith:
@mdwagner Great question! The main reason is that people use a hash when you really shouldn't. There is almost no reason to use hash except for in specs because we're testing params.

In real apps you'd either get values from the params in Lucky or you'd pass them as named arguments (because they're type-safe): `SaveUser.create!(params, org_id: 1)`. So for example in the apps I've written there has never been a need to use hash.

So there are a lot of spots in the specs that need this to test param handling, but in real apps it isn't as good because you lose type-safety. Does that make sense?

BTW I took a look at the branch and it is looking good!

wontruefree:
what about background jobs or operation tools like CLIs and stuff?

paulcsmith:
@wontruefree Great question. In general I'd say the named tuple is the way to go if possible

```crystal
class SaveMessageJob < Mosquito::QueuedJob
  params body : String

  def perform
    SaveMessage.create(body: body)
  end
end
```

CLI can be a bit trickier since you may want to use Avram's validations. In that case a hash may make sense 👍 `SaveMessage.create(Avram::Params.new({"body": ARGV.first?.to_s}))`

The main problem we're trying to solve with disallowing hash directly is that people use it without realizing it and end up losing type-safety (and making their apps more complex because they try to merge things together)

wontruefree:
I just want to make sure it is flexible enough with things that I have done on other projects.  Lucky already has a task system it might be interesting to make that a way to run tasks against the SaveOperations

mdwagner:
[...]

PR #312, opened 2020-03-01
Disallow raw Hash in SaveOperation

mdwagner:
PR for #204 
**BREAKING CHANGE**

**Reasoning:**
> The main problem we're trying to solve with disallowing hash directly is that people use it without realizing the alternative and end up losing type-safety (and making their apps more complex because they try to merge things together)


paulcsmith:
Also gonna open an issue to see if we can add a helpful error message when people do use a hash and shows them how to use a named tuple instead

PR #314, opened 2020-03-07
Add 'skip_schema_enforcer'

### LINKED ISSUE #308: SchemaMismatchError if an ephemeral table conflicts with the existing database (because of SchemaEnforcer)
Let's say you're adding usernames to your user model when before you only had an email address. One problem is, you must add values the new `username` column using `fill_existing_with`, which doesn't take a dynamic value.

One solution is to:

* Create a nullable field and `fill_existing_with: :nothing`
* Create a temporary model with an optional `username`
* Use that temp model to generate unique usernames
* Migrate so the `username` field is now required

That's what I did in [my app](https://github.com/edwardloveall/lucky_bits/blob/5004faf83d2918e615231c90c00a24b5435d303d/db/migrations/20190120141952_generate_usernames_for_users.cr#L5).

But now that Avram's `ensure_correct_column_mappings!` is added, it complains that the temporary model (which is still technically a model) has a field (nullable `username`) that doesn't match the database (non-nullable `username`)

@paulcsmith made a suggestion:

>We should have a way to ignore models in the enforcer for stuff like this! Could you open an issue? I’m thinking a simple macro like “skip_column_enforcer” or something

paulcsmith:
Closes #308

### DIFF
--- a/src/avram/schema_enforcer.cr
+++ b/src/avram/schema_enforcer.cr
@@ -1,10 +1,11 @@
 module Avram::SchemaEnforcer
-  MODELS_TO_ENFORCE = [] of Avram::Model.class
+  ALL_MODELS     = [] of Avram::Model.class
+  MODELS_TO_SKIP = [] of String # Stringified class name
 
   macro setup(table_name, columns, type, *args, **named_args)
-    include Avram::SchemaEnforcer
-
     def self.ensure_correct_column_mappings!
+      return if Avram::SchemaEnforcer::MODELS_TO_SKIP.includes?(self.name)
+
       attributes = [
         {% for attribute in columns %}
           { name: :{{attribute[:name]}}, nilable: {{ attribute[:nilable] }}, type: {{ attribute[:type] }} },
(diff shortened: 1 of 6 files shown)


src/avram/schema_enforcer.cr
```diff
@@ -23,7 +24,7 @@ module Avram::SchemaEnforcer
       ).ensure_matches! attributes
     end
 
-    {% if !type.resolve.abstract? %}
+    {% if !type.resolve.abstract? && SHOULD_ENFORCE_SCHEMA %}
```
jwoertink: These macro parts are so hard to test 😝

src/avram/schema_enforcer.cr
```diff
@@ -10,7 +10,7 @@ module Avram::SchemaEnforcer
 
   macro setup(table_name, columns, type, *args, **named_args)
     def self.ensure_correct_column_mappings!
-      return if MODELS_TO_SLIP.includes?(self)
+      return if MODELS_TO_SKIP.includes?(self)
```
jwoertink: 😂

paulcsmith: ship it not

paulcsmith:
This was supposed to be my easy morning project...lol

PR #317, opened 2020-03-14
[WIP] Make success/failure easier with custom types!

paulcsmith:


stephendolan:
@jwoertink / @matthewmcgarvey I tried my hand at rebasing this one from `master`, but I was feeling a bit lost.

Without a description, I'm not certain if it's something we'd be better off just closing and revisiting later, but I'd love to make that decision so that we can set this poor thing free 😄

jwoertink:
Yeah, we can close this out and revisit later. I think this was just on the road to make supporting custom types easier. https://github.com/luckyframework/avram/issues/497

PR #323, opened 2020-03-16
Add soft delete functionality to Avram

### LINKED ISSUE #320: Add Avram::SoftDelete
Depends on https://github.com/luckyframework/avram/issues/319

Module to include in models that adds “soft_delete”. Another module to include in queries to add “only_soft_deleted”, “not_soft_deleted” and other scopes. We could also consider going with the “discard” name to match the ruby gem below

This lib would be similar to https://github.com/jhawthorn/discard

paulcsmith:
Closes #320


src/avram/soft_delete/model.cr
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def restore : self
+    save_operation_class.update!(self, soft_deleted_at: nil)
+  end
+
+  abstract def save_operation_class
+
+  def soft_deleted? : Bool
+    !!soft_deleted_at
```
wontruefree: does lucky not have a `present?` method?

paulcsmith: No but we should!

src/avram/soft_delete/model.cr
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def restore : self
+    save_operation_class.update!(self, soft_deleted_at: nil)
+  end
+
+  abstract def save_operation_class
+
+  def soft_deleted? : Bool
+    soft_deleted_at.present?
```
paulcsmith: @wontruefree FIxed!

PR #325, opened 2020-03-16
Added reset_where method for specific columns.

### LINKED ISSUE #319: Add “reset_where” to queryable
So you can remove where clauses. Helpful for removing defaults where’s

```crystal
# let’s say UserQuery.new is overridden and adds deleted_at.is_nil
# you could use reset to remove the where
UserQuery.new.reset_where(&.deleted_at)
``` 

Works similarly to distinct_on where you use a block so it reads better.

wout:
This will reset where for a specific column. Initially, you could also call `query.reset_where` without a block, which would reset all where clauses. But I removed that because it could lead to unexpected/unwanted results.

Closes #319

### DIFF
--- a/spec/query_builder_spec.cr
+++ b/spec/query_builder_spec.cr
@@ -29,6 +29,23 @@ describe Avram::QueryBuilder do
     query.statement.should eq "SELECT * FROM users"
   end
 
+  it "can reset where for a specific column" do
+    query = new_query
+      .where(Avram::Where::GreaterThan.new(:age, "18"))
+      .where(Avram::Where::LessThan.new(:age, "81"))
+      .where(Avram::Where::Equal.new(:name, "Pauline"))
+    query.args.should eq ["18", "81", "Pauline"]
+
+    query.reset_where("name")
+
+    query.statement.should eq "SELECT * FROM users WHERE age > $1 AND age < $2"
+    query.args.should eq ["18", "81"]
+
+    query.reset_where(:age)
+
+    query.args.should be_empty
+  end
+
   it "selects all" do
     new_query.statement.should eq "SELECT * FROM users"
     new_query.args.should eq [] of String
--- a/src/avram/criteria.cr
+++ b/src/avram/criteria.cr
@@ -135,6 +135,11 @@ class Avram::Criteria(T, V)
     rows.query.group_by(column)
   end
 
+  # :nodoc:
+  def __reset_where : Avram::QueryBuilder
+    rows.query.reset_where(column)
+  end
+
   private def add_clause(sql_clause) : T
     sql_clause = build_sql_clause(sql_clause)
     rows.query.where(sql_clause)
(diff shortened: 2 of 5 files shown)


src/avram/criteria.cr
```diff
@@ -135,6 +135,11 @@ class Avram::Criteria(T, V)
     rows.query.group_by(column)
   end
 
+  # :nodoc:
+  def __reset_where : Avram::QueryBuilder
```
wontruefree: Is there a way to make this private?
It feels weird to have to use `__` in a lang that has real method protection.

wout: I agree, it is something I would do in JS. The question is, should it be private? Things like `Thread` for example, are considered private in Crystal, while they are accessible.

paulcsmith: It may be possible but would require reworking how Criteria and Queries work. This needs to be public so the Queryable can call it, but it shouldn't be used by the public. It can be and it'll work but it just reads strangely so adding the `__` hides it a bit more.

We also do this in `distinct_on` so that you do `MyQuery.new.distinct_on(&.name)` isntead of `MyQuery.new.name.distinct_on` which reads super weird

spec/query_builder_spec.cr
```diff
@@ -29,6 +29,23 @@ describe Avram::QueryBuilder do
...
+    query.args.should eq ["18", "81", "Pauline"]
+
+    query.reset_where("name")
+
+    query.statement.should eq "SELECT * FROM users WHERE age > $1 AND age < $2"
+    query.args.should eq ["18", "81"]
+
+    query.reset_where(:age)
+
+    query.args.should be_empty
```
paulcsmith: Could you add a statement test here to match what was done on line 41?

src/avram/criteria.cr
paulcsmith: Hmm I'm not suer what you mean. If marked as `private` I am fairly certain it will not work since we are calling it from another class

spec/query_builder_spec.cr
wout: That's what I mentioned on Gitter :joy: . The result of `where_sql` is cached, so the test would fail. You can't call `statement` twice because it will always return the first result.

paulcsmith: Ahhh! 👍

src/avram/criteria.cr
wout: No, you're right, there isn't another way. You can't make it private. But I think the `__` can be omitted. Some methods and object in Crystal are considered private, while being publicly accessible. A mention in the docs would be sufficient, I think.

wout:
> Looks good! Thanks @wout. If someone can figure out a better way to "hide" or make that method private and wants to do a PR I'll accept it. I can't thin of a way but if we can that'd be cool.

You're welcome!

Just a thought, maybe something like this would make it less awkward:

```crystal
def private_reset_where
  ...
end
```

Makes it instantly obvious that this is not something for an end user.

paulcsmith:
I like that! Once I merge this in could you create a PR changing `__distinct_on` and `__reset_where` to use `private`?

wout:
Of course, coming up!

Issue #322, opened 2020-03-16
Allow update with Queryable


paulcsmith:
Still debating the exact API but something like this:

```crystal
UserQuery.new.age.gte(18).update(adult: true)
```

wontruefree:
It might be nice to pass in an operation as well.

```cr
UserQuery.new.age.gte(18).apply(UpdateUser.update!(adult: true))
```

would be: 

```cr
UpdateUser.update!(UserQuery.new.age.gte(18), adult: true)
```

PR #326, opened 2020-03-16
Log failed queries

paulcsmith:
Ran into this while working on #322. I wanted to see the failed query.

The downside to this is that if you happen to rescue a `PQ::PQError` later in your app the failed query will still be logged. I think this is acceptable though since I can't think of a reason not to.

I've added a setting though just in case you don't want to log failed queries

### DIFF
--- a/src/avram.cr
+++ b/src/avram.cr
@@ -21,6 +21,7 @@ module Avram
     setting logger : Dexter::Logger = Dexter::Logger.new(nil)
     setting query_log_level : ::Logger::Severity?
     setting save_failed_log_level : ::Logger::Severity? = ::Logger::Severity::WARN
+    setting query_failed_log_level : ::Logger::Severity? = ::Logger::Severity::ERROR
     setting database_to_migrate : Avram::Database.class, example: "AppDatabase"
     setting time_formats : Array(String) = [] of String
   end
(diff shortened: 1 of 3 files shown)


src/avram/pool_statement_logging.cr
```diff
@@ -32,5 +47,13 @@ module DB
         Avram.logger.log(level, {query: @query, args: logging_args})
       end
     end
+
+    private def log_error(*args_, args : Array? = nil)
+      Avram.settings.query_log_level.try do |level|
```
jwoertink: Should this be `log_failed_log_level`?

paulcsmith: Whoops!

src/avram.cr
```diff
@@ -21,6 +21,7 @@ module Avram
     setting logger : Dexter::Logger = Dexter::Logger.new(nil)
     setting query_log_level : ::Logger::Severity?
     setting save_failed_log_level : ::Logger::Severity? = ::Logger::Severity::WARN
+    setting query_failed_log_level : ::Logger::Severity? = ::Logger::Severity::ERROR
```
jwoertink: oh, that makes more sense 😅

PR #330, opened 2020-03-17
Allow saving empty arrays

### LINKED ISSUE #294: Can not save empty array
In my model: 

```crystal
table do
    ...
    column links : Array(String)
    ...
  end
```

In my migration:

```crystal
add links : Array(String), default: [] of String
```

when saving: 

```crystal
SaveMyModel.create!(links: [], ...)
```

fails with 

```bash
Exception Could not save SaveMyModel.
22:19:39 web.1   |    ▸ links: is required
22:19:39 web.1   |   (Avram::InvalidOperationError)
22:19:39 web.1   |    from lib/avram/src/avram/save_operation.cr:304:7 in 'save!'
```

paulcsmith:
Closes #294


src/avram/object_extensions.cr
```diff
@@ -0,0 +1,30 @@
+class Object
+  def blank_for_validates_required? : Bool
```
paulcsmith: The problem boils down to https://github.com/kostya/blank #blank? returning `true` for arrays and for `Bool` if it is false.

So what I did was add this new method we can use in `validate_required`.

But this makes me wonder if we should fork the `blank` library and just add these in so this behavior is like this for `blank?

The problem is it may be unexpected for users of Rails since Rails `false.blank? # true` and `[].blank? # true`. IMO this behavior is confusing and I'd expect those to return `false` but maybe it's just me!

cc @edwardloveall @jwoertink

paulcsmith: My thought is that either way I can merge this in for now to fix the linked issue. But would love to figure out if we should just make these changes to `blank?` directly in a later PR

PR #332, opened 2020-03-17
[BREAKING CHANGE] allow setting attributes from SaveOperation.new, also remove 'on' option for needs

### LINKED ISSUE #318: Operation.new should take keyword args
Currently, only `create` and `update` take named args, like so:

`Membership.create(group_id: 1, user_id: 2)`
`Membership.update(membership, group_id: 2)`

If you ever need to test this however, you may want to just instantiate a membership without running it to test validations or other in-operation logic. To do this, you have to actually run the `create` or `update` methods. I think running `new` and passing in the same arguments would be more expected.

However, we can't right now. `new` only takes params which are a pain to create especially if they are nested. Ideally we could do something like:

```crystal
Membership.new(group_id: 1, user_id: 2)
```

paulcsmith:
Closes #318

### DIFF
--- a/spec/operation_needs_spec.cr
+++ b/spec/operation_needs_spec.cr
@@ -20,27 +20,7 @@ private class NeedsSaveOperation < Needs::SaveOperation
   needs optional : String = "bar"
 end
 
-private class NeedsWithOnOptionSaveOperation < Needs::SaveOperation
-  needs created_by : String, on: :create
-  needs updated_by : String, on: :update
-  needs saved_by : String, on: :save
-end
-
 describe "Avram::SaveOperation needs" do
-  it "doesn't change the initializer if an 'on' option is used'" do
-    params = Avram::Params.new({"name" => "Paul"})
-    user = UserBox.create
-
-    operation = NeedsWithOnOptionSaveOperation.new(params)
-    operation.created_by.should be_nil
-    operation.updated_by.should be_nil
-    operation.saved_by.should be_nil
-    operation = NeedsWithOnOptionSaveOperation.new(user, params)
-    operation.created_by.should be_nil
-    operation.updated_by.should be_nil
-    operation.saved_by.should be_nil
-  end
-
   it "sets up a method arg for save, update, and new" do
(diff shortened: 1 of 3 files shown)


spec/operation_needs_spec.cr
```diff
@@ -20,27 +20,7 @@ private class NeedsSaveOperation < Needs::SaveOperation
   needs optional : String = "bar"
 end
 
-private class NeedsWithOnOptionSaveOperation < Needs::SaveOperation
-  needs created_by : String, on: :create
-  needs updated_by : String, on: :update
-  needs saved_by : String, on: :save
```
paulcsmith: Needs with on is rarely used and it is confusing since if you use `on` it means the type must be nilable. This is not ideal and caused more complexity in the macros. So we decided to remove it

src/avram/needy_initializer_and_save_methods.cr
```diff
@@ -114,46 +113,23 @@ module Avram::NeedyInitializerAndSaveMethods
...
-            operation.{{ attribute[:name] }}.value = {{ attribute[:name] }}
-          end
-        {% end %}
-      {% end %}
-
-      {% for attribute in ATTRIBUTES %}
-        unless {{ attribute.var }}.is_a? Nothing
-          operation.{{ attribute.var }}.value = {{ attribute.var }}
-        end
-      {% end %}
```
paulcsmith: All this is now handled in `new`

src/avram/needy_initializer_and_save_methods.cr
```diff
@@ -168,57 +144,77 @@ module Avram::NeedyInitializerAndSaveMethods
...
-      )
+        {{ needs_method_args.id }}
+        {{ attribute_method_args.id }}
+    )
+      @record = nil
       @params = Avram::Params.new
+      set_attributes({{ attribute_params.id }})
+    end
+
+    def set_attributes({{ attribute_method_args.id }})
```
paulcsmith: Extracted this so `new` can use it.

PR #334, opened 2020-03-17
Upsert

### LINKED ISSUE #298: Add upsert ability to SaveOperations
Sometimes you want to save a record, sometimes you want to create a record. But there's also some cases when you're not sure which one you need, so an `upsert`. 

```crystal
SaveUser.upsert do |op, user|
  #...
end
```

The SQL would end up looking something like this using `ON CONFLICT`
```
INSERT INTO users (id, name) VALUES (1, 'Billy') ON CONFLICT DO UPDATE SET name='Billy';
```

paulcsmith:
Closes #298

Needs

- [x] Document the macro `upsert_lookup_columns`
- [x] Document how to use `upsert` and `upsert!`
- [x] Create in luckyframework/website to add section in the guides https://github.com/luckyframework/website/issues/716
- [x] Extract a #save method that accepts a block so we don't have duplicate code

### DIFF
--- a/spec/operations/save_operation_spec.cr
+++ b/spec/operations/save_operation_spec.cr
@@ -57,6 +57,10 @@ private class ParamKeySaveOperation < ValueColumnModel::SaveOperation
   param_key :custom_param
 end
 
+private class UpsertUserOperation < User::SaveOperation
+  upsert_lookup_columns :name, :nickname
+end
+
 private class OverrideDefaults < ModelWithDefaultValues::SaveOperation
   permit_columns :greeting, :drafted_at, :published_at, :admin, :age, :money
 end
@@ -155,6 +159,107 @@ describe "Avram::SaveOperation" do
 
     operation.changes.has_key?(:name).should be_true
     operation.changes[:name].should be_nil
+  end
+
+  describe "upsert upsert_lookup_columns" do
+    describe ".upsert" do
+      it "updates the existing record if one exists" do
+        existing_user = UserFactory.create &.name("Rich").nickname(nil).age(20)
+        joined_at = Time.utc.at_beginning_of_second
+
+        UpsertUserOperation.upsert(
+          name: "Rich",
+          nickname: nil,
+          age: 30,
+          joined_at: joined_at
+        ) do |_operation, user|
+          UserQuery.new.select_count.should eq(1)
+          user = user.not_nil!
+          user.id.should eq(existing_user.id)
+          user.name.should eq("Rich")
+          user.nickname.should be_nil
+          user.age.should eq(30)
+          user.joined_at.should eq(joined_at)
+        end
+      end
+
+      it "creates a new record if match one doesn't exist" do
+        user_with_different_nickname =
+          UserFactory.create &.name("Rich").nickname(nil).age(20)
+        joined_at = Time.utc.at_beginning_of_second
+
+        UpsertUserOperation.upsert(
+          name: "Rich",
+          nickname: "R.",
+          age: 30,
+          joined_at: joined_at
+        ) do |_operation, user|
+          UserQuery.new.select_count.should eq(2)
+          # Keep existing user the same
+          user_with_different_nickname.age.should eq(20)
+          user_with_different_nickname.nickname.should eq(nil)
+
+          user = user.not_nil!
+          user.id.should_not eq(user_with_different_nickname.id)
+          user.name.should eq("Rich")
+          user.nickname.should eq("R.")
+          user.age.should eq(30)
+          user.joined_at.should eq(joined_at)
+        end
+      end
+    end
+
+    describe ".upsert!" do
+      it "updates the existing record if one exists" do
+        existing_user = UserFactory.create &.name("Rich").nickname(nil).age(20)
+        joined_at = Time.utc.at_beginning_of_second
+
+        user = UpsertUserOperation.upsert!(
+          name: "Rich",
+          nickname: nil,
+          age: 30,
+          joined_at: joined_at
+        )
+
+        UserQuery.new.select_count.should eq(1)
+        user = user.not_nil!
+        user.id.should eq(existing_user.id)
+        user.name.should eq("Rich")
+        user.nickname.should be_nil
+        user.age.should eq(30)
+        user.joined_at.should eq(joined_at)
+      end
+
+      it "creates a new record if one doesn't exist" do
+        user_with_different_nickname = UserFactory.create &.name("Rich").nickname(nil).age(20)
+        joined_at = Time.utc.at_beginning_of_second
+
+        user = UpsertUserOperation.upsert!(
+          name: "Rich",
+          nickname: "R.",
+          age: 30,
+          joined_at: joined_at
+        )
+
+        UserQuery.new.select_count.should eq(2)
+        # Keep existing user the same
+        user_with_different_nickname.age.should eq(20)
+        user_with_different_nickname.nickname.should eq(nil)
+
+        user = user.not_nil!
+        user.id.should_not eq(user_with_different_nickname.id)
+        user.name.should eq("Rich")
+        user.nickname.should eq("R.")
+        user.age.should eq(30)
+        user.joined_at.should eq(joined_at)
+      end
+
+      it "raises if the record is invalid" do
+        expect_raises(Avram::InvalidOperationError) do
+          UpsertUserOperation.upsert!(name: "")
+        end
+      end
+    end
   end
 
   describe "#errors" do
--- a/src/avram/needy_initializer_and_save_methods.cr
+++ b/src/avram/needy_initializer_and_save_methods.cr
@@ -125,7 +125,6 @@ module Avram::NeedyInitializerAndSaveMethods
         if operation.save
           yield operation, operation.record
         else
-          operation.published_save_failed_event
           yield operation, nil
         end
       {% end %}
@@ -165,12 +164,8 @@ module Avram::NeedyInitializerAndSaveMethods
       {% if with_bang %}
         operation.update!
       {% else %}
-        if operation.save
-          yield operation, operation.record.not_nil!
-        else
-          operation.published_save_failed_event
-          yield operation, operation.record.not_nil!
-        end
+        operation.save
+        yield operation, operation.record.not_nil!
       {% end %}
     end
   end
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -8,6 +8,7 @@ require "./param_key_override"
 require "./inherit_column_attributes"
 require "./validations"
 require "./operation_errors"
+require "./upsert"
 
 abstract class Avram::SaveOperation(T)
   include Avram::DefineAttribute
@@ -20,6 +21,7 @@ abstract class Avram::SaveOperation(T)
   include Avram::NestedSaveOperation
   include Avram::MarkAsFailed
   include Avram::InheritColumnAttributes
+  include Avram::Upsert
 
   enum SaveStatus
     Saved
@@ -33,7 +35,10 @@ abstract class Avram::SaveOperation(T)
 
   @record : T?
   @params : Avram::Paramable
-  getter :record, :params
+
+  # :nodoc:
+  setter :record
+  getter :params, :record
   property save_status : SaveStatus = SaveStatus::Unperformed
 
   abstract def attributes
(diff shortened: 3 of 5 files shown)


wontruefree:
So this does not use on conflict.  Is that going to be added in a later PR?

paulcsmith:
@wontruefree Yep! Ideally it will use `on conflict`, but that'll require adding new methods to the query builder so figured we can at least get this out there in the meantime, then improve it later

src/avram/save_operation.cr
```diff
@@ -30,6 +32,7 @@ abstract class Avram::SaveOperation(T) < Avram::Operation
     @@schema_class = T
   end
 
+  property record
```
wontruefree: Why is this exposed?

paulcsmith: So that it can be set in the upsert if it finds an existing record

wontruefree: I see it now.

confact:
@paulcsmith A idea i have for find_or_create is to maybe add posibility to use a block.
then it use the parameters to find it and if not find call the block to add those fields. It makes more sense in my head using that.

as example:
```
email = "hakan@example.com"
User.find_or_create(email: email) do |u|
  u.email = email
  u.username = "confact"
  u.role = :user 
  u.name = params["name"]
end
```

paulcsmith:
@confact I like that syntax, but for a couple of reasons it is tricky to get type-safe and to work with param parsing. But I like the idea. I tried a few APIs before and couldn't get something that worked, but I'll think on it and see if we can do something closer to that style.

stephendolan:
@jwoertink I've merged in `master` to this branch and resolved the conflicts/updates necessary. What are the next steps to get this merge-able?

I'm happy to take some guidance and attempt to tackle finishing this one up myself, too!

src/avram/save_operation.cr
```diff
@@ -33,8 +35,12 @@ abstract class Avram::SaveOperation(T)
 
   @record : T?
   @params : Avram::Paramable
+
   getter :record, :params
+
   property save_status : SaveStatus = SaveStatus::Unperformed
+  property record
+  property save_status
```
jwoertink, line 43: This is duplicated.

src/avram/save_operation.cr
```diff
@@ -33,8 +35,12 @@ abstract class Avram::SaveOperation(T)
 
   @record : T?
   @params : Avram::Paramable
+
   getter :record, :params
+
   property save_status : SaveStatus = SaveStatus::Unperformed
+  property record
```
jwoertink, line 42: Let's move the `getter :record` to this so we're not defining that getter twice.

src/avram/unique_columns.cr
```diff
@@ -0,0 +1,43 @@
...
+      existing_record = find_existing_unique_record(operation)
+
+      if existing_record
+        existing_record
+      else
+        operation.save!
+      end
+    end
+
+    def self.upsert!(*args, **named_args)
```
jwoertink, line 14:
```suggestion
    def self.upsert!(*args, **named_args) : T
```

src/avram/unique_columns.cr
```diff
@@ -0,0 +1,43 @@
+module Avram::UniqueColumns
+  macro unique_columns(*attribute_names)
+    def self.find_or_create!(*args, **named_args)
```
jwoertink, line 3:
```suggestion
    def self.find_or_create!(*args, **named_args) : T
```

jwoertink (changes requested):
@stephendolan Overall this looks good. I left a few suggestions to fix. The only major thing is we don't have the block forms of these methods. So we can either add those in, or roll with this as is and add the block forms as a separate PR. I'm good either way.

stephendolan:
@jwoertink I opted not to introduce all the block variants in this PR so that we can get it cleaned out of the queue and at least have the basics available for folks in the next release.

If we merge this as-is, I can follow up with an issue to document the need for those additional overloads.

spec/operations/save_operation_spec.cr
```diff
@@ -157,6 +161,92 @@ describe "Avram::SaveOperation" do
...
+        nickname: nil,
+        age: 30,
+        joined_at: Time.utc
+      )
+
+      UserQuery.new.select_count.should eq(1)
+      user = user.not_nil!
+      user.should eq(existing_user)
+      # Uses existing age. Does not do an update.
+      user.age.should eq(20)
```
matthewmcgarvey, line 179: This took a while to understand. We're saying find_or_create but it's not using all the values we pass to the method

matthewmcgarvey (commented):
I feel iffy about this. Here's some things going through my mind, they might or might not be worth anything, just what I'm thinking:
- As I commented, `unique_columns` is weird to me because the call site provides data that isn't used in the fetch. It feels disjointed
- It's introducing two methods/macros that have to be used together
- We are depending on the query object that we generated

Because we separate out queries and operations I'm hesitant to believe this is the best way to go so I looked into Elixir's Ecto ORM. They don't have a `find_or_create` method or an `upsert`. Instead, they provide options on their `Repo.insert` function to handle conflicts.

https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2

Can we talk before merging this?

jwoertink:
Yeah, I'm ok with holding off on this.

stephendolan:
I'm definitely cool with a different approach, as long as this doesn't sit for another year untouched 😆 .

Having just implemented this manually in two apps, it definitely feels like an area where it'll be easy for end-users of Lucky to make mistakes until we provide the capability.

confact:
Yea, I think it is good to have something like this. I am doing find and check if nil to create now in many places so it would for sure save a lot of code.

But the current flow is confusing to me. I am not sure what it is using to find it and what it will update. is it some of them or all of them?

If we find a good way to handle that in a good way it would be good. But I understand the issue @paulcsmith talked about with the parameters and block, so a ruby on rails-like method might be hard.

Not sure exactly how to fix it. I think it is a good call to wait out and think it through.

src/avram/save_operation.cr
```diff
@@ -33,7 +35,8 @@ abstract class Avram::SaveOperation(T)
 
   @record : T?
   @params : Avram::Paramable
-  getter :record, :params
+  property :record
```
wontruefree, line 38: hmmm.  I see why you did this but should the be a part of the public api?

spec/operations/save_operation_spec.cr
```diff
@@ -57,6 +57,10 @@ private class ParamKeySaveOperation < ValueColumnModel::SaveOperation
   param_key :custom_param
 end
 
+private class UpsertUserOperation < User::SaveOperation
+  upsert_lookup_columns :name, :nickname
```
jwoertink, line 61: I like how this reads.

spec/operations/save_operation_spec.cr
```diff
@@ -157,6 +161,54 @@ describe "Avram::SaveOperation" do
...
+      it "creates a new record if one doesn't exist" do
+        existing_user = UserFactory.create &.name("Rich").nickname(nil).age(20)
+        joined_at = Time.utc.at_beginning_of_second
+
+        user = UpsertUserOperation.upsert!(
+          name: "Rich",
+          nickname: "R.",
+          age: 30,
+          joined_at: joined_at
+        )
```
jwoertink, line 242: So this says "Look up a user by name = "Rich" and nickname = "R." which doesn't exist, right? With the bang method, should this raise an exception like `Avram::RecordNotFound`? Or maybe `user` should be `nil` here?

src/avram/save_operation.cr
jwoertink: @wontruefree you mean maybe it should be like `public getter` and `private setter`? I could see something like that

src/avram/upsert.cr
```diff
@@ -0,0 +1,32 @@
+module Avram::Upsert
+  macro upsert_lookup_columns(*attribute_names)
+    def self.upsert!(*args, **named_args) : T
+      operation = new(*args, **named_args)
+      existing_record = find_existing_unique_record(operation)
+
+      if existing_record
+        operation.record = existing_record
```
jwoertink, line 8: @wontruefree Ah, I think it's because here it would need to be public somehow. I guess an alternative could be doing something like `operation.set_record(_record: existing_record)`, but that probably just muddles up the code more than we save by just making it public and not documenting it. What do you think?

src/avram/upsert.cr
```diff
@@ -0,0 +1,32 @@
+module Avram::Upsert
+  macro upsert_lookup_columns(*attribute_names)
```
jwoertink, line 2: Out of curiosity, what does the error look like if you misspell a column name, or add a name here that doesn't exist? `upset_lookup_columns :bloopity`

jwoertink (commented):
This is looking nice! Just a few questions, but I dig the direction.

src/avram/save_operation.cr
paulcsmith: Good point. _probably_ not, or at the very least should not be documented

spec/operations/save_operation_spec.cr
paulcsmith: With the bang version it should only raise if the record is invalid. Otherwise it will create a new one if the record is not found. I should add a test for that though!

src/avram/upsert.cr
paulcsmith: I'll keep it public but not document. There may be other extensions/shards that need to hook into it. Having it there, but not documented should leave it open, but not leave regular users confused about what it is for

jwoertink: That sounds good.

src/avram/upsert.cr
paulcsmith: Pretty good actually! Shows the macro that called it and gives a nice suggestion (assuming it is a typo and not totally wrong)

Screen Shot 2021-07-12 at 2 10 33 PM

jwoertink: boom! That's way nice 👍

src/avram/needy_initializer_and_save_methods.cr
```diff
@@ -165,12 +164,8 @@ module Avram::NeedyInitializerAndSaveMethods
       {% if with_bang %}
         operation.update!
       {% else %}
-        if operation.save
-          yield operation, operation.record.not_nil!
-        else
-          operation.published_save_failed_event
```
paulcsmith, line 171: This was moved to `SaveOperation#save` so we don't need to remember to call it in multiple places

src/avram/upsert.cr
```diff
@@ -1,4 +1,62 @@
...
+  # determining if a conflicting record exists. This is done using the macro
+  # `Avram::Upsert.upsert_lookup_columns`
+  #
+  # **Note:** In almost _every_ case the `upsert_lookup_columns` should have a **unique index** defined
+  # in the database to ensure no conflicting records are created, even from outside Avram.
+  #
+  # ## Full Example
+  #
+  # ```crystal
+  # class User < Model
```
jwoertink, line 22:
```suggestion
  # class User < BaseModel
```

src/avram/upsert.cr
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  # :nodoc:
+  macro included
+    {% for method in ["upsert", "upsert!"] %}
+      # Performs a create or update depending on if there is a conflicting row in the database.
+      #
+      # See `Avram::Upsert.upsert_lookup_columns` for full documentation and examples.
+      def self.{{ method.id }}(*args, **named_args)
+        \{% raise "Please call 'upsert_lookup_columns' on #{@type} before using '{{ method.id }}'" %}
```
jwoertink, line 100: Just thinking like a newbie here, this might lead me to believe I need to do something like:

```crystal
SaveUser.upsert_lookup_columns
SaveUser.upsert!(...)
```

Maybe we can word it like "Please include the `upsert_lookup_columns` macro in #{@type} before using '{{ methods.id }}'" ?

jwoertink (approved):
This is great stuff! Overall, the implementation seems fairly simple which is nice. Just a few small wording suggestions, but if you're cool with it, merge away!

src/avram/upsert.cr
paulcsmith: Good idea! THat makes a lot of sense

matthewmcgarvey (approved):

PR #335, opened 2020-03-18
Improve nilable_eq error

paulcsmith:
Understanding when to use nilable_eq can be complex. I think have some
scenarios with examples will help.

### DIFF
--- a/src/avram/criteria.cr
+++ b/src/avram/criteria.cr
@@ -25,12 +25,38 @@ class Avram::Criteria(T, V)
   private def check_nilable!(_value : Nil)
     {% raise <<-ERROR
 
-      The 'eq' method cannot compare values that may be 'nil'
+      The 'eq' method can't compare a column to a value that may be 'nil'.
 
-      Try this...
+      ▸ If you didn't realize the value might be nil...
 
-        ▸ Use 'nilable_eq'. This method allows checking against nilable values
-        ▸ Ensure the value is not nilable by wrapping it in an 'if' or using 'not_nil!'
+          # Try using an if/case to conditionally add wheres to your query
+          query = MyQuery.new
+          name = name_that_might_be_nil
+
+          if name
+            # We can be sure value is not nil and can safely use it
+            query.name(name)
+          else
+            # Don't add the name criteria, give me all users if 'name' is nil
+            query
+          end
+
+      ▸ If you want to allow comparing to 'nil'...
+
+          # Use 'nilable_eq' to allow querying against nil.
+          #
+          # For example if you have an optional 'nickname' column and you want
+          # to allow people to query it with a String to find people with a
+          # nickname, or Nil to find people without a nickname:
+          UserQuery.new.nickname.nilable_eq(nickname_that_can_be_nil)
+
+      ▸ If the compiler is wrong and the value can't be 'nil'...
+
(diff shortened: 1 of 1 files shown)


src/avram/criteria.cr
```diff
@@ -25,12 +25,35 @@ class Avram::Criteria(T, V)
...
-      Try this...
+      ▸ If you're ok comparing this column to 'nil'...
 
-        ▸ Use 'nilable_eq'. This method allows checking against nilable values
-        ▸ Ensure the value is not nilable by wrapping it in an 'if' or using 'not_nil!'
+          # Use 'nilable_eq' to allow querying against nilable/NULL values.
+          #
+          # For example if you have a 'name' column and you want to allow
+          # people to query with a 'String' to find people with a given name, or 'nil'
+          # to find people without a name:
```
edwardloveall: This feels oddly phrased to me. Also, people without a name might be a strange thing. How about a `User`'s `invited_by_id`?

---

> For example if you have a User model with an 'invited_by_id' column, to find all people invited by a specific user:
UserQuery.new.invited_by_id.nilable_eq(42)

src/avram/criteria.cr
```diff
@@ -25,12 +25,35 @@ class Avram::Criteria(T, V)
...
+          UserQuery.new.name.nilable_eq(name_that_can_be_nil)
+
+      ▸ If the compiler is wrong and the value can't be 'nil'...
+
+          # Use 'not_nil!' to tell Crystal that the value won't actually be 'nil'
+          UserQuery.new.name(name_that_isnt_actually_nil.not_nil!)
```
edwardloveall: `name` works here though 👍

src/avram/criteria.cr
```diff
@@ -25,12 +25,35 @@ class Avram::Criteria(T, V)
...
+
+      ▸ If you didn't realize the value might be nil...
+
+          # Try using an if/case to conditionally add wheres to your query
+          query = MyQuery.new
+          name = name_that_might_be_nil
```
edwardloveall: What do you think about

```suggestion
          name = value_that_might_be_nil
```

paulcsmith: Nice! I like it

src/avram/criteria.cr
paulcsmith: Yeah you are right. Since name would almost always be not nil this does seem odd. I also think the invited_by_id has a similar problem. How about nickname? Since that is something that seems like you might want to check for a specific nickname or people with no nickname?

paulcsmith:
Thanks for the review @edwardloveall! In general does this seem like an improvement or does it just muddy things?

src/avram/criteria.cr
edwardloveall: Yeah, `nickname` works!

edwardloveall:
I think more documentation (as long as it's clear) is an improvement.

src/avram/criteria.cr
paulcsmith: 👍

paulcsmith:
@jwoertink Good point. I moved the "If you didn't realize this may be nil" to the top. That's typically been what I run into. Will merge once it is green. Thanks for the reviews!

PR #339, opened 2020-03-24
added enum support

confact:
Got help a bit from @paulcsmith with this.

It is a bit hacky. 

Instead of using enum Status, you use this macro.

Like this:
```
class Issue <  BaseModel
avram_enum Status do
  Opened = 1
end
```
and for the int32 type add `Status` as type instead.

It shall now work, but with some changes in create and update:
```
status: Issue::Status.new(:opened)
```

No tests as I am not sure how you would like to test this macro?

### DIFF
--- a/spec/support/boxes/issue_box.cr
+++ b/spec/support/boxes/issue_box.cr
@@ -1,7 +1,7 @@
 class IssueBox < BaseBox
   def initialize
-    status Issue::AvramStatus.new(0)
-    role Issue::AvramRole.new(0)
+    status Issue::AvramStatus.new(:opened)
+    role Issue::AvramRole.new(:issue)
   end
 
   def build_model
--- a/spec/support/boxes/issue_box.cr
+++ b/spec/support/boxes/issue_box.cr
@@ -1,0 +1,14 @@
+class IssueBox < BaseBox
+  def initialize
+    status Issue::AvramStatus.new(0)
+    role Issue::AvramRole.new(0)
+  end
+
+  def build_model
+    Issue.new(
+      id: 123_i64,
+      status: 0,
+      role: 0
+    )
+  end
+end
--- a/spec/support/boxes/issue_box.cr
+++ b/spec/support/boxes/issue_box.cr
@@ -1,7 +1,7 @@
 class IssueBox < BaseBox
   def initialize
-    status Issue::AvramStatus.new(:opened)
-    role Issue::AvramRole.new(:issue)
+    status Issue::Status.new(:opened)
+    role Issue::Role.new(:issue)
   end
 
   def build_model
--- a/spec/support/issue.cr
+++ b/spec/support/issue.cr
@@ -1,0 +1,23 @@
+class Issue < BaseModel
+  COLUMN_SQL = "issues.id, issues.status, issues.role"
+
+  avram_enum Status do
+    Opened = 0
+    Closed = 1
+    Duplicated = 2
+  end
+
+  avram_enum Role do
+    Issue = 0
+    Bug = 1
+    Critical = 2
+  end
+
+  table do
+    column status : Issue::AvramStatus
+    column role : Issue::AvramRole
+  end
+end
+
+class IssueQuery < Issue::BaseQuery
+end
--- a/spec/support/issue.cr
+++ b/spec/support/issue.cr
@@ -2,14 +2,14 @@ class Issue < BaseModel
   COLUMN_SQL = "issues.id, issues.status, issues.role"
 
   avram_enum Status do
-    Opened = 0
-    Closed = 1
+    Opened     = 0
+    Closed     = 1
     Duplicated = 2
   end
 
   avram_enum Role do
-    Issue = 0
-    Bug = 1
+    Issue    = 0
+    Bug      = 1
     Critical = 2
   end
 
--- a/spec/support/issue.cr
+++ b/spec/support/issue.cr
@@ -2,9 +2,9 @@ class Issue < BaseModel
   COLUMN_SQL = "issues.id, issues.status, issues.role"
 
   avram_enum Status do
-    Opened     = 0
-    Closed     = 1
-    Duplicated = 2
+    Opened
+    Closed
+    Duplicated
   end
 
   avram_enum Role do
--- a/spec/support/issue.cr
+++ b/spec/support/issue.cr
@@ -14,8 +14,8 @@ class Issue < BaseModel
   end
 
   table do
-    column status : Issue::AvramStatus
-    column role : Issue::AvramRole
+    column status : Issue::Status
+    column role : Issue::Role
   end
 end
 
--- a/spec/support/issue.cr
+++ b/spec/support/issue.cr
@@ -8,9 +8,9 @@ class Issue < BaseModel
   end
 
   avram_enum Role do
-    Issue    = 0
-    Bug      = 1
-    Critical = 2
+    Issue    = 1
+    Bug      = 2
+    Critical = 3
   end
 
   table do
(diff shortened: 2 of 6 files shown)


wontruefree:
I think you might want to add some specs for this.

confact:
@wontruefree added two tests. Will add more later if that is not enough :)

spec/support/issue.cr
```diff
@@ -0,0 +1,23 @@
+class Issue < BaseModel
+  COLUMN_SQL = "issues.id, issues.status, issues.role"
+
+  avram_enum Status do
+    Opened = 0
```
jwoertink: Do you have to define these values? It looks like it just proxies to the crystal `enum`, so maybe we can leave them off here?

spec/support/issue.cr
```diff
@@ -0,0 +1,23 @@
...
+    Closed = 1
+    Duplicated = 2
+  end
+
+  avram_enum Role do
+    Issue = 0
```
jwoertink: What if we keep the values set here, but change them to be like 1, 2, 3 so we can showcase that you can override the values if you need?

spec/support/issue.cr
```diff
@@ -0,0 +1,23 @@
...
+    Bug = 1
+    Critical = 2
+  end
+
+  table do
+    column status : Issue::AvramStatus
```
jwoertink: Does this have to be namespaced? Or can you just do `column status : AvramStatus`?

spec/support/boxes/issue_box.cr
```diff
@@ -0,0 +1,14 @@
+class IssueBox < BaseBox
+  def initialize
+    status Issue::AvramStatus.new(0)
```
jwoertink: Does this allow you to do `status Issue::AvramStatus::Opened`?

spec/support/issue.cr
confact: Yes, it is just proxies to enum, so we could leave them off here. Can remove them later :)

spec/support/issue.cr
confact: Sure, as they are just sent directly to enum this will work. I can do that later :)

spec/support/issue.cr
confact: Sadly yes, it is generating a class directly under Issue. So it has to be. Haven't found it possible. I might be wrong and will be happily wrong here.

spec/support/boxes/issue_box.cr
confact: Sadly no. The macro create two things. an enum called `Status`, and a class that works as a wrapper as `AvramStatus` we use as a type. Even Issue::Status would be best here it won't work. And AvramStatus is not an Enum, so ::Opened won't work.

It will work to send `:opened` instead of 0 though.

spec/support/issue.cr
confact: Fixed this. So resolving it :)

spec/support/issue.cr
confact: Fixed this now so resolve this :)

src/avram/charms/enum.cr
```diff
@@ -0,0 +1,69 @@
...
+
+    def value
+      @enum
+    end
+
+    def blank?
+      @enum.nil?
+    end
+
+    module Lucky
```
akadusei, line 37: The enum *wrapper* can be simplified thus:

```suggestion
  struct Avram{{ enum_name }}
    def self.adapter
      Lucky
    end

    # You may need to prefix with {{ @type }}
    #
    #   {{ @type }}::{{enum_name}}
    def initialize(@enum : {{ enum_name }})
    end

    def initialize(enum_value : Int32)
      @enum = {{ enum_name }}.from_value(enum_value)
    end

    def initialize(enum_value : String)
      @enum = {{ enum_name }}.from_value(enum_value.to_i)
    end

    forward_missing_to @enum

    module Lucky
```
Struct, because they are a better fit for immutable type wrappers.

Forwarding to `@enum` means one can do:
```crystal
status.opened? # <= instead of `status.value.opened?`
status.value # <= instead of `status.value.value` (`.value` called twice)
```

paulcsmith:
@akadusei i think that if that works that is a very nice api! @confact could you give that a try and see if it passes? I think that’d be a pretty cool API!

spec/support/issue.cr
akadusei: `column status : AvramStatus` (without the leading `Issue::`) should work OK.

I'm using a modified version of this PR, as suggested in my reviews, and in my case, `column status : Status` (without the `Issue::` namespace or `Avram` prefix) works OK.

confact:
@paulcsmith @akadusei this looked pretty neat. Only issue I found is that value wouldn't return constant, but the value of the enum, here it would be `Int32` - So I added a getter to `@enum` variable to access it to read the constant.

not 100% sure about that, would do you guys think?

spec/type_extensions/enum_spec.cr
```diff
@@ -0,0 +1,21 @@
...
+    issue.role.value.should eq(Issue::Role::Issue)
+  end
+
+  it "update enum" do
+    issue = IssueBox.create
+
+    updated_issue = Issue::SaveOperation.update!(issue, status: Issue::AvramStatus.new(:closed))
+
+    updated_issue.status.value.should eq(Issue::Status::Closed)
+    updated_issue.role.value.should eq(Issue::Role::Issue)
```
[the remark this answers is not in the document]
confact: This won't work, sadly. As it can't access the constant on enum and such. I added a getter on `@enum` to get the constant.

src/avram/charms/enum.cr
confact: Really nice, fixed this. Thanks @akadusei

akadusei:
> @paulcsmith @akadusei this looked pretty neat. Only issue I found is that value wouldn't return constant, but the value of the enum, here it would be `Int32` - So I added a getter to `@enum` variable to access it to read the constant.
> 
> not 100% sure about that, would do you guys think?

I think it's great. Having a `getter` for the raw value in a wrapper makes a lot of sense. This way, one does not ever need to access the actual enum directly. Maybe we should mark the actual enum `private`?

I would make the wrapper a `struct` instead of a `class` because enums are value types, and structs, in general, make sense for immutable type wrappers. This is not required though, so maybe a `class` is OK.

spec/support/boxes/issue_box.cr
```diff
@@ -1,7 +1,7 @@
 class IssueBox < BaseBox
   def initialize
-    status Issue::AvramStatus.new(:opened)
-    role Issue::AvramRole.new(:issue)
+    status Issue::Status.new(:opened)
+    role Issue::Role.new(:issue)
```
paulcsmith, line 4: Could this be `Issue::Role::Issue`? This is totally fine as is Just curious if it is possible to do that with boxes

confact: Sadly no, as the constants are inside the enum, this is the wrapper. I haven't found a way to directly use enums.

jwoertink: Would `Issue::AvramRole::Issue` work?

jwoertink: Thinking about this, I don't think it would work. It would end up throwing an undefined method `parse` for that enum constant. I think my only hangup was that I wasn't too keen on the interface; however, I do see why it's like this.

We can probably just roll with this as is, and always change it up later if we come up with a new idea. Like `Issue::Status[:closed]` for example... This is a pretty decent start and definitely provides a lot more than what we have now. 👍

confact:
@jwoertink I am not 100% keen either with this. But it is the best way to do it for now. As soon we have updated the type extensions to not be extensions I am sure we can do this in a better way.

For now, this is the only way to do it.

paulcsmith:
@confact Makes sense! It is still quite nice IMO and I'm excited to merge this. Thanks @confact and @akadusei!

Issue #188, opened 2019-07-31
HStore support


jwoertink:
https://www.postgresql.org/docs/current/hstore.html

In reality, when creating a new database, you're more likely to define a `jsonb` column than hstore. However, if you already have an hstore field defined, we should have a way to pull that data out and map it to `Hash(String, String)`. Since hstore is a flat key/value pair, you'll only ever get `{} of String => String` from it. Maybe we could have a special type for this. 

```crystal
table do
  # Treated the same as Hash(String, String)
  column data : HStore
end
```

Issue #122, opened 2019-07-01
Generate UUID v4 in Postgres


paulcsmith:
https://github.com/luckyframework/avram/pull/121#discussion_r299126499

This will require adding an extension so we should probably add a helper to enable the extension as well

PR #356, opened 2020-04-26
Adding in extension support

jwoertink:
This PR will be needed to do #122 and #188 

This adds in helper methods for enabling and disabling extensions.

```crystal
enable_extension "uuid-ossp"
disable_extension "uuid-ossp"
update_extension "hstore", to: "2.0"
```

### DIFF
--- a/spec/migrator/alter_extension_statement_spec.cr
+++ b/spec/migrator/alter_extension_statement_spec.cr
@@ -1,0 +1,13 @@
+require "../spec_helper"
+
+describe Avram::Migrator::AlterExtensionStatement do
+  it "builds the proper SQL to update an extension" do
+    statement = Avram::Migrator::AlterExtensionStatement.new("uuid-ossp")
+    statement.build.should eq %{ALTER EXTENSION "uuid-ossp" UPDATE;}
+  end
+
+  it "updates to a specific version" do
+    statement = Avram::Migrator::AlterExtensionStatement.new("hstore", to: "2.0")
+    statement.build.should eq %{ALTER EXTENSION "hstore" UPDATE TO '2.0';}
+  end
+end
--- a/src/avram/migrator/alter_extension_statement.cr
+++ b/src/avram/migrator/alter_extension_statement.cr
@@ -1,0 +1,16 @@
+class Avram::Migrator::AlterExtensionStatement
+  def initialize(@name : String, @to : String? = nil)
+  end
+
+  def build
+    String.build do |sql|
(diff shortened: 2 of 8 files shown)


spec/migrator/alter_extension_statement_spec.cr
```diff
@@ -0,0 +1,13 @@
...
+
+describe Avram::Migrator::AlterExtensionStatement do
+  it "builds the proper SQL to update an extension" do
+    statement = Avram::Migrator::AlterExtensionStatement.new("uuid-ossp")
+    statement.build.should eq %{ALTER EXTENSION "uuid-ossp" UPDATE;}
+  end
+
+  it "updates to a specific version" do
+    statement = Avram::Migrator::AlterExtensionStatement.new("hstore", to: "2.0")
+    statement.build.should eq %{ALTER EXTENSION "hstore" UPDATE TO '2.0';}
```
paulcsmith, line 11: I didn’t know altering extensions was a thing! Nice that you added it

src/avram/migrator/alter_extension_statement.cr
```diff
@@ -0,0 +1,16 @@
...
+  def build
+    String.build do |sql|
+      sql << %{ALTER EXTENSION "#{@name}" UPDATE}
+      sql << to_version if @to
+      sql << ";"
+    end
+  end
+
+  def to_version
+    "TO '#{@to}'"
```
paulcsmith, line 14:
```suggestion
    " TO '#{@to}'"
```

This should fix the spec failure

spec/migrator/alter_extension_statement_spec.cr
jwoertink: There's actually a TON of altering you can do, but just specifying a specific version seemed to be the most common use I found. We can always extend this later if people start needing more options.

jwoertink:
There's a few small options you could technically use like specifying the schema if it's not public, or installing an extension from a specific older version, but those seem to be super edge case since we currently don't support schemas outside of public, and I think you can just use the `update_extension` and specify a different version if you need to roll it back.

spec/migrator/alter_extension_statement_spec.cr
paulcsmith: Sounds perfect to me! People can always drop down to `execute` if they need to 👍

jwoertink: Exactly. We have escape hatches for it, so this should cover the 90% 😄

PR #357, opened 2020-04-26
Fix select_count edge case of raising an exception

### LINKED ISSUE #297: No results on a select_count
This may not be Avram, but I need to start somewhere. My latest lucky app keeps getting this error:

```text
Exception no results 
    lib/db/src/db/statement.cr:18:7 in 'scalar:args' <file>
    lib/avram/src/avram/queryable.cr:199:7 in 'select_count' <file>
    /usr/share/crystal/src/float.cr:228:3 in 'perform_action' <file>
    lib/lucky/src/lucky/route_handler.cr:0:7 in 'call' <file>
    /usr/share/crystal/src/http/server/handler.cr:26:7 in 'call' <file>
    /usr/share/crystal/src/http/server/handler.cr:26:7 in 'call' <file>
    /usr/share/crystal/src/http/server/handler.cr:26:7 in 'call' <file>
    /usr/share/crystal/src/http/server/handler.cr:26:7 in 'call' <file>
    /usr/share/crystal/src/http/server/handler.cr:26:7 in 'call' <file>
    lib/lucky/src/lucky/context_extensions.cr:13:5 in 'call' <file>
    /usr/share/crystal/src/http/server/request_processor.cr:48:11 in 'process' <file>
    /usr/share/crystal/src/fiber.cr:255:3 in 'run' <file>
    ??? <file>
```

Notice that the stack trace doesn't actually say where in my code this is being called. The only way I can tell is from looking at bugsnag and seeing the path that was called and the action. So that leads me to the only bit of code on that action calling `select_count` which is:

`results.clone.reset_limit.reset_offset.select_count` results is just a Query object. 

The strange thing that makes me think this may not be Avram is just that, it's obviously running a SQL query, but instead of getting back Int64, it's raising a no results exception. This should always be a number....  hmm... Maybe we can catch this error and just return 0?

jwoertink:
Fixes #297

In a more complex case, postgres won't return any value when trying to do a count. This causes DB to just raise an exception instead of returning 0.

### DIFF
--- a/src/avram/queryable.cr
+++ b/src/avram/queryable.cr
@@ -182,6 +182,8 @@ module Avram::Queryable(T)
   def select_count : Int64
     query.select_count
     exec_scalar.as(Int64)
+  rescue e : DB::NoResultsError
+    0_i64
   end
 
   def each
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -532,6 +532,11 @@ describe Avram::Query do
         UserQuery.new.offset(1).select_count
       end
     end
+
+    it "returns 0 if postgres returns no results" do
+      query = UserQuery.new.distinct_on(&.name).average_score.gt(5).group(&.name).group(&.id).select_count
+      query.should eq 0
+    end
   end
 
   describe "#not with an argument" do


src/avram/queryable.cr
```diff
@@ -182,6 +182,8 @@ module Avram::Queryable(T)
   def select_count : Int64
     query.select_count
     exec_scalar.as(Int64)
+  rescue e : DB::NoResultsError
+    0_i64
```
jwoertink, line 186: We're currently doing this thing where we log when a query fails. I wonder if it would be strange to still return this value, but also throw a warning log saying something like "Hey, you just ran this query BLAH, but it's not doing what you think it should be....".... In my case, that may flood the logs a bit, but just a thought..

paulcsmith: I think this is fine if I understand this correctly. It seems like the query would not raise if there were items right? In other words the query is valid and it just didn't have results because there weren't any? If that's the case then I think this is great.

Otherwise I almost think this should raise and we have a `select_count!

jwoertink: Yeah, exactly. If you have records to return, then you get that number back, but if there's no records, then instead of getting 0, PG gives you nothing. Something to do with how groups work probably. Like, no groups returned, so there was nothing to count on.

paulcsmith: Gotcha. I think 0 makes sense without a log then. If we do decide to log at some point we can add a new log class so it’s easy to disable/enable. But that can wait until someone wants it. Imo this is the expected behavior so is probably fine

PR #363, opened 2020-05-10
Adding new SubmitOperation class.

jwoertink:
### What

This PR adds in a new subclass to `Avram::Operation`. Current name I went with was `SubmitOperation`, but I'm not sold on the name if anyone has any better suggestions. This new class does a few things:

* Designed for use when your operation is not backed by a model
* Allows for using `attribute` just as you would now with a `SaveOperation` or even just `Operation`.
* Allows for using `needs` like `SaveOperation` because a regular `Operation` doesn't.
* Sets up structure for future allowing callbacks to match up with `SaveOperation`. 

The eventual goal would be that you either inherit from this class or `SaveOperation`, and never really `Operation` as the base class. However, by branching off, this will allow people to slowly roll over as they need without breaking anything out there currently. 

### Using

To use this, you would create your class, and inherit from it. Then call the `run` method on your operation passing in your params and such.

```crystal
class SignInUser < Avram::SubmitOperation
  param_key :user
  include UserFromEmail

  attribute email : String
  attribute password : String

  # You must define this method,
  # And it should return the thing you want (or nil)
  def submit
    user = user_from_email
    validate_credentials(user)

    user
  end

  private def validate_credentials(user)
    if user
      unless Authentic.correct_password?(user, password.value.to_s)
        password.add_error "is wrong"
      end
    else
    end
  end
end

post "/sign_in" do
    SignInUser.run(params) do |operation, authenticated_user|
      if authenticated_user
        # sign in
      else
        # failed
      end
    end
  end
```

### Future API

Eventually, the above might look like this:

```crystal
before_submit :validate_credentials

def submit
  user
end

memoize def user
  user_from_email
end

private def validate_credentials
  # validate user
end
```

Then you'd have access to the errors if it failed

```crystal
SignInUser.run(params) do |operation, authenticated_user|
  # assuming `authenticated_user` is nil
  # you can check out your `operation.errors` like normal      
end
```

### DIFF
--- a/spec/operation_spec.cr
+++ b/spec/operation_spec.cr
@@ -41,6 +41,15 @@ private class ParamKeySaveOperation < Avram::Operation
   param_key :custom_param
 end
 
+private class NeedyOperation < Avram::SubmitOperation
+  needs test_operation : TestOperation
+  attribute secret : String
+
+  def execute
+    "The secret to #{test_operation.class} is #{secret.value}"
+  end
+end
+
 describe Avram::Operation do
   it "has create/update args for non column attributes" do
     UserWithVirtual.create(password: "p@ssword") do |operation, _user|
@@ -107,6 +116,16 @@ describe Avram::Operation do
         :name => ["is required"],
         :age  => ["is not old enough"],
       })
+    end
+  end
+
+  describe "using needs" do
+    it "allows needs to be passed in" do
+      test_op = TestOperation.new
+      params = Avram::Params.new({"secret" => "shhh"})
+      NeedyOperation.run(params, test_operation: test_op) do |op, result|
+        result.should eq "The secret to TestOperation is shhh"
+      end
     end
   end
 end
--- a/spec/operation_spec.cr
+++ b/spec/operation_spec.cr
@@ -50,6 +50,16 @@ private class NeedyOperation < Avram::SubmitOperation
   end
 end
 
+private class FailedSubmitOp < Avram::SubmitOperation
+  attribute option : String
+
+  def submit
+    validate_required option
+
+    option
+  end
+end
+
 describe Avram::Operation do
   it "has create/update args for non column attributes" do
     UserWithVirtual.create(password: "p@ssword") do |operation, _user|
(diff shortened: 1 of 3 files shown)


wout:
Love this! I literally needed this a few days ago for an email subscription form making an API call to Mailjet.

As for the name, what about `DiscreteOperation`?

jwoertink:
That's a good word, but I think it's too close to "Discreet" meaning more secretive. I actually liked when we had `VirtualForm`, but maybe using `Virtual` could get confusing with the fact that it was deprecated and removed. 🤔

wout:
Good point, discrete/discreet can be confusing. I've always found it odd to be used for graphics cards in laptops 😀️. Second on my list was `DistinctOperation`.
`Virtual` is also not quite fitting I think, because it's not really virtual, just diverging from its standard behavior. And while I am writing this, why not `DivergentOperation`?

jwoertink:
I originally wanted to just beef up the parent class, but there was a few issues doing that. The `needs` methods aren't created in the parent class because they're tightly coupled with needing a `@record` and setting up the create/update methods we well. Just moving this up a level means that all operations have to specify some `@record : T` variable. `SaveOperation` is made generic right now, and would require `Operation` also becoming generic. 

To get these to be separate, I had to break things out which caused a LOT of refactoring, and breaking changes. I figured going this route would be less invasive while letting people migrate over. 

Now, I'm all for your suggested path, I'm just not sure how to do that without essentially rewriting the whole thing from the ground up 😝

paulcsmith:
> Now, I'm all for your suggested path, I'm just not sure how to do that without essentially rewriting the whole thing from the ground up 😝


Unfortunately I think that's exactly what needs to happen. That's why this has been on the to-do list for so long haha. If you wanna give it a go feel free. If not I can try to tackle it. It's gonna be a beast of a PR lol

jwoertink:
Alright, I can have a whack at it. As long as we're on the same page. I didn't want to put in a ton of work and it be like "oh, you can do it in 1 line of code" 😂

paulcsmith:
Haha it'll def require a lot of work. I think in the end the code itself won't be much more complex it'll just need to be moved around and save operations will need to be redone to use the new `Operation` base class.

Feel free to open a draft PR early and ping me if you have any thoughts as you go. I'm sure there will be plenty of weird edge cases 😂

Issue #365, opened 2020-05-12
Rename column in migrations


wout:
Shouldn't we add a `rename` method to rename columns using migrations?

```crystal
alter table_for(User) do
  rename :old_name, :new_name
end
```

jwoertink:
I was like "we already have it...." Then I went to go look for it, and couldn't find it 😅  I guess we overlooked this. Good catch!

jwoertink:
Looks like this is now a thing! Thanks for tackling that 🎉

Issue #321, opened 2020-03-16
Consistency in migrations between add / remove


jwoertink:
When we run `alter` we have to use this syntax:

```crystal
alter table_for(Whatever) do
  add new_column : String
  remove :old_column
end
```

In `remove`, we have to pass a symbol, but for `add` we can just pass the name. If you pass the column name to `remove`, you'll get the undefined method error.

paulcsmith:
> If you pass the column name to remove, you'll get the undefined method error.

Could you clarify this part? I'm not sure I understand

jwoertink:
```crystal
  def migrate
    alter table_for(User) do
      add name : String, fill_existing_with: "anonymous"
    end
  end

  def rollback
    alter table_for(User) do
     # This fails with undefined method `name`
      remove name
    end
  end
```

paulcsmith:
Oh I see! Gotcha 👍 I agree that should work or at least raise a better error. I'm leaning toward "one way" and raising if not given a symbol. What do you think? I think the bare name is good in macros at the class level but inside other methods I think it's weird and makes it look like it is calling a method

jwoertink:
Yeah, I guess that makes sense. If you're in the class space, it's fine, but in the method space it's a little weird. I definitely think it should all be the same for all of them. Less guessing. though..... hold up...

`add :name : String` looks weird... 🤔

paulcsmith:
Oh yeah for add I think the type declaration is the way to go. But for remove it doesn't make sense since it doesn't matter what the type is. It is a little weird but probably the most clear since if you did `remove name : Int32` it would work and would be super weird haha.

I'm leaning toward symbol in `remove`, and raise compile error with help text if given something else. That way if you do use it "wrong" it'll point you in the right direction. What do you thin?k

jwoertink:
I'm not a big fan on the guessing game (i.e. when is it symbol, when is it not?). Looking at the [guides](https://luckyframework.org/guides/database/migrations#using-fill_existing_with-and-default-values) it looks like we intermix them. 

I guess it looks like the rule is if you pass a type declaration, then no symbol, and if you don't pass a type declaration, then you do pass a symbol...  I can get behind that as long as we raise some nice errors if you do the opposite, and then document on the guides. I like what @edwardloveall says with the element of least surprise.

edwardloveall:
How bout `remove name`?

paulcsmith:
@edwardloveall Definitely possible but here were my thoughts

[...]

PR #366, opened 2020-05-12
Add rename and rename_belongs_to for migrations.

wout:
A proposal as suggested in #365. This makes me think about picking up #321 as well, since the same is true.

### DIFF
--- a/src/avram/migrator/alter_table_statement.cr
+++ b/src/avram/migrator/alter_table_statement.cr
@@ -61,17 +61,18 @@ class Avram::Migrator::AlterTableStatement
   end
 
   def alter_statements : Array(String)
-    if (rows + renamed_rows + dropped_rows).empty?
-      [] of String
-    else
-      statement = String.build do |statement|
+    alterations = renamed_rows.map do |statement|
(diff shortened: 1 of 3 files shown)


wout:
Good you asked for the migration. As it turns out, you can only do one `RENAME`  per `ALTER TABLE` statement. I didn't know that. :blush: 

https://stackoverflow.com/questions/23274679/renaming-multiple-columns-in-one-statement-with-postgresql

Although not the same issue, I also included the conversion of the `rename` method to a macro here as well.

src/avram/migrator/alter_table_statement.cr
```diff
@@ -166,13 +168,36 @@ class Avram::Migrator::AlterTableStatement
     ]
   end
 
-  def remove(name : Symbol)
-    dropped_rows << "  DROP #{name.to_s}"
+  {% symbol_expected_message = "%s expected a symbol like ':user', instead got: '%s'" %}
```
paulcsmith, line 171: This is a really cool idea 👍

wout: Yeah, I was really happy with that one. :blush:

PR #369, opened 2020-05-21
[WIP] Operation Refactor Sandbox

### LINKED ISSUE #7: Handle updating nested has_one form
Related to https://github.com/luckyframework/lucky_record/pull/247 and #52

### LINKED ISSUE #20: Issues with blank strings as params
Given
```crystal
class OptionalValuesModel < LuckyRecord::Model
  table optional_values_model do
    column optional_int32 : Int32?
    column optional_int64 : Int64?
    column optional_float64 : Float64?
    column optional_string : String?
    column optional_bool : Bool?
    column optional_time : Time?
    column optional_uuid : UUID?
  end
end
#...
class OptionalValuesModelForm < OptionalValuesModel::BaseForm
  fillable :optional_int32, :optional_int64, :optional_float64, :optional_string, :optional_bool, :optional_time, :optional_uuid 
end
```

```crystal
it "coerces blank strings to nil values" do
  form = OptionalValuesModelForm.new(
    {
      "optional_int32" => "",
      "optional_int64" => "",
      "optional_float64" => "",
      "optional_string" => "",
      "optional_bool" => "",
      "optional_time" => "",
      "optional_uuid" => ""
    }
  )
  form.optional_int32.value.should be_nil
  form.optional_int64.value.should be_nil
  form.optional_float64.value.should be_nil
  form.optional_string.value.should be_nil
  form.optional_bool.value.should be_nil
  form.optional_time.value.should be_nil
  form.optional_uuid.value.should be_nil
  form.valid?.should be_true
end
```

However, none of the type extensions (other than for `String`, of course) knows how to handle blank strings and will be marked invalid as a result, and the test fails. See https://github.com/snadon/unlucky (specifically files related to a `Thing` model) which demonstrates this in a less abstract way.

### LINKED ISSUE #27: saving nilable item with form
when save empty field it raises error

Should allow save nilable values in database
needed change in src/lucky_record/form.cr:
```
def set_{{ field[:name] }}_from_param(_value)
  parse_result = {{ field[:type] }}::Lucky.parse(_value)
  if parse_result.is_a? LuckyRecord::Type::SuccessfulCast
    {{ field[:name] }}.value = parse_result.value
  else
    {{ field[:name] }}.add_error "is invalid"
  end
end
```
to this:
```
def set_{{ field[:name] }}_from_param(_value)
  parse_result = {{ field[:type] }}::Lucky.parse(_value)
  if parse_result.is_a? LuckyRecord::Type::SuccessfulCast
    {{ field[:name] }}.value = parse_result.value
  else
    {% if field[:nilable] %}
      {{ field[:name] }}.value = nil
    {% else %}
      {{ field[:name] }}.add_error "is invalid"
    {% end %}
  end
end
```
or cast emty string as nil

### LINKED ISSUE #28: Luckyform treats nullable has_one fillable as required fillable
Reproduce
Put a nullable has_one field to model
Make it fillable for the form (***_id)
make a visible form with text_input
submit the form with blank text_input
and LuckyForm complains the validity of the field.

I hope it can respect the nullable nature of the model.

### LINKED ISSUE #41: Raise nicer error if no nested params are in the params sent to a form
> **This still needs some thought. Please don't work on this yet**

It can be confusing when using forms in this scenario

https://gitter.im/luckyframework/Lobby?at=59f61f2d5a1758ed0f645e2f

After further thought, I think it should raise a helpful error at runtime, but ALSO raise a compile time error if you try to pass it params and have no allowed any fields

Also consider user `params.nested?` instead of `params.nested!` and only raising when you actually try to *save*. That way you can instantiate a form without nested params, but when you save it, it'll give you an error if the form name is not there.

### LINKED ISSUE #43: Handle has_many nested forms
```crystal
class TaskForm < Task::BaseForm
  allow title
end

class TaskListForm < TaskList::BaseForm
  allow title
  has_many task_forms : TaskForm
end

# Use in an action
TaskListForm.create(params) do |task_list, form|
  # Handle like a normal form
end

# In a page
text_input @task_list_form.title
@task_list_form.task_forms.each do |task_form|
  text_input task_form.title
end
```

## Some ideas for implementation

I think that a good starting point would be adding `many_nested` and `many_nested?` to `LuckyRecord::Paramable` Basically this would just be an abstract def that would need to then be implemented by `LuckyRecord::Params`.

Then you'd need to go the [lucky repo](https://github.com/luckyframework/lucky) and implement `many_nested` in `Lucky::Params`. For that it should:

* Look for form params in this format: `form_name:field_name[position]`. Example: `tasks:title[0]` would be the title for the first task. `tasks:body[0]` would be the body for the first task. Get it with `params.many_nested(:task)`
* JSON is much easier. It looks for an array under the given form name. E.g. `params.many_nested(:tasks)` gets `{tasks: [{title: "title", body: "body"}]}`
* It treats all JSON fields as string. So: `{users: [{age: 30}]}` would return `[{"age" => "30"}]` in Crystal

## Prior art

* https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html
* https://www.pluralsight.com/guides/ruby-on-rails-nested-attributes

### LINKED ISSUE #69: Method for working with VirtualForms more easily
It is a common pattern when using [`VirtualForms`](https://luckyframework.org/guides/saving-with-forms/#virtual-forms) to yield the form object and one or more additional objects when the form is valid. For example:

```crystal
class MyForm < Avram::VirtualForm
  virtual some_field : String

  def submit
    validate_required some_field

    if valid?
      # Only yield the user if form is valid
      yield self, @user
    else
      yield self, nil
    end
  end
end

# Used like this
MyForm.submit do |form, user|
  if user
    # Do something
  else 
    # Do something else
  end
end
```

This can get repetitive and error-prone.

> Another real-life example is here: 
https://github.com/luckyframework/lucky_cli/pull/312/files#diff-b6e4ba970ba3c90d012a7fded94d57b2

## Proposed solution

From comment here: https://github.com/luckyframework/avram/issues/69#issuecomment-511030543

```crystal
class SearchUsers < Avram::VirtualOperation
  attribute query : String
  step { validate_size query, min: 2 }
  
  def result
    if valid? # If no validation errors on attributes
      UserQuery.search(query.value)
    else
      UserQuery.new.none
    end
  end
end

SearchUsers.run(params) do |operation, results|
  # do something
end
```

An example of the `RequestResetPassword` operation. Original here (https://github.com/luckyframework/lucky_cli/blob/d78429c8e40f97af20437cb8199de51c4962159f/src/base_authentication_app_skeleton/src/operations/request_password_reset.cr#L1)

```crystal
class RequestPasswordReset < Avram::VirtualOperation
  # You can modify this in src/operations/mixins/user_from_email.cr
  include UserFromEmail
  step validate_email

  attribute email : String

  def result
    user_from_email
  end

  def validate_email
    validate_required email
    if user_from_email?
    end
  end
end
```

### LINKED ISSUE #72: Add method to params object to check if it has key for a Form
```crystal
if params.has_key_for?(SaveUser)
  SaveUser.new(params)
else
  SaveUser.new
end
```

https://gitter.im/luckyframework/Lobby?at=5c942f1a8aa66959b64d3efb

### LINKED ISSUE #83: How to handle when the starting field value is present and an invalid param is sent
Let's say you have an `age : Int32` field.

You have a record with `age` as `32`. You then fill out an HTML form and set age to `thirty two`. Avram will correctly say that this is invalid. The problem is that because it is invalid `age.value` will be the old value `32`. I wonder what we should do? Is this correct behavior? We can't set it to `"thirty two"` because it is the wrong type. Maybe it should be `nil`? Or maybe this is just a weird edge case where there is no "correct" way to handle it

### LINKED ISSUE #276: SaveOperation.has_one should still run `valid?` on nested operation if parent is invalid
Otherwise errors will not appear for the nested one

### LINKED ISSUE #281: Rescue Lucky::MissingNestedParamError and raise helpful message
If you are trying to create a record using a SaveOperation and pass `params`, but the params are not  used, you'll get a `Lucky::MissingNestedParamError`

We  should:

* Add  a new `Avram::MissingParamKeyError`
* `Lucky::MissingNestedParamError` can inherit from it to customize  the message
*  Add helpful message to `Avram::MissingParamKeyError`

The message would look something like this:

```
The 'SaveUser' operation looked for params inside the "user" key, but there were none.

Try this...

  > If the operation doesn't need HTTP params, don't pass 'params' to the operation
  > If you need HTTP params make sure they are nested under a "user" key
  > If you want to use a different key, set the param key in 'SaveUser': https://luckyframework.org/guides/database/validating-saving#perma-param-key
```

### LINKED ISSUE #313: When you try to pass a hash to operation methods like create/update, show how to use named args instead
Related to #312 

Something like:

> You can't pass a hash directly to create/update. Instead pass named arguments, or convert the hash to params. https://link-to-guide-on-how-to-do-this

### LINKED ISSUE #331: Add after_save and before_save callbacks to Factories
```crystal
  def initialize
    after_save do
      post_id(PostFactory.create.id) unless post_id.present?
    end
  end

    after_save do |comment|
      tag = TagQuery.new.name(tag_name).first? || TagFactory.create &.name(tag_name)
      TaggingFactory.create &.comment_id(comment.id).tag_id(tag.id)
    end
  end
end
```

Note that these are *not* macros. These are available at the instance level so you get maximum flexibility. I imagine it adds Procs to an `AFTER_SAVE` and `BEFORE_SAVE` constants. The constants are an array of Procs like `Array(Proc(Comment, Nil))` and `Array(Proc(Nil, Nil))` for before save. Then on `create` it runs the array of procs

### LINKED ISSUE #336: Allow setting different param per attribute
This would be helpful for things like a search param where the common param is just “q”
 
```
attribute query : String, param_name: :q
```

### LINKED ISSUE #355: Param nesting should be optional for Avram::Operation
Currently, if you want to create an operation and pull fields off of the params, you have to make sure the fields are nested. For example:

```crystal
class Search < AvramOperation
  attribute query : String

  def submit
    puts breed.value
  end 
end
```

To use this operation, the `query` attribute has to be nested within a `search` key. In JSON, this would look like:

```json
{
  "search": {
    "query": "my query"
  }
}
```

I am currently experiencing two issues with this right now:

1. An error is raised if you initialize an operation with params and the param_key is missing.
2. It is not possible to use params if you do not want to use nested attributes

These two issues stem from wanting to use an `Avram::Operation` for a search endpoint. Ideally I would have the endpoint look like `https://www.example.com/search?q=my-search` and be able to handle when no query is passed in as well as the flat param. Instead, I have to check for the nested param before creating the operation and not pass in params if the nested param is missing.

The relevant code is here: https://github.com/luckyframework/avram/blob/6bae55c6e580e676eaa47aa3cb0baa4fbc5f3656/src/avram/define_attribute.cr#L86-L88

Instead of forcing nesting, I believe it should be optional. Unless there is a solution that I'm missing, this would seem to be a problem anytime you'd like to pass in query parameters into a GET request or similar situations.

### LINKED ISSUE #373: has_one should have automatic validator
Given model A which has_one model B, not nilable; and model B which belongs_to model A, the foreign key is in Model B. Model A does not automatically validate that Model B exists before it is saved, but it should. For context, see https://gitter.im/luckyframework/Lobby?at=5ec87bab2280c80cbfc19651

### LINKED ISSUE #377: SaveOperation fails with blank permitted_columns
Ref: https://gitter.im/luckyframework/Lobby?at=5ecc5bc8a91f120a6cc15952

Brain may be fried so I could have missed something, but I think this doesn't work:

```crystal
class Category < Avram::Model
  table do
    column color : String
    column budget : Int32?
  end
end

class SaveCategory < Category::SaveOperation
  permit_columns :color, :budget
end

params = Avram::Params.new({"category" => {"color" => "#fff", "budget" => ""}})
SaveCategory.create(params) do |operation, category|
  # category is nil here
  operation.errors[:buget] #=> "is invalid"
end
```

In this case, the column in both the migration and model were added as nilable. When the form is submitted, the param sends an empty string for the nilable column. 
The stack it takes is this:

* https://github.com/luckyframework/avram/blob/master/src/avram/needy_initializer_and_save_methods.cr#L177
* https://github.com/luckyframework/avram/blob/master/src/avram/needy_initializer_and_save_methods.cr#L218
* https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L87
* https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L133
* https://github.com/luckyframework/avram/blob/master/src/avram/charms/int32_extensions.cr#L16
* https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L138

Since this all happens on initialization, the `budget` attribute is already set with an error before any `before_save` callbacks have happened. I'm sure I'm missing something here, so I've marked this as needing investigation for now...

/cc. @franciscoadasme

### LINKED ISSUE #432: Callbacks do not run if no column attribute changed
https://github.com/luckyframework/avram/blob/cce8fee61768b3bc4cbc28fb4f038421a26e1df6/src/avram/save_operation.cr#L300

In `Avram::SaveOperation`s, `after_save` and `after_commit` callbacks do not run if no database column attribute changed. I think they should.

The current situation means that if an update operation (let's call this the child operation) is performed in any of the said callbacks of another update operation (let's call this the parent operation), the child operation is **not** guaranteed to run.

This is because the parent operation may not have any column attributes changed, so none of the said callbacks run, even though the parent operation is marked as `SaveStatus::Saved`.

Possible Use Cases:

- In a `ResetPassword` (a `User::SaveOperation`), one may want the password reset token deleted `after_commit`, whether or not the password actually changed. (Note: In this operation, the password is the only column that can change).

- In a 1:1 association, one may give a parent operation `attribute`s to be passed to a child operation, to be saved by the child in the parent's `after_save` callback. If `#update` is called on the parent but no column attribute of the parent changed, the parent would be marked as `SaveStatus::Saved` but the child operation won't run.

My proposed solution would be to update the `#save` method to:

```crystal
def save : Bool
  if valid? && (!persisted? || changes.any?)
    # ...
  elsif valid? && changes.empty?
    after_save(record.not_nil!)
    after_commit(record.not_nil!)
    true
  else
    # ...
  end
# UPDATE: 2020-07-30
# Mark as failed if a nested operation rolls back a database transaction
rescue Rollback
  mark_as_failed
  false
end
```

One could always check with `#changes` (or similar) to skip (or not) any actions to be performed in any of the said callbacks.

### LINKED ISSUE #466: Confusing error when trying to use a nilable attribute
I created a regular `Avram::Operation`, and set an attribute type to nilable

```crystal
class DoThing < Avram::Operation
  attribute prop : String?

  def submit
    #...
  end
end
```

Then I got this error message (with full `--error-trace` on)

```

 1 | require "./start_server"
     ^
Error: while requiring "./start_server"


In src/start_server.cr:1:1

 1 | require "./app"
     ^
Error: while requiring "./app"


In src/app.cr:7:1

 7 | require "./operations/**"
     ^
Error: attribute must use just one type
```

I think we can make [this error](https://github.com/luckyframework/avram/blob/master/src/avram/define_attribute.cr#L39) better. It should at least point to the Type.

jwoertink:
Overrides #363 

This will just be a saving sandbox for me to try out some refactors. The actual PRs will be broken out in to smaller ones with less commits allowing them to be easily digested. 

The end goal will be:

* Give a unified API between a SaveOperation and a non-db backed Operation
* Allow Operation to use `needs`, and callbacks
* Cleanup Operation API so the user doesn't need to define a ton of `yield` blocks
* Moves SaveOperation to a similar API. This API should be very close to how it is now, but maybe with a few minor changes.
* Removed all of the old interfaces that gave errors from deprecated previous versions
* Cleaning up a few compiler errors around operations

Currently in WIP state, so these goals could change, or increase as I get further along in this. 

Here's a list of issues that I would like to take a stab at fixing along with this. I can't guarantee all of them, but I want a single place to reference them as I go along.

* [ ] - Fix #7 
* [x] - Fix #20
* [x] - Fix #27 
* [ ] - Fix #28 
* [ ] - Fix #41 
* [ ] - Fix #43 
* [x] - Fix #69 
* [x] - Fix #72 
* [ ] - Fix #83 
* [ ] - Fix #276 
* [ ] - Fix #281 
* [x] - Fix #313 
* [ ] - Fix #331
* [ ] - Fix #336 
* [ ] - Fix #355 
* [ ] - Fix #373
* [x] - Fix #377 
* [x] - Fix #432
* [ ] - Fix #466


src/avram/callbacks.cr
```diff
@@ -1,4 +1,21 @@
...
+    def after_run(_object)
+    end
+
+    def before_save
+    end
+
+    def after_save(_record : T) forall T
+    end
+
+    def after_commit(_record : T) forall T
```
matthewmcgarvey, line 15: I see this module is included in `Avram::Operation` and I'm thinking that these callbacks don't make sense outside of database operations.  Could this module be refactored to allow declaring life cycle events that be subscribed to? Then, `Avram::Operation` could declare before and after run events and `Avram::SaveOperation` could declare the others. Something like:

```crystal
module Avram::Callbacks
  macro register_life_cycle_event(event_name, *args)
    macro {{ event_name.id }}(method_name)
      def {{ event_name.id }}(
        {% for param in args %}
          {{ param.id }},
        {% end %}
      )
        \{% if @type.methods.map(&.name).includes?(:before_run.id) %}
          previous_def
        \{% else %}
          super
        \{% end %}

        \{{ method_name.id }}
      end
    end
  end
end
```

And then in the operations:

```crystal
abstract class Avram::Operation
  include Avram::Callbacks

  register_life_cycle_event :before_run
  register_life_cycle_event :after_run, object
end

abstract class Avram::SaveOperation(T) < Avram::Operation
  register_life_cycle_event :before_save
  register_life_cycle_event :after_save, record : T
  register_life_cycle_event :after_commit, record : T
end
```

I have no idea if this actually works but this allows `Avram::Callbacks` to be a generic utility and not add methods to `Avram::Operation` that shouldn't be there.

matthewmcgarvey (commented):
Haven't gotten the chance to look through it all yet but I hope this is a useful thought

src/avram/callbacks.cr
matthewmcgarvey: I made a repo with a somewhat working example of my thoughts https://github.com/matthewmcgarvey/crystal-callbacks

jwoertink: I like the idea of subscribing to events. Right now there's quite a bit of duplication which I hope to cleanup before this PR is fully done. The one thing I'm unclear on is how we would differentiate between when a method would be called before certain things, or after. Like, this would not stop someone from writing `register_life_cycle_event :around_save`, but how would that work? I'll have to think on this more after I start to clean up some of the duplication. That might give me a little more clarity. Thanks for the suggestions!

matthewmcgarvey: @jwoertink that repo linked above has been updated to work (kind of) as you are wanting. Base classes will call `register_event :run` and then there will be macro methods for `before_save`, `after_save`, and `around_save`.  The macros got quite hairy 🙈

jwoertink: Sweet! I'll take a look. Now that I have all these specs passing, I need to start looking at cleaning up some of my duplication and adding in the new bits for this. It's about to get pretty crazy! 😂

wontruefree:
Will there need to be updates to the guides on the main website?

jwoertink:
@wontruefree yup https://github.com/luckyframework/website/issues/444 got the issue already. I'm still making some changes to the operations, so even though I'm merging in stuff, I'm trying things out and going back to redo a few things as it gets a little more clear as to how this all should work.

matthewmcgarvey:
@jwoertink You think we can close out this PR? I have no idea how much of this has been implemented and if there are any issues that can be closed out that are listed.

jwoertink:
Ok, I don't think I need this anymore. Closing it out!

Issue #379, opened 2020-05-28
Primary key is hard coded in SaveOperation's setup macro


gaku-sei:
Lucky allows to customize the type of primary key, and the name, which is really handy! But if you happen to use the type UUID, and change the column to something like "uuid" instead of "id", it will raise an error:

<img width="865" alt="Capture d’écran 2020-05-28 à 12 28 35" src="https://user-images.githubusercontent.com/3982371/83095734-f281d400-a0de-11ea-9206-344cd763781e.png">

It seems this line:
https://github.com/luckyframework/avram/blob/bb576790a832b41db10ccf8087fa9a43934a9b6f/src/avram/save_operation_template.cr#L23
Should look like `{{ primary_key_name.id }}.value ||= UUID.random()`

I'd be glad to make a pr if needed  👍

jwoertink:
Ah, good call! Yeah, if you want to do a PR, we're always happy to get those 😄

PR #380, opened 2020-05-29
Fix primary key name in SaveOperationTemplate when key type is UUID

gaku-sei:
Should fix https://github.com/luckyframework/avram/issues/379

While it compiles, and pass the tests (using `./script/test`), ~I'm not sure it actually fixes the issue (I'm very new to Crystal and I couldn't find a way to use local shards yet ^^')~ I could test it, and it seems to work 😄


jwoertink:
@gaku-sei I'd love to get this in the next release. Let me know if you have time to add a spec in for this. If not, then no worries, I can handle that. Thanks!

jwoertink:
Closing in favor of #401. Thanks for finding the fix on this @gaku-sei

PR #381, opened 2020-05-29
Fix issue with blank strings and type extensions

### LINKED ISSUE #377: SaveOperation fails with blank permitted_columns

paulcsmith:
Closes #377


spec/type_extensions/bool_spec.cr
```diff
@@ -6,4 +6,21 @@ describe Bool do
...
+    it "parses empty string as nil" do
+      parse("").should be_a(Avram::Type::SuccessfulCast(Nil))
+    end
+
+    it "parses strings" do
+      parse("true").value.should eq(true)
+      parse("1").value.should eq(true)
+      parse("false").value.should eq(false)
+      parse("0").value.should eq(false)
+    end
```
paulcsmith, line 20: Backfilled these specs

spec/type_extensions/enum_spec.cr
```diff
@@ -3,14 +3,22 @@ require "../spec_helper"
 include LazyLoadHelpers
 
 describe "Enum" do
-  it "check enum" do
+  it "parses String enum" do
+    Issue::Status::Lucky.parse("1").value.should eq(Issue::Status.new(Issue::AvramStatus::Closed))
+  end
```
paulcsmith, line 8: Added this spec to cover existing functionality

src/avram/type.cr
```diff
@@ -14,8 +14,7 @@ module Avram::Type
   def parse(values : Array(String))
     casts = values.map { |value| parse(value) }
     if casts.all?(&.is_a?(SuccessfulCast))
-      values = casts.map { |c| c.as(SuccessfulCast).value }
-      parse(values)
+      casts
```
paulcsmith, line 17: I think the original code fixed a bug in the Crystal compiler that now seems to be fixed :D

spec/type_extensions/bool_spec.cr
jwoertink: nice!

spec/type_extensions/int_spec.cr
```diff
@@ -0,0 +1,67 @@
+require "../spec_helper"
+
```
jwoertink, line 2: good call on this file 👍

src/avram/type.cr
```diff
@@ -14,13 +14,19 @@ module Avram::Type
   def parse(values : Array(String))
     casts = values.map { |value| parse(value) }
     if casts.all?(&.is_a?(SuccessfulCast))
-      values = casts.map { |c| c.as(SuccessfulCast).value }
-      parse(values)
+      # values = casts.map { |c| c.as(SuccessfulCast).value }
```
wontruefree, line 17: Should this be removed.

src/avram/where.cr
```diff
@@ -30,7 +30,7 @@ module Avram::Where
   end
 
   abstract class ValueHoldingSqlClause < SqlClause
-    getter value : String | Array(String) | Array(Int32)
+    getter value : String | Array(String) | Array(Int32) | Nil
```
jwoertink, line 33: This was needed in order to get everything to compile. Though, adding it may lead to other issues. This was also brought up in a previous comment https://github.com/luckyframework/avram/issues/27#issuecomment-486836365 But if we're going to treat `nil` as a `SuccessfulCast`, and treat empty strings as nil, then this just may be a side affect we deal with.

jwoertink:
I think this will also close #27

src/avram/charms/int16_extensions.cr
```diff
@@ -19,7 +19,14 @@ struct Int16
       SuccessfulCast(Array(Int16)).new values
     end
 
+    def parse(value : Int32)
+      SuccessfulCast(Int16).new value.to_i16
+    rescue OverflowError
+      FailedCast.new
+    end
```
matthewmcgarvey, line 26: This method is a duplicate of the one on line 35

matthewmcgarvey (changes requested):
Just the minor duplication, everything else seems fine

src/avram/charms/int16_extensions.cr
jwoertink: oh, good catch. I think one of them should probably be like Int64 and Int32...

jwoertink: actually, no, because this is Int16 specific.

matthewmcgarvey (approved):

jwoertink:
This is also related to the https://github.com/luckyframework/avram/pull/369 series.

src/avram/where.cr
jwoertink: Good news! `UserQuery.new.name(nil)` still raises a compile-time error saying to not do that. `UserQuery.new.where(:name, nil)` does work.

jwoertink:
Ok, I ran a few small tests, and it seems to still work well. It catches the compile-time errors it needs to from what I could tell. I'm gonna merge this in, and we can start playing around with it to really feel it out before next release.

bcardiff:
👋  I think this is breaking something at lucky repo. `spec/lucky/action_spec.cr` no longer compiles 

```
In spec/lucky/action_spec.cr:104:25

 104 | param required_page : Int32
                             ^
Error: method must return Int32 but it is returning (Int32 | Nil)
```

Because

```cr
  macro param
    ...
      if result.is_a? {{ type }}::Lucky::SuccessfulCast
        result.value # <-- this is nilable
      else
    ...
```

Because

```cr
struct Int32
  ...
  module Lucky
    ...
    def parse(value : String)
      return parse(nil) if value.blank? # <-- this line
      SuccessfulCast(Int32).new value.to_i
    rescue ArgumentError
      FailedCast.new
    end
```

Is this expected/known? Apologies if I am pushing things on edge #test-ecosystem

jwoertink:
🤦  ah man.... good catch. I guess my tests weren't extensive enough! I may need to revert this change then 😞  Thanks for the update @bcardiff I appreciate this a ton since it caught something this repo didn't.

Issue #367, opened 2020-05-13
create_index name collisions


BrucePerens:
These two statements will result in a collision on the name of the table generated.

create_index :url_paths, [:folder_id, :redirect_id, :forum_id, :page_id, :product_id, :manufacturer_id, :topic_id, :user_id, :post_id]
    create_index :url_paths, [:folder_id, :redirect_id, :forum_id, :page_id, :product_id, :manufacturer_id, :topic_id, :user_id, :post_id, **:path**], unique: true

The index name is truncated and the second index name thus does not include any mention that it's different from the first index in that :path has been added.

Suggested solution: don't use a truncated string of concatenated field names. The index name should be the table name followed by a numerical hash of all of the concatenated field names, perhaps in base 36 so that it is a compact alphanumeric representation.

jwoertink:
Oh, good suggestion! Thanks for reporting

jwoertink:
This is fixed by #386

PR #386, opened 2020-06-10
Custom name for index

alexneisc:
I suggest adding the ability to set custom names for the index.

Postgresql has a length limit on the index name. 

These two indexes will have the same name:
`create_index :items, [:column1, :column2, :column3, :column4, :column5, :column6, :column7, :column8]`
`create_index :items, [:column1, :column2, :column3, :column4, :column5, :column6, :column7, :column8, :column9, :column10]`

The name will be "items_column1_column2_column3_column4_column5_column6_column7_c". So when creating the second index, an error occurs:
`relation "items_column1_column2_column3_column4_column5_column6_column7_c" already exists (Exception)`

In this case, we can set a custom name for the second index and avoid the error.
`create_index :items, [:column1, :column2, :column3, :column4, :column5, :column6, :column7, :column8], name: :all_columns_index`

This can help with the problem from #367

### DIFF
--- a/src/avram/migrator/drop_index_statement.cr
+++ b/src/avram/migrator/drop_index_statement.cr
@@ -17,19 +17,26 @@ require "./index_statement_helpers"
 # DropIndexStatement.new(:users, [:email, :username] if_exists: true, on_delete: :cascade).build
 # # => "DROP INDEX IF EXISTS users_email_username_index CASCADE;"
 # ```
+#
+# For index by name:
+#
+# ```
+# DropIndexStatement.new(:users, name: :custom_index_name).build
+# # => "DROP INDEX custom_index_name;"
+# ```
 class Avram::Migrator::DropIndexStatement
   include Avram::Migrator::IndexStatementHelpers
 
   ALLOWED_ON_DELETE_STRATEGIES = %i[cascade restrict]
 
-  def initialize(@table : Symbol, @columns : Columns, @if_exists = false, @on_delete = :do_nothing)
+  def initialize(@table : Symbol, @columns : Columns? = nil, @if_exists = false, @on_delete = :do_nothing, @name : String? | Symbol? = nil)
   end
 
   def build
     String.build do |index|
       index << "DROP INDEX"
       index << " IF EXISTS" if @if_exists
-      index << " #{@table}_#{columns.join("_")}_index"
+      index << " #{index_name}"
       index << on_delete_strategy(@on_delete)
     end
   end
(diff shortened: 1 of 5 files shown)


src/avram/migrator/drop_index_statement.cr
```diff
@@ -22,14 +22,14 @@ class Avram::Migrator::DropIndexStatement
 
   ALLOWED_ON_DELETE_STRATEGIES = %i[cascade restrict]
 
-  def initialize(@table : Symbol, @columns : Columns, @if_exists = false, @on_delete = :do_nothing)
+  def initialize(@table : Symbol, @columns : Columns | Nil = nil, @if_exists = false, @on_delete = :do_nothing, @name : String? | Symbol? = nil)
   end
```
paulcsmith, line 26:
```suggestion
  @columns : Columns?
  @name : Symbol | String | Nil
  
  def initialize(@table : Symbol, @columns : Columns, @if_exists = false, @on_delete = :do_nothing)
  end
  
  def initialize(@table : Symbol, @name : String | Symbol, @if_exists = false, @on_delete = :do_nothing)
  end
```

This makes it so you can't leave both column and string blank, and makes it clear only one should be used but not both

src/avram/migrator/drop_index_statement.cr
```diff
@@ -53,4 +53,14 @@ class Avram::Migrator::DropIndexStatement
...
     end
   end
+
+  private def index_name
+    if @name
+      @name
+    elsif @columns
+      "#{@table}_#{columns.join("_")}_index"
+    else
+      raise "No name or columns specified for drop_index"
```
paulcsmith, line 63: I believe you can remove this if the the initializers ensure that one or the other is set since this will never be hit

alexneisc:
@paulcsmith Thanks for your comment! I am happy to do something useful for Lucky :)

Tomorrow I will deal with the changes. I will definitely deal with the warning about the length of the index. This is an exciting task for me

src/avram/migrator/drop_index_statement.cr
alexneisc: @paulcsmith Hm. Good point. Looks great. But there is some ambiguity.
`alias Columns = Symbol | Array(Symbol)`

If I send the parameter as a String, I get `@name`. But if I send the parameter as a Symbol, I get `@columns`. I think this can be confusing. 

I'm not sure how to do the right thing. What do you think?

paulcsmith:
@alexneisc Thank you so much! Glad it is exciting work :D

alexneisc:
@paulcsmith As I mentioned earlier, `Columns` is of type `Symbol | Array (Symbol)`. So if we want to remove index by its name, we can accidentally delete the index by column. `alias Columns = Symbol | Array(Symbol)`

So I left the previous code with a little change. Now it has become a little more compact.

Ready for suggestions for improving the code :)

paulcsmith:
Agreed. We can do some method signature cleanup later since it is more complex than I thought! Thanks for the clear explanation and the nice work on this @alexneisc

PR #396, opened 2020-06-21
[WIP] Add file attribute for uploads

wout:
This is the proposal for https://github.com/luckyframework/lucky/discussions/1195. All tests pass and it works as expected in a brand new Lucky app. It needs some polishing and drying though. You'll find some comments in the code with thoughts and concerns.

Here is a working example, ripped from the test app:

```cr
  permit_columns name
  file_attribute :avatar

  before_save do
    if file = avatar.value
      avatar.add_error("must be a .png image") unless file.filename.ends_with?(".png")
    else
      avatar.add_error("is required")
    end

    avatar.value.try do |file|
      avatar_path.value = "/path/to/#{name.value}/#{file.filename}" if avatar.valid?
    end
  end
end
```

The large majority of changes are in Avram, but `Lucky::UploadedFile` requires a few additions too. It needs to include and implement `Avram::Uploadable`, a new interface and module:

```cr
class Lucky::UploadedFile
  include Avram::Uploadable
  
  # ...
  
  def filename : String
    metadata.filename.to_s
  end

  def blank? : Bool
    filename.blank?
  end
end
```

That's required to play nice with `Avram::Attribute`, so we hardly need changes there.

### DIFF
--- a/spec/support/upload_params.cr
+++ b/spec/support/upload_params.cr
@@ -1,0 +1,16 @@
+class Avram::UploadParams < Avram::Params
+  include Avram::Paramable
+
+  @uploads : Hash(String, Avram::UploadedFile) = {} of String => Avram::UploadedFile
+
+  def initialize(@uploads)
+  end
+
+  def nested_file?(key) : Hash(String, Avram::UploadedFile)
+    @uploads
+  end
+
+  def nested_file(key) : Hash(String, Avram::UploadedFile)
+    @uploads
+  end
+end
--- a/src/avram/define_attribute.cr
+++ b/src/avram/define_attribute.cr
@@ -100,4 +100,57 @@ module Avram::DefineAttribute
       end
     end
   end
+
+  macro file_attribute(key)
+    {% unless key.is_a?(SymbolLiteral) %}
+      {% raise "file_attribute must be declared with a Symbol" %}
+    {% end %}
+
+    {% name = key.id %}
+
+    @_{{ name }} : Avram::Attribute(Avram::Uploadable?)?
+
+    ensure_base_attributes_method_is_present
+    
+    def attributes
+      ([{{ name }}] + previous_def + super).uniq
+    end
+
+    def {{ name }}
+      _{{ name }}.permitted
+    end
+
+    private def _{{ name }}
+      @_{{ name }} ||= Avram::Attribute(Avram::Uploadable?).new(
+        name: :{{ name }},
+        param: {{ name }}_param,
+        value: nil,
+        param_key: self.class.param_key
+      ).tap do |attribute|
+        if {{ name }}_param_given?
+          set_{{ name }}_from_param(attribute)
+        end
+      end
+    end
+
+    private def {{ name }}_param
+      if file = params.nested_file?(self.class.param_key)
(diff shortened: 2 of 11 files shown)


src/avram/define_attribute.cr
```diff
@@ -100,4 +100,57 @@ module Avram::DefineAttribute
       end
     end
   end
+
```
wout, line 103: There's a lot of duplication here between `attribute` and `file_attribute`. I tried extracting similar code to a separate macro, but can't make that work. I also thought about calling `attribute` inside that `file_attribute` macro, which does work, but it requires a lot of reorganisation. Before I start on that, I'd like some input first.

spec/support/upload_params.cr
```diff
@@ -0,0 +1,16 @@
+class Avram::UploadParams < Avram::Params
```
wout, line 1: I tried changing `Avram::Params` first, but kept running into many issues because it's used in many places. That's why I decided to subclass it, just to test uploads specifically.

src/avram/define_attribute.cr
```diff
@@ -100,4 +100,57 @@ module Avram::DefineAttribute
...
+  macro file_attribute(key)
+    {% unless key.is_a?(SymbolLiteral) %}
+      {% raise "file_attribute must be declared with a Symbol" %}
+    {% end %}
+
```
wout, line 108: In the `attribute` macro, there is a line with

```cr
{% ATTRIBUTES << type_declaration %}
```

I believe it's for inheritance? This is another thing I couldn't figure out. You can't create a type declaration in a macro, e.g.:

```cr
ATTRIBUTES << {{name}} : Avram::Uploadable
```

So I'll need some help here. :)

src/avram/uploadable.cr
```diff
@@ -0,0 +1,24 @@
+module Avram::Uploadable
```
wout, line 1: I was wondering, this serves both as an interface and a module. Should the interface be separated from the module?

wout:
I've been thinking about it and I believe we can tweak it so it's more intuitive:

```cr
  permit_columns name
  attribute avatar : MyLib::UploadedFile

  before_save do
    typeof(avatar.value) # => MyLib::UploadedFile
  end
end
```

If `MyLib::UploadedFile` implements the `Avram::Uploadable` interface.

src/avram/define_attribute.cr
wout: Solved!

src/avram/define_attribute.cr
wout: Solved!

src/avram/uploadable.cr
wout: I've now separated them, so that's solved as well. :)

wout:
Here is a small update. In the meantime, I reworked the `define_attribute.cr` module so it accepts file uploads as well. We no longer need for a separate `file_attribute` macro.

The next thing I want to tackle is uploads with multiple files and cast them to `Array(Avram::Uploadable)`. More on that tomorrow.

wout:
A new PR is coming up.

PR #397, opened 2020-06-22
Check for a connection before running db.create task

### LINKED ISSUE #398: Unable to check connection before creating DB
If you run the `lucky db.verify_connection` task before running `db.create` it will fail because the database doesn't exist. If you run it after the `db.create`, it's completely redundant because you just made a successful connection. Unless you run create, change your connection settings, then run verify_connection 🤔  lol...

We should have a way to verify the connection to the postgres engine itself before running create

jwoertink:
Fixes #398 

We currently have a `db.verify_connection` task, but this task only checks that you can connect to the database. When you run `db.create`, the database doesn't exist yet which means we have no way to check for the connection. 

This update checks that you can connect to the postges engine before running the db.create. This allows us to catch the error and return a nicer error message to help users figure out what they need to do.


jwoertink:
🤨  maybe time to move Avram to Gh actions too...

src/avram/tasks/db/create.cr
```diff
@@ -19,8 +19,37 @@ class Db::Create < LuckyCli::Task
...
+      hostname: Avram::Migrator::Runner.db_host.to_s,
+      username: Avram::Migrator::Runner.db_user.to_s,
+      password: Avram::Migrator::Runner.db_password.to_s,
+      port: Avram::Migrator::Runner.db_port.to_s
+    )
+  end
+
+  private def run_connection_check
+    output = IO::Memory.new
+    command = "psql -q -l #{connection_url}"
```
wontruefree, line 45: Is there a way to do this through the postgres wire protocol?

The only reason I ask is because I dont know if this will be a command available on boxes that only install the postgres client libraries.  It might be a CLI util distributed with any postrges package and in that case it is less of an issue. 

Also it feels weird to subshell out every time but create is run once so I dont think that is a big deal.

jwoertink: Not that I've found. There's tools like `pg_isready`, and a few other options, but all of them require you have the postgres client tools installed which you need to call `createdb` anyway, so you'll be guaranteed to have `psql` installed. 

We actually shell out for a lot of different commands [createdb](https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L64), [dropdb](https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L54), [restore](https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L107), and [dump](https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L118). 

If the `pg` shard ends up supporting all of these natively through a C API where we don't have to even have these CLI tools installed, I'd love to move to that. Probably a way down the road update though.

wontruefree: sounds great I was just asking :)

wontruefree: we actually might be able to reduce a lot of these calls.  There is direct SQL for them https://www.postgresql.org/docs/9.1/manage-ag-createdb.html

jwoertink: That's true. It would be good to start looking in to how we can clean all this up. After this next release I definitely want to focus on some "spring" cleaning of all the repos. Thanks for the heads up!

jwoertink:
boom! Look at that. It kept failing on Travis, but passes on Gh Actions 🌮 🎉

src/avram/tasks/db/create.cr
```diff
@@ -19,8 +20,22 @@ class Db::Create < LuckyCli::Task
   end
 
   def call
-    Avram::Migrator.run do
-      Avram::Migrator::Runner.create_db(@quiet)
+    AM::Runner.run("psql -q -l #{connection_url}", output: IO::Memory.new)
```
paulcsmith, line 23: So this just tries to connect but doesn't really do anything right? If so what do you think about extracting a private def called `attempt_to_connect`

jwoertink: Yeah, it just hops in to the psql. If it's able to, then we have a connection to the engine which means we can run db.create. I like that. makes it clear.

PR #404, opened 2020-06-24
Allow file uploads with attribute

wout:
## Purpose
This allows us to use `Lucky::UploadedFile` as a regular attribute in operations.

Related discussion:
https://github.com/luckyframework/lucky/discussions/1195

### Usage
This is what it looks like in an operation:

```cr
  permit_columns name

  attribute avatar : Lucky::UploadedFile

  before_save do
    # ...
  end
end
```

To make file uploads easier for external libraries, you can also define your own class, as long as it inherits from `Lucky::UploadedFile` OR it includes `Avram::Uploadable`:

```cr
class FancyUploadedFile
  include Avram::Uploadable
end

  # ...
  attribute avatar : FancyUploadedFile
  # ...
end
```

The returned object will be of the declared type:

```cr
SaveUser < User::SaveOperation
  attribute avatar : FancyUploadedFile
  
  before_save do
    typeof(avatar.value) # => FancyUploadedFile
  end
end
```

A complete operation with validations would look like this:

```cr
  attribute avatar : Lucky::UploadedFile

  before_save do
    if file = avatar.value
      avatar.add_error("must be a .png image") unless file.filename.ends_with?(".png")
    else
      avatar.add_error("is required")
    end

    avatar.value.try do |file|
      if avatar.valid?
        # send to S3, store locally, ...
        avatar_path.value = "/path/to/#{file.filename}"
      end
    end
  end
end
``` 

### Required other changes
In Lucky only `Lucky::UploadedFile` needs a few changes. Basically, all methods except `initialize` can go and it should include `Avram::Uploadable` which now adds the required methods:

```cr
# This class represents an uploaded file from a form
class Lucky::UploadedFile
  include Avram::Uploadable

  # Creates an UploadedFile using a HTTP::FormData::Part.
  #
  # The new file will be assigned the **name** of the
  # provided HTTP::FormData::Part and the **tempfile** will
  # be assigned the body of the HTTP::FormData::Part
  def initialize(part : HTTP::FormData::Part)
    @name = part.name
    @tempfile = File.tempfile(part.name) do |tempfile|
      IO.copy(part.body, tempfile)
    end
    @metadata = HTTP::FormData.parse_content_disposition(
      part.headers["Content-Disposition"]).last
  end
end
```

If this PR is accepted, I'll create a PR for Lucky with those changes.

Just as a reminder, here is a list of things that still need to be done:
* [ ] - Simplify `Lucky::UploadedFile` and include `Avram::Uploadable`
* [ ] - Move specs that are no longer relevant for `Lucky::UploadedFile` to a test class for `Avram::Uploadable`

### Future considerations
I'm not done with uploads, yet! :smile:  At first I wanted to include multiple file uploads as well, but I think that should go in a separate PR because it's a whole different beast. Once this is approved, I will start working on that as well.

## Checklist
* [ ] - An issue already exists detailing the issue/or feature request that this PR fixes
* [X] - All specs are formatted with `crystal tool format spec src`
* [X] - Inline documentation has been added and/or updated
* [X] - Lucky builds on docker with `./script/setup`
* [X] - All builds and specs pass on docker with `./script/test`

### DIFF
--- a/src/avram/uploadable.cr
+++ b/src/avram/uploadable.cr
@@ -1,0 +1,24 @@
+module Avram::Uploadable
+  abstract def tempfile : File
+  abstract def metadata : HTTP::FormData::FileMetadata
+  # Typically, this should return the filename as found in the `metadata`.
+  abstract def filename : String
+  # This should test if the filename is a blank string.
+  abstract def blank? : Bool
+
+  module Lucky
+    include Avram::Type
+
+    def parse(value : Avram::Uploadable)
+      SuccessfulCast(Avram::Uploadable).new(value)
+    end
+
+    def parse(values : Array(Avram::Uploadable))
+      SuccessfulCast(Array(Avram::Uploadable)).new(values)
+    end
+
+    def parse(value : String?)
+      FailedCast.new
+    end
+  end
+end
--- a/src/avram/uploadable.cr
+++ b/src/avram/uploadable.cr
@@ -1,20 +1,50 @@
+# Include this module in classes that represent an uploaded file from a form
 module Avram::Uploadable
-  abstract def tempfile : File
-  abstract def metadata : HTTP::FormData::FileMetadata
-  # Typically, this should return the filename as found in the `metadata`.
-  abstract def filename : String
-  # This should test if the filename is a blank string.
-  abstract def blank? : Bool
+  getter name : String
+  getter tempfile : File
+  getter metadata : HTTP::FormData::FileMetadata
+
+  macro included
+    # :nodoc:
+    def initialize(file : Avram::Uploadable)
+      @name = file.name
+      @tempfile = file.tempfile
+      @metadata = file.metadata
+    end
+  end
+
+  # Returns the path of the File as a String
+  #
+  # ```
+  # uploaded_file_object.path # => String
+  # ```
+  def path : String
+    @tempfile.path
+  end
+
+  # Returns the original file name as a String
+  #
+  # ```
+  # uploaded_file_object.filename # => String
+  # ```
+  def filename : String
+    metadata.filename.to_s
+  end
+
+  # If no file was selected in the form's file input, this will return `true`
+  #
+  # ```
+  # uploaded_file_object.blank? # => Bool
+  # ```
+  def blank? : Bool
+    filename.blank?
+  end
 
   module Lucky
     include Avram::Type
 
     def parse(value : Avram::Uploadable)
       SuccessfulCast(Avram::Uploadable).new(value)
-    end
-
-    def parse(values : Array(Avram::Uploadable))
-      SuccessfulCast(Array(Avram::Uploadable)).new(values)
     end
 
     def parse(value : String?)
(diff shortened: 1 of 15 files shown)


src/avram/uploadable.cr
```diff
@@ -0,0 +1,54 @@
+# Include this module in classes that represent an uploaded file from a form
+module Avram::Uploadable
+  getter name : String
+  getter tempfile : File
+  getter metadata : HTTP::FormData::FileMetadata
```
oneiros, line 5: Do you think it might be possible to abstract this metadata somewhat?

I love your approach of being able to use this module outside of `Lucky::UploadedFile`, but this means it could be used for files that arrived some other way, not necessarily via HTTP. These could of course simply construct their own `HTTP::FormData::FileMetadata`, but it would feel a bit awkward imho.

Also, avram is supposed to be a general purpose ORM. Seeing things related to HTTP here is at least a little bit confusing.

Maybe it might suffice to simply add the attributes from `HTTP::FormData::FileMetadata` that are relevant here directly (possibly size and the different timestamps). Personally I would be fine with removing this altogether. At this point we already have the complete file to determine the size reliably and I have never looked at what timestamps the browser sends with files before.

oneiros:
Thank you so much for working on this. This would be a huge step forward for file uploads in Lucky!

src/avram/uploadable.cr
wout: Very good point! We only need it for the original `filename`. I think the `metadata` property can move back to `Lucky::UploadedFile` and `Avram::Uploadable` can declare an interface for `filename`:

```cr
module Avram::Uploadable
  abstract filename : String
  # ...
end
```

So that includers can implement it however they like:

```cr
module Lucky::UploadedFile
  getter metadata : HTTP::FormData::FileMetadata
  # ...
  def filename : String
    metadata.filename.to_s
  end
end
```

I think we'll also need to rethink the `name` property as well (as discussed in https://github.com/luckyframework/lucky/issues/1147). I can't seem to find a place where it's actually used. For clarity, at least we could change it to `part_name`.

igor-alexandrov:
Guys, what is status of this PR?

jwoertink:
@igor-alexandrov it's currently in "needs to be reviewed" status. I'm sure this will be in the next major release since this is a big one.

wout:
@paulcsmith That's what I [did initially](https://github.com/luckyframework/avram/pull/404/commits/9ecc4d80dcd7c598996ffb1b2f29b58dd1924e9b). :joy: I went the extra mile because I found the current proposal to be more intuitive. And it's also more in line with what other people were expecting, too.

But I get your concern about the conditionals. To be honest, it felt a bit awkward to add the if statements there. I'll attend to it as soon as I have a bit of time.

paulcsmith:
@wout Oh I'm sorry. Maybe I missed where people felt this was more intuitive. Maybe the if statements are worth it if it makes more sense to users of the API? To me I like that I don't need to remember the file type, but maybe I'm missing a use-case!

I personally prefer the `file_attribute`, but can be swayed if people don't like that idea!

wout:
Closing this PR in favor of #428.

PR #407, opened 2020-06-25
Add associations list to SaveOperation

matthewmcgarvey:
**This adds no functionality.**

In order to complete https://github.com/luckyframework/avram/issues/385 we need to know about associations in the SaveOperation. Model was already passing in the associations, we just had to hold on to it. This is copying how SaveOperation holds on to column attributes.

## Just Talking

By adding this, there are now two sets of the same list in two different places. The Model has two lists for column attributes and associations and now SaveOperation has two lists for the same thing. The only difference I see is that SaveOperation is adding parent column attributes / associations to the lists. https://github.com/luckyframework/avram/blob/629e89d5d3463185c49e12a0bca15af74d448aad/src/avram/inherit_column_attributes.cr#L13-L17

Is there any way to consolidate the two lists?

### DIFF
--- a/src/avram/inherit_associations.cr
+++ b/src/avram/inherit_associations.cr
@@ -1,0 +1,24 @@
+module Avram::InheritAssociations
+  macro included
+    macro inherited
+      inherit_associations
+    end
+  end
+
+  macro inherit_associations
+
+    \{% if !@type.constant(:ASSOCIATIONS) %}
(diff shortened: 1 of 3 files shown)


matthewmcgarvey:
This change might not be necessary https://github.com/luckyframework/avram/pull/403#issuecomment-649553325

PR #411, opened 2020-06-27
Clone instead of mutate when changing query criteria

### LINKED ISSUE #194: Aggregate methods bork subsequent queries
If you do a `select_count` on a query, and then use that same query to do something right after, the query will most likely fail (exceptions apply).

```crystal
query = UserQuery.new.where_tasks(TaskQuery.new.completed(true))
puts query.select_count

# This line fails with PG error:
# ERROR:  column "user.id" must appear in the GROUP BY clause or be used in an aggregate function
query.first
```

This is because in the `query.first`, the SQL is now

```
SELECT COUNT(*)
FROM users
INNER JOIN tasks ON users.id = tasks.user_id
ORDER BY users.id ASC
LIMIT 1
```

It looks like [this line](https://github.com/luckyframework/avram/blob/master/src/avram/query_builder.cr#L178-L179) is the main cause since it resets the selections

### LINKED ISSUE #195: Calling first breaks subsequent aggregate queries
Along with https://github.com/luckyframework/avram/issues/194

This doesn't work:

```crystal
query = UserQuery.new.where_tasks(TaskQuery.new.completed(true))
query.first

# This line fails with
# Can't use aggregates (count, min, etc.) with limit or offset.
query.select_count
```

An easy fix might be when using `first` or `last`, since those aren't chainable, we just dup the query and run first on that. (maybe? 🤔 )

But the error should still stay in the case of `something.limit(1).select_count`.

### LINKED ISSUE #242: We should automatically clone queries after every criteria change
That way we don’t get unexpected errors like #194 #195 

The idea would be when you call any criteria method it clones the query, so every new change is a new query

bnjamin:
I would like some feedback on if this a valid path forward or if there is a better way.

There are 2 breaking changes in this PR:
- The clone method on QueryBuilder no longer takes an argument.
- queries are no longer mutated, so you need to chain your criteria calls.

Fixes #194 
Fixes #195
Fixes #242

### DIFF
--- a/spec/array_column_spec.cr
+++ b/spec/array_column_spec.cr
@@ -1,8 +1,5 @@
 require "./spec_helper"
 
-private class BucketQuery < Bucket::BaseQuery
-end
-
 describe "Array Columns" do
   it "fails when passing a single value to an array query" do
     BucketBox.new.numbers([1, 2, 3]).create
--- a/spec/query_builder_spec.cr
+++ b/spec/query_builder_spec.cr
@@ -270,8 +270,8 @@ describe Avram::QueryBuilder do
         .order_by(Avram::OrderBy.new(:id, :asc))
         .limit(1)
         .offset(2)
-      cloned_query = new_query
-        .clone(old_query)
+      cloned_query = old_query
+        .clone
         .where(Avram::Where::GreaterThan.new(:age, "20"))
         .limit(10)
         .offset(5)
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -159,10 +159,9 @@ describe Avram::Query do
       UserBox.new.name("First").create
       UserBox.new.name("Last").create
 
-      user = (user_query = UserQuery.new).first?
+      user = UserQuery.new.first?
       user.should_not be_nil
       user.not_nil!.name.should eq "First"
-      user_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users ORDER BY users.id ASC LIMIT 1"
     end
 
     it "returns nil if no record found" do
@@ -237,10 +236,9 @@ describe Avram::Query do
       UserBox.new.name("First").create
       UserBox.new.name("Last").create
 
-      last = (user_query = UserQuery.new).last?
+      last = UserQuery.new.last?
       last.should_not be_nil
       last && last.name.should eq "Last"
-      user_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users ORDER BY users.id DESC LIMIT 1"
     end
 
     it "returns nil if last record is not found" do
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -224,9 +224,10 @@ describe Avram::Query do
       UserBox.new.name("First").create
       UserBox.new.name("Last").create
 
-      last = UserQuery.last?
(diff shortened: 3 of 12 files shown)


jwoertink:
Hey @bnjamin, thanks for submitting this! It's a pretty big change to a very crucial part in the framework, so it might be a few days before one of us can really sit down and go over it. I just wanted to let you know we saw this, and appreciate the PR!

bnjamin:
Sure, that is no problem. One thing I'm struggling with is deciding between these 2 patterns(maybe there's a third even better one):

The first one is like this:

```crystal
def limit(amount) : self
  clone.limit!(amount)
end

protected def limit!(amount) : self
  query.limit(amount)
  self
end
```

This pattern is what they use in [ActiveRecord](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/relation/query_methods.rb#L375-L387). I like that it is easy to see which methods mutate and which don't.

The second one is like this:

```crystal
def limit(amount) : self
  clone.tap &.query.limit(amount)
end
```

I like that it is shorter though a bit more convoluted.

bnjamin:
> I don't know how to fix the 2 failing specs for the #first? and #last? methods, I don't see how I can capture the cloned queries to inspect the generated sql.

I removed them for now

paulcsmith:
@bnjamin Hi there! This is an incredibly important change and I am so glad you are helping us make it. I've been super busy and I want to make sure I have time to thoroughly review this. So far I think it is going in the right direction but I hope to take a deeper look soon. 

If I don't get back to you in about a week feel free to ping me. There are so many PRs to review I sometimes miss coming back to some. 

Thanks again!

stephendolan:
@paulcsmith pinging you on @bnjamin's behalf 😉  in case this did slip under the radar!

paulcsmith:
I did forget! Thank you :D

@bnjamin I think this approach is looking pretty good. I tried to think of some other ways to do it, but I think this is probably the best way forward! I thin koriginally you said you had some trouble with a couple specs, but it looks like everything is covered now. Is that correct? Any other edge cases you need help with?

If this looks good from your end, LMK and I'll do another in-depth pass and see if we can get this merged! 👍 Will be quite the breaking change, but will be worth it long term and will be much better to do now than after 1.0!

bnjamin:
I had trouble with these 2 assertions in the specs
https://github.com/luckyframework/avram/blob/master/spec/query_spec.cr#L124
https://github.com/luckyframework/avram/blob/master/spec/query_spec.cr#L193

Because we don't mutate the query anymore, I have no way to get the query that is used to generate the SQL used when calling `first?` or `last?` on `Queryable`. I just accepted that this was not possible anymore and deleted the assertions.

If this is OK I think this is good to go.

paulcsmith:
Hi @bnjamin! Just checking in to see if https://github.com/luckyframework/avram/pull/411#issuecomment-678585825 makes sense and if you think you'll be able to get to it. If not thats a-ok. Just lmk and we'll try to finish this up in the coming weeks

bnjamin:
Hi @paulcsmith, I added the missing specs using the instrumentation, it was very easy to use!

paulcsmith (approved):
Thanks so much for working on this! This will probably be tricky for people to upgrade to but I think this is a much safer and more predictable way to handle querying 🥳 

Better now than never. Thanks again!

PR #419, opened 2020-07-07
Reverting the db.create task connection test

### LINKED ISSUE #417: Lucky `db.create`, `db.reset` fail on some systems because `psql` won't run for the user, or without a database.
On my Debian system, security of Postgres may be configured more tightly than other systems. To run `psql` as a normal user (not the Postgres user), a Postgres role must exist for that Unix user AND a database with the same name as the Unix user must exist or an existing database to connect to must be specified with the `-d` argument. This could be the administration database, which is usually `postgres`, but the name of that database is configurable.

So, the Postgres existence test `attempt_to_connect` here fails:
https://github.com/luckyframework/avram/blob/3fe881faee2da2bc63ae0fed49ecccca1876b0dc/src/avram/tasks/db/create.cr#L32-L44

Because that test fails, `lucky db.reset` and `lucky db.create` fail. `lucky db.setup` uses `createdb` and succeeds.

It's probably best to remove the test, or just test that `psql` or `createdb` exist.

jwoertink:
Fixes #417 

It turns out that if your postgres username is the same as your OS username, and you don't pass a database name to `psql`, then it tries to assume you want to connect to a database of that name. If that database name doesn't exist (because it really shouldn't), then it fails. This also means that `./script/setup` fails. 

It's ok anyway, because the `create_db` method already throws errors based on connection. In this PR I'm moving those error messages to Error classes as well. This keeps all of these errors in a single location.

### DIFF
--- a/spec/tasks/db_create_spec.cr
+++ b/spec/tasks/db_create_spec.cr
@@ -3,7 +3,7 @@ require "../spec_helper"
 describe Db::Create do
   it "raises a connection error when unable to connect" do
     Avram.temp_config(database_to_migrate: DatabaseWithIncorrectSettings) do
-      expect_raises(Avram::ConnectionError, /Failed to connect to database/) do
+      expect_raises(Avram::PGNotRunningError, /It looks like Postgres is not running/) do
         Db::Create.new(quiet: true).call
       end
     end
--- a/spec/tasks/db_create_spec.cr
+++ b/spec/tasks/db_create_spec.cr
@@ -3,7 +3,7 @@ require "../spec_helper"
 describe Db::Create do
   it "raises a connection error when unable to connect" do
     Avram.temp_config(database_to_migrate: DatabaseWithIncorrectSettings) do
-      expect_raises(Avram::PGNotRunningError, /It looks like Postgres is not running/) do
+      expect_raises(Exception, /It looks like Postgres is not running/) do
         Db::Create.new(quiet: true).call
       end
     end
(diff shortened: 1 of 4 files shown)


spec/tasks/db_create_spec.cr
```diff
@@ -3,7 +3,7 @@ require "../spec_helper"
 describe Db::Create do
   it "raises a connection error when unable to connect" do
     Avram.temp_config(database_to_migrate: DatabaseWithIncorrectSettings) do
-      expect_raises(Avram::ConnectionError, /Failed to connect to database/) do
+      expect_raises(Exception, /It looks like Postgres is not running/) do
```
jwoertink, line 6: For some reason, this has to be `Exception`, which is technically correct, but when using the actual subclass, it fails. No clue why.

src/avram/errors.cr
```diff
@@ -93,4 +93,55 @@ module Avram
...
+  
+        Try this...
+  
+          ▸ If you are on macOS  you can install postgres tools from #{macos_postgres_tools_link}
+          ▸ If you are on linux you can try running #{linux_postgres_installation_instructions}
+          ▸ If you are on CI or some servers, there may already be a database created so you don't need this command"
+  
+  
+        ERROR
+      end
```
edwardloveall, line 113: Not that it's a problem, but I'm curious about why the use of `String.build` instead of passing the heredoc to `super`.

jwoertink: Because I wrote this while I was still waking up 😛  lol. It's unnecessary for sure 😂

PR #424, opened 2020-07-15
Adding default values for columns in models

### LINKED ISSUE #390: default is acting not like it suppose to
When I add a default value to a column, say a boolean it should use that if not set in operation.

Now operation fails as I haven't filled the column. So I still have to set the default value of `false` or similar. check https://github.com/kindmetrics/kindmetrics/blob/master/src/actions/sign_ups/create.cr#L5 for example.

If I set a default value I want it to be used if I don't fill it in, why even have one if not?

jwoertink:
Fixes #390 

This adds the ability to set your default values for columns in your models. If your columns require default values, you'll still have to do it in the migrations as well.

If this gets merged in, then the next step will be to update the SchemaEnforcer to ensure that the migration and model both have the same default value set.


spec/support/model_with_default_values.cr
```diff
@@ -0,0 +1,12 @@
+# columns in migration defined with default values
+# db/migrations/20180802180357_test_defaults.cr
+class ModelWithDefaultValues < BaseModel
+  table :test_defaults do
+    column greeting : String = "Hello there!"
+    column drafted_at : Time = :now
```
paulcsmith, line 6: I'm not 100% sure, but would `Time.utc` work so we don't need a special case for `:now`? The reason I *think* it would work is that we're using the value "as-is" meaning it will call `Time.utc` every time the default is called, and not just when the program is comipled. If that's the case, the `:now` handling can be removed and we can tell people to use `Time.utc`

spec/save_operation_spec.cr
```diff
@@ -355,6 +359,48 @@ describe "Avram::SaveOperation" do
...
+        params = Avram::Params.new({"greeting" => "Hi", "admin" => "true", "age" => "4", "money" => "100.23", "published_at" => 1.day.ago.to_s, "drafted_at" => 1.week.ago.to_s})
+        OverrideDefaults.create(params) do |operation, record|
+          record.should_not eq nil
+          r = record.not_nil!
+          r.greeting.should eq "Hi"
+          r.admin.should eq true
+          r.age.should eq 4
+          r.money.should eq 100.23
+          r.published_at.should be_a Time
+          r.drafted_at.should be_a Time
```
paulcsmith, line 400:
```suggestion
        published_at = 1.day.ago
        drafted_at = 2.days.ago
        params = Avram::Params.new({"greeting" => "Hi", "admin" => "true", "age" => "4", "money" => "100.23", "published_at" => published_at.to_s, "drafted_at" => drafted_at.to_s})
        OverrideDefaults.create(params) do |operation, record|
          record.should_not eq nil
          r = record.not_nil!
          r.greeting.should eq "Hi"
          r.admin.should eq true
          r.age.should eq 4
          r.money.should eq 100.23
          r.published_at.should be_a published_at
          r.drafted_at.should be_a drafted_at
```

I think it'd be good to be specific about the expectation just in case it isn't setting the time correctly. You may need to do `at_beginnning_of_second` depending on the precision of the time column. Not sure

spec/support/model_with_default_values.cr
jwoertink: In this case, my thought was that the SchemaEnforcer will eventually check that the value for the column in model and migration are the same. We support using `:now` in the migration. It makes sense in the migration since it just maps to `NOW()`, but my thought is that it would be weird setting it to `NOW()` in the db, but `Time.utc` in the model..... Or maybe that's one of those "overthinking it" moments?

spec/save_operation_spec.cr
jwoertink: Yeah, that one is a little tricky. In rails, there's the Timecop gem that's really nice for doing this kind of stuff. Should we look at using something like that? (provided it exists...), or would this suggested change be good enough?

jwoertink:
Ok, I added a spec for the SchemaEnforcer to make sure it doesn't just randomly bork. The other specs are all passing, so I don't think this will affect anything currently. Then in another PR I'd like to update that SchemaEnforcer to account for the default values better.

jwoertink:
Also, just learned this neat little trick:

This will return all of the columns that have a default value in your database.
```sql
select table_name, column_name, column_default from information_schema.COLUMNS where table_schema = 'public' and column_default is not null;
```

Screen Shot 2020-07-15 at 2 43 12 PM

This could be a good use for the schema enforcer

spec/support/model_with_default_values.cr
paulcsmith: Ah I see what you mean. I think that it is probably ok to just check that it is a time for the SchemaEnforcer. IMO not worth the complexity to make it work with a symbol, especially since I'd guess 99% of the time people set the default to now and will use `Time.utc` so will likely be in sync

spec/save_operation_spec.cr
paulcsmith: I think something like this is probably good enough (at least for now)

spec/schema_enforcer_spec.cr
```diff
@@ -46,6 +52,11 @@ describe Avram::SchemaEnforcer do
     TempTable.ensure_correct_column_mappings!
   end
 
+  it "does not raise an error when the columns have defaults" do
+    # Should not raise
+    ModelWithPerfectlyNormalColumns.ensure_correct_column_mappings!
+  end
```
paulcsmith, line 58: Nice!

spec/support/model_with_default_values.cr
jwoertink: ok, cool. I'll remove the `:now` stuff and then merge after it's all green. Thanks!

Issue #425, opened 2020-07-16
Delete Strategy for DatabaseCleaner


russ:
I've been noticing that the DatabaseCleaner.truncate method is pretty slow. Looking to the Ruby DatabaseCleaner I saw they have a delete strategy. Giving that a quick try sped my specs way up.

jwoertink:
This was merged in.

PR #426, opened 2020-07-16
Delete strategy for DatabaseCleaner

russ:
In reference to https://github.com/luckyframework/avram/issues/425


russ:
Testing this further, I see that delete will have issues if you have foreign key constraints setup on your tables. With that fact, I'm not sure how useful this might be. For my app, I'm opting to call `SomeModelQuery.delete` on different models to delete the data in the correct order.

jwoertink:
If the foreign key has an `ON DELETE CASCADE`, then it should work. `add_belongs_to author : User, on_delete: :cascade`.

russ:
Good point. I am testing on a really old rails schema and the constraints aren't 100%.

PR #428, opened 2020-07-18
Add file_attribute for file uploads.

wout:
This PR adds file uploads as @paulcsmith suggested initially in https://github.com/luckyframework/lucky/discussions/1195. That is, with a `file_attribute` macro rather than through regular attributes.

### DIFF
--- a/spec/attribute_spec.cr
+++ b/spec/attribute_spec.cr
@@ -19,6 +19,11 @@ describe "Avram::Attribute" do
       empty_array = Avram::Attribute.new(name: :empty_array, param: nil, value: [] of String, param_key: "test_form")
       empty_array.value.should_not be_nil
     end
+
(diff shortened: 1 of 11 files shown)


wout:
I'll have to refresh my memory, it's been a while...

paulcsmith:
I believe that you'd need to `include Avram::Uploadable` to `Lucky::UploadedFile` https://github.com/luckyframework/lucky/blob/master/src/lucky/uploaded_file.cr. Then I _think_ it should "just work" but you never know :P

wout:
Right, that was it. Thanks! :smile: 

I'll submit a PR today, but I think there should be at least a dedicated specs for `Lucky::UploadedFile` first.

wout:
@paulcsmith PR created!

PR #433, opened 2020-07-24
Moving Database url to credentials

### LINKED ISSUE #423: Stray newline in config results in creating extra database
Starting a new Lucky project, since my database username defaulted to my OS username rather than `"postgres"`, I replaced the appropriate line of the generated `config/database.cr` as follows (for some reason it went over my head that `.env` would be a better way to do this):

```crystal
settings.url = ENV["DATABASE_URL"]? || Avram::PostgresURL.build(
  database: database_name,
  hostname: ENV["DB_HOST"]? || "localhost",
  port: ENV["DB_PORT"]? || "5432",
  username: ENV["DB_USERNAME"]? || `whoami`, # <= oops! spot the mistake
  password: ENV["DB_PASSWORD"]? || ""        # <=
)
```

I then ran `lucky db.create` and got a very weird error (this was after manually fixing #422):

```
Unhandled exception: --: 2: --: -h: not found
 (Exception)
  from lib/avram/src/avram/migrator/runner.cr:126:7 in 'run'
  from lib/avram/src/avram/migrator/runner.cr:65:5 in 'create_db'
  from lib/avram/src/avram/tasks/db/create.cr:24:7 in 'call'
  from lib/avram/src/avram/tasks/db/create.cr:1:1 in 'print_help_or_call'
  from lib/lucky_cli/src/lucky_cli/runner.cr:27:9 in 'run'
```

Despite the error a database was created, but named after my username, instead of `myapp_development` as expected!

After some source diving, I figured out where I'd made a misstep. The backticks around `whoami` also capture the newline at the end of the command's output, splitting Avram's `createdb` command line into two separate commands:

```
createdb -U cora
 -h localhost -p 5432 myapp_development
```

The first line is treated as equivalent to `createdb cora`, and the second line generates the `-h: not found` error.

Fortunately `dropdb -U cora` (and similar) gives an error, so it's a bit harder to destroy data you weren't expecting to with `lucky db.drop`. Still, the current behavior gives a very obscure error, and also silently creates an extra unwanted database. Maybe these config values need to be rendered "shell-safe" before inserting them into a command line.

jwoertink:
Fixes #423 

This is sort of a giant refactor and a pretty big breaking change just to fix a small issue, but here's my thought behind it.

## What

This PR removes (is there a way we can deprecate habitat settings?) the `url` setting and changes it to `credentials`. This new `credentials` setting is of type `Avram::PostgresURL` instead of just `String` like it was before. This means that the value of this setting returns an object that gives us access to all of the bits like hostname, username, etc...

```crystal
# Returns nil, or returns the object
Avram::PostgresURL.parse(ENV["DB_URL"]?)

# Returns the object instead of string
credentials = Avram::PostgresURL.build(opts...)

credentials.url #=> "postgres://same/deal

# we can now set the query!
credentials.query #=> pool=5
```

Also note that if your database name is blank, this now raises an error telling you.

## Why

My thought here is that when a user sets up credentials, they may make a system call to `whoami` which could have an extra linebreak in it. Or, they may pull in data from an `ENV` leaving it up to each developer on a team to properly format the postgres url for their own machine. Since we focus on type safety, `ENV["POSTGRES_URL"] = "poo emoji"` obviously isn't valid. What if we could catch that way before trying to connect to the DB? Or what if that input has characters that might not be "command-line friendly" (i.e. &). We can add in validations for these sorts of things.

Also, when you build an app that doesn't use a database, you still may need to configure avram. Right now we just say "set it to something like 'unused'". With this, we have an official way `Avram::PostgresURL.void`.

The last thing is, after some research, I've found that postgres gives you a small performance boost by connecting through a unix socket over a TCP socket. Right now, our URL builder is designed to create a TCP socket URL. I *think* by leaving the hostname blank it will just use a unix socket.... but I'm not 100% sure. In any case, with this object we could update it to better support a unix socket out of the box. 

## Last bits

Things left to do:

* [x] - Create an interface for apps that don't use a database
* [x] - Check that this doesn't affect any of the tasks that may be missing specs
* [x] - ...
* [x] - Profit!


spec/tasks/db_verify_connection_spec.cr
```diff
@@ -2,7 +2,8 @@ require "../spec_helper"
 
 describe Db::VerifyConnection do
   it "throws a helpful error" do
-    TestDatabase.temp_config(url: "postgres://eat@joes/crab_shack") do
+    creds = Avram::PostgresURL.parse("postgres://eat@joes/crab_shack")
```
edwardloveall, line 5: 🦀

src/avram/migrator/runner.cr
```diff
@@ -13,32 +13,35 @@ class Avram::Migrator::Runner
   end
 
   def self.db_name
-    (URI.parse(database_url).path || "")[1..-1]
+    credentials.database
```
edwardloveall, line 16: Wow this is so much nicer!

src/avram/postgres_url.cr
```diff
@@ -1,39 +1,72 @@
...
   end
 
   private def set_url_port(io)
-    io << ":#{port}" unless port == ""
+    io << ":#{port}" unless port.empty?
   end
 
   private def set_url_host(io)
-    io << hostname
+    io << hostname unless hostname.empty?
```
edwardloveall, line 69: Is this the possible unix socket part?

src/avram/postgres_url.cr
```diff
@@ -1,39 +1,72 @@
...
+    @query : String = ""
   )
   end
 
-  def self.build(**args)
+  def self.build(**args) : PostgresURL
     new(**args).build
   end
 
+  def self.parse(nothing : Nil)
```
edwardloveall, line 19: I'm not sure if this is a crystal convention, but generally I'll put in an underscore if I never reference the param:
```suggestion
  def self.parse(_ : Nil)
```

src/avram/postgres_url.cr
```diff
@@ -1,39 +1,72 @@
...
       set_url_protocol(io)
       set_url_creds(io)
       set_url_host(io)
       set_url_port(io)
       set_url_db(io)
     end
+    self
+  end
+
+  private def sanitize_inputs
```
edwardloveall, line 47: The code I write, the more I want simple in-and-out (pure) functions. Less room for unexpected values and strange mutation bugs. How do you feel about moving each of these into their `set_url_<part>` method and taking care of `strip`ing or doing whatever needs to be done there? It also stops this method from doing double duty (sanitizing and validating the `@database` value).

spec/postgres_url_spec.cr
```diff
@@ -0,0 +1,21 @@
...
+      Avram::PostgresURL.parse("")
+    end
+  end
+
+  describe "parse" do
+    it "returns nil when there's nothing to parse" do
+      Avram::PostgresURL.parse(nil).should eq nil
+    end
+  end
+end
```
edwardloveall, line 21: Would it worth testing the unix command thing while you're in here considering that's what kicked off the whole refactor?

src/avram/migrator/runner.cr
jwoertink: yeah, we were calling `URI.parse` a lot, so I figure this has to be some sort of benefit this way.

src/avram/postgres_url.cr
jwoertink: Yeah. Sort of... If you're running on localhost, and do `psql -U postgres` that will connect using a unix socket, but if you do `psql -U postgres -h localhost` (which we do now), that will use the tcp socket. Now, that's only one case that this handles though. The socket path is set in some postgres configuration. The thing I'm not too sure of is, can you set a different unix socket? Like if my config is set to `/tmp/socks`, can I do point to `/tmp/shoes` ? 🤷  no clue.

spec/postgres_url_spec.cr
jwoertink: Well, this thing doesn't actually do any sort of connecting. Just building of the string. I could test that a unix socket style string is formatted. Like `postgres:///test_db`

src/avram/postgres_url.cr
jwoertink:
```
In src/avram/postgres_url.cr:19:19

 19 | def self.parse(_ : Nil)
                      ^
Error: unexpected token: UNDERSCORE
```

Turns out you can't do that 😂

edwardloveall: Boo. What about `_nothing`?

spec/postgres_url_spec.cr
edwardloveall: Mmm fair. Up to you.

jwoertink:
Boom! Solved the last few issues, so it's good now 😄 Thanks for the suggestions @edwardloveall

src/avram/postgres_url.cr
```diff
@@ -1,42 +1,82 @@
...
 
   def initialize(
     @database : String,
-    @hostname : String,
+    @hostname : String = "",
     @username : String = "",
     @password : String = "",
-    @port : String = ""
+    @port : String = "",
+    @query : String = ""
```
paulcsmith, line 11: Cold we make these nilable instead of empty strings? I wonder if that'd clarify some of the code

src/avram/postgres_url.cr
```diff
@@ -1,42 +1,82 @@
 class Avram::PostgresURL
```
paulcsmith, line 1: Thoughts on renaming this `Avram::Credentials` since that seems to be what we're calling this now? I think the "credentials" name is a good one!

jwoertink: I'm cool with that!

src/avram/postgres_url.cr
jwoertink: Yeah, we can do that 👍

paulcsmith:
Also, one Q. You said this is a big breaking change, but it seems like it should be pretty simple for people to upgrade. Is it harder than it seems for users to upgrade to this?

jwoertink:
No, it should be easy to upgrade. I meant "big" by the PR wasn't small (i.e. a few lines), and of course breaking change because you won't be able to compile. I should have said "big PR, and breaking change". 😂 

```crystal
settings.url = ENV["DB_URL"]? || Avram::PostgresURL.build()

# becomes

settings.credentials = Avram::Credentials.parse(ENV["DB_URL"]?) || Avram::Credentials.build()
```

That should be it for users.

paulcsmith:
Oh perfect! Yeah that seems totally reasonable to me!

src/avram/postgres_url.cr
jwoertink: Just realized this might make the code a little more complicated... Right now I'm stripping spaces from all of the components, but making them nilable means I have to check if there's a value before stripping on all of them

```crystal
@username = username.strip
io << username unless username.empty?

# becomes

if username && !username.empty?
  @username = username.strip

  io << username
end
```

hmm.. maybe that's fine? 🤔  After writing that, it might be clearer even though it's more lines.

jwoertink: How set are you on having these nilable? 😅  I'm fighting the compiler now..

```crystal
if username
  @username = username.to_s.strip

  # even though the compiler knows it's not nil here, the getter is still nilable
  io << URI.encode_www_form(username) unless username.not_nil!.empty?
end
```

There's a few things that get cleaned up, but a few other things that become a lot more difficult. Also, the type of port becomes `Int32?` now because that's what `URI.parse` will return.

src/avram/credentials.cr
```diff
@@ -0,0 +1,94 @@
...
+    io << hostname unless hostname.empty?
+  end
+
+  private def set_url_creds(io)
+    @username = username.strip
+    @password = password.strip
+    io << URI.encode_www_form(username) unless username.empty?
+    io << ":#{URI.encode_www_form(password)}" unless password.empty?
+    io << "@" unless username.empty?
+  end
```
paulcsmith, line 83: Maybe rather than assigning th instance var here you could do the stripping in the getter

```
def username
  @username.try(&.strip).presence
end
```

Then you can build it like:

```
username.try do |u|
  io << URI.encode_www_form(u)
end
```

jwoertink: oh, that's interesting. Let me try that.

jwoertink:
That cleanup worked out well @paulcsmith  👍

src/avram/credentials.cr
```diff
@@ -0,0 +1,125 @@
...
+    build(database: "unused")
+  end
+
+  def self.build(**args) : Credentials
+    new(**args).build
+  end
+
+  def self.parse(nothing : Nil)
+    nil
+  end
```
paulcsmith, line 24: Could this be removed? I'm worried this will lead to accidental errors or confusion where people try to parse `Nil`

src/avram/credentials.cr
```diff
@@ -0,0 +1,125 @@
+class Avram::Credentials
+  getter url : String = ""
```
paulcsmith, line 2: Rather than storing the `url` I think it'd be great to make a `url` method that builds the URL from the other isntance vars. The reason for that is that I think it is a bit easier to follow and is less likely to have bugs. For example, if you did `Avram::Credentials.new("foobar").url` it would return `nil` right now, but if `url` was a method that built the URL it would work. It could probably be memoized but I imagine it isn't used much so may not be worth it. Up to you!

I think this could work almost as-is by renaming `build` to `url`

src/avram/credentials.cr
jwoertink: The issue there is if you do `Credentials.parse(ENV["NOTHING"]?)`, but that method only takes a string, then it would always fail and not allow you to fallback to building the URL the other way. I guess we could just generate apps only using the manual builder, and add a comment that says something like "If you want to use an ENV var, use .parse" 🤔 

Let me know if that makes sense.

src/avram/credentials.cr
jwoertink: ok, I think I see what you're saying. I'm gonna make a change and push that up to see if I'm on the right track.

jwoertink:
Ok, I moved this down to using the initializer which I think looks nicer. To do this, I needed to move the building of the url in to the initializer so that way it will raise the invalid DB exception at build time. Otherwise you wouldn't know until you actually called `url` which could be buried in your app.

As for removing the `parse(nothing : Nil)` part... I was trying to think of how to handle this code: 

```crystal
credentials = Creds.parse(ENV["THAT_DOESNT_EXIST"]?) || Creds.new(...)
```

The first one has to have a way to either turn in to the object, or return `nil` so it can fallback to the other one. One thing we could do for generated apps here is something like this:

```crystal
# To load from an ENV, use:
# Creds.parse(ENV["MY_ENV"])
credentials = Creds.new(...)
```

But that would make things a little hard if you wanted to use built connection locally and ENV in production.... So we could do

```crystal
if Lucky::Env.production?
  credentials = Creds.parse(...)
else
 credentials = Creds.new(...)
end
```

We sort of do this in other configure spots, so maybe not horrible. The last suggest I guess we could do something like:

```crystal
def self.parse?(url : String?) : Credentials?
  return nil unless url

   #...
end
```

That gives us the fallback like we have now, and removes that extra empty method.

jwoertink:
Sounds good to me. Worst case is we roll this out and refactor it later I guess.

PR #434, opened 2020-07-27
Fix ameba issues

### LINKED ISSUE #431: Add Ameba
I tried to run Ameba locally and here is what I got:

```
Finished in 795.01 milliseconds

257 inspected, 74 failures.
```

What do you think about added Ameba to Avram?

igor-alexandrov:
Closes #431 

The first commit fixes all Ameba issues. The second commit adds Ameba to GitHub Actions, unfortunately I wasn't able to do it with official Ameba Action (https://github.com/crystal-ameba/github-action) because of this issue: https://github.com/crystal-ameba/github-action/issues/8. But we can always revisit configuration later.

### DIFF
--- a/spec/define_attribute_spec.cr
+++ b/spec/define_attribute_spec.cr
@@ -103,9 +103,8 @@ describe "attribute in operations" do
       operation.terms_of_service.value.should be_nil
       post.should_not be_nil
 
-      # ameba:disable Lint/ShadowingOuterLocalVar
-      Operation.update(post.not_nil!, best_kind_of_bear: "koala bear") do |operation, _post|
-        operation.best_kind_of_bear.value.should eq("koala bear")
+      Operation.update(post.not_nil!, best_kind_of_bear: "koala bear") do |op, _post|
+        op.best_kind_of_bear.value.should eq("koala bear")
       end
     end
   end
--- a/spec/define_attribute_spec.cr
+++ b/spec/define_attribute_spec.cr
@@ -103,7 +103,8 @@ describe "attribute in operations" do
       operation.terms_of_service.value.should be_nil
       post.should_not be_nil
 
-      Operation.update(post.not_nil!, best_kind_of_bear: "koala bear") do |operation, post|
+      # ameba:disable Lint/ShadowingOuterLocalVar
+      Operation.update(post.not_nil!, best_kind_of_bear: "koala bear") do |operation, _post|
         operation.best_kind_of_bear.value.should eq("koala bear")
       end
     end
--- a/spec/preloading_spec.cr
+++ b/spec/preloading_spec.cr
@@ -98,7 +98,7 @@ describe "Preloading" do
       comment = CommentBox.create &.post_id(post.id)
 
       # ameba:disable Lint/UselessAssign
-      should_not_be_preloaded = CommentBox.create &.post_id(another_post.id)
+      _should_not_be_preloaded = CommentBox.create &.post_id(another_post.id)
 
       posts = Post::BaseQuery.new.preload_comments.limit(1)
 
@@ -124,8 +124,7 @@ describe "Preloading" do
   it "preloads has_many through" do
     with_lazy_load(enabled: false) do
       tag = TagBox.create
-      # ameba:disable Lint/UselessAssign
-      unused_tag = TagBox.create
+      _unused_tag = TagBox.create # ameba:disable Lint/UselessAssign
       post = PostBox.create
       other_post = PostBox.create
       TaggingBox.create &.tag_id(tag.id).post_id(post.id)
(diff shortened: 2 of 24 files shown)


spec/preloading_spec.cr
```diff
@@ -96,7 +96,9 @@ describe "Preloading" do
       post = PostBox.create
       another_post = PostBox.create
       comment = CommentBox.create &.post_id(post.id)
-      _should_not_be_preloaded = CommentBox.create &.post_id(another_post.id)
+
+      # ameba:disable Lint/UselessAssign
```
wontruefree, line 100: why is this disabled?

src/avram/schema_enforcer.cr
```diff
@@ -180,14 +182,16 @@ module Avram::SchemaEnforcer
...
+            class #{model_class} < BaseModel
               # Great for models used in migrations, or for legacy schemas
               skip_schema_enforcer
             end
 
 
         TEXT
       end
+
+      message
```
wontruefree, line 194: Is this added for clarity?

src/avram/tasks/db/verify_connection.cr
```diff
@@ -20,30 +20,28 @@ class Db::VerifyConnection < LuckyCli::Task
   end
 
   def call
```
wontruefree, line 22: Was this a tabs/spaces change?

spec/preloading_spec.cr
```diff
@@ -122,7 +124,8 @@ describe "Preloading" do
   it "preloads has_many through" do
     with_lazy_load(enabled: false) do
       tag = TagBox.create
-      _unused_tag = TagBox.create
+      # ameba:disable Lint/UselessAssign
+      unused_tag = TagBox.create
```
paulcsmith, line 128: Is there a way to tell ameba to ignore variables starting with `_`? That's how we name variables that aren't used but are there just to clarify what they do.

If we can't do that then I think it'd be better to add a comment: `# unused tag` instead of disabling the rule

spec/support/lazy_load_helpers.cr
```diff
@@ -1,4 +1,5 @@
 module LazyLoadHelpers
+  # ameba:disable Style/RedundantBegin
```
paulcsmith, line 2: Could this be changed to just `ensure` without the `begin` so we don't need to disable the rule?

spec/define_attribute_spec.cr
```diff
@@ -103,7 +103,8 @@ describe "attribute in operations" do
       operation.terms_of_service.value.should be_nil
       post.should_not be_nil
 
-      Operation.update(post.not_nil!, best_kind_of_bear: "koala bear") do |operation, post|
+      # ameba:disable Lint/ShadowingOuterLocalVar
+      Operation.update(post.not_nil!, best_kind_of_bear: "koala bear") do |operation, _post|
         operation.best_kind_of_bear.value.should eq("koala bear")
```
paulcsmith, line 108: Maybe these could be solved by naming the var `op`?

igor-alexandrov: Yes, renaming to `op` will solve `Lint/ShadowingOuterLocalVar` here and in a bunch of other places.

spec/preloading_spec.cr
igor-alexandrov: For now Ameba does not ignore variables with `_`, so I had to use `# ameba:disable Lint/UselessAssign` comment.

src/avram/tasks/db/verify_connection.cr
igor-alexandrov: Maybe it was tabulated with tabs and this changed automatically when redundant `begin` block has been removed.

src/avram/schema_enforcer.cr
igor-alexandrov: Yes, I think it looks more obvious than a 30-line if/else block.

spec/preloading_spec.cr
```diff
@@ -122,7 +124,7 @@ describe "Preloading" do
   it "preloads has_many through" do
     with_lazy_load(enabled: false) do
       tag = TagBox.create
-      _unused_tag = TagBox.create
+      _unused_tag = TagBox.create # ameba:disable Lint/UselessAssign
```
jwoertink, line 127:
```suggestion
      TagBox.create # unused tag
```

spec/preloading_spec.cr
```diff
@@ -140,7 +142,7 @@ describe "Preloading" do
       item = LineItemBox.create
       other_item = LineItemBox.create
       product = ProductBox.create
-      _unused_product = ProductBox.create
+      _unused_product = ProductBox.create # ameba:disable Lint/UselessAssign
```
jwoertink, line 145:
```suggestion
      ProductBox.create # unused product
```

spec/preloading_spec.cr
```diff
@@ -96,6 +96,8 @@ describe "Preloading" do
       post = PostBox.create
       another_post = PostBox.create
       comment = CommentBox.create &.post_id(post.id)
+
+      # ameba:disable Lint/UselessAssign
       _should_not_be_preloaded = CommentBox.create &.post_id(another_post.id)
```
jwoertink, line 101:
```suggestion
      CommentBox.create &.post_id(another_post.id) # should not be preloaded
```

igor-alexandrov:
@jwoertink sometimes the most obvious way of reaching the goal is so hard to find :) Thanks.

Issue #393, opened 2020-06-16
Default columns should generate values from postgres


jwoertink:
If you're using something like Hasura to connect through Lucky, the GraphQL may need to insert records in to postgres at some point. If you happen to have Avram setup with UUID, then your [ID](https://github.com/luckyframework/avram/issues/122), and the [timestamps](https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L327-L328) are all generated from Crystal. This means that inserting a record through GraphQL fails.

To temporarily fix this, you can create a migration that runs this:

```crystal
def migrate
  enable_extension "uuid-ossp" # NOTE: this requires Avram > 0.15? (as of today, it isn't released yet)
  execute("ALTER TABLE users ALTER COLUMN id SET DEFAULT uuid_generate_v4();")
  execute("ALTER TABLE users ALTER COLUMN created_at SET DEFAULT CURRENT_TIMESTAMP;")
  execute("ALTER TABLE users ALTER COLUMN updated_at SET DEFAULT CURRENT_TIMESTAMP;")
end
```

/cc. @russ  @KCErb

KCErb:
@jwoertink thanks for finding this and the fix! I've been planning to take advantage of Live Queries/subscriptions from GraphQL and not so much mutations so I wouldn't have caught this for a while (I think). Thank you!

jwoertink:
Yeah, @russ is actually the one that found it. He's using your lucky-hasura deal. 😃

KCErb:
Really? That's great I had no idea!

Thanks @russ please share your notes with me as bump into rough edges. You're apparently now more experienced than I am in my own project 😲 !

russ:
@KCErb haha. Not really but your repo is helping me learn all of this stuff. The only other issue I've run into is cors related stuff but I can submit a pull request for that on your repo. I'll let you know over there if I run into anything else.

paulcsmith:
@jwoertink The only thing missing is the `updated_at` needs a trigger and function to set the timestamp when the row is updated.

Also I *think* the UUID is now generated by default, but if not I agree it should be 👍

matthewmcgarvey:
Just as a note `uuid-ossp` has a note in their docs:

```
If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.
```

matthewmcgarvey:
[...]

Issue #337, opened 2020-03-23
Consider using a trigger for timestamps


paulcsmith:
https://x-team.com/blog/automatic-timestamps-with-postgresql/

bdtomlin:
I like it! One of the great things about being committed to PostgreSQL is the ability to do stuff like this.

paulcsmith:
Agreed!! This is really why I don't think Avram will ever support other databases. Too many goodies in PG and I don't want to be slowed down with figuring out multiple ways to do everything.

I'd like to document how to use other ORMs if you *really* need mysql or something (legacy databases, corp reqs, etc.) but for now though just PG seems to be a good way to go!

PR #436, opened 2020-07-31
Adding new triggers feature to generate postgres triggers.

jwoertink:
This is a stepping stone to closing out 

https://github.com/luckyframework/avram/issues/337
https://github.com/luckyframework/avram/issues/393

This PR introduces the ability to set simple triggers. As with most Postgres things, it can get super hairy, so if you need something complex, it's best to hop out of Avram anyway. These should cover the 90% use case though.

This create a BEFORE UPDATE trigger called "trigger_set_timestamp" that will call the "set_timestamp()" function

```crystal
create_trigger table_for(User), "set_timestamp"
```

You can create a trigger for BEFORE/AFTER and on INSERT/UPDATE/DELETE by passing some options:

```crystal
create_function "update_counts", <<-SQL
  NEW.some_count = OLD.some_count + 1
  RETURN NEW;
SQL

create_trigger :users, "update_counts", callback: :after, on: [:insert, :update]
```

### DIFF
--- a/src/avram/migrator/create_trigger_statement.cr
+++ b/src/avram/migrator/create_trigger_statement.cr
@@ -1,0 +1,25 @@
+class Avram::Migrator::CreateTriggerStatement
+  def initialize(
+    @table_name : Symbol,
+    @trigger_name : String,
+    @function : String,
+    callback @trigger_when : Symbol = :before,
+    on @trigger_operation : Array(Symbol) = [:update]
+  )
+  end
+
+  def build
+    <<-SQL
+    CREATE TRIGGER #{@trigger_name}
+    #{@trigger_when.to_s.upcase} #{operation_statement} ON #{@table_name}
+    FOR EACH ROW
+    EXECUTE PROCEDURE #{@function}();
+    SQL
+  end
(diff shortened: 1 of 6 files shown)


src/avram/migrator/create_trigger_statement.cr
```diff
@@ -0,0 +1,25 @@
+class Avram::Migrator::CreateTriggerStatement
+  def initialize(
+    @table_name : Symbol,
+    @trigger_name : String,
+    @function : String,
+    callback @trigger_when : Symbol = :before,
+    on @trigger_operation : Array(Symbol) = [:update]
```
jwoertink, line 7: I originally wanted to use Enum instead of Symbol for both of these so we could validate that you only pass `:before` or `:after` and `:insert, :update, :delete`. However, crystal doesn't support using enum inside of an array and passing an array of Symbol. You'd have to pass an Array of that enum which would get pretty gnarly.

paulcsmith: Makes sense. I think the symbols will work well here

PR #439, opened 2020-08-18
Adds build method - helper to get types

Xosmond:
Hello! Thanks for maintaining Avram!

I wanted to generate a named couple to send it as params so I had to write a macro for that, also I had to define the params types to I created an alias for that, also I wanted to pass attributes overwriting so I modify some methods.

Is there a different way to do this? Is this code useful for you? Let me know what do you think. I have been working with this together with `Minitest.cr` and works smoothly 👍


Xosmond:
Hey @paulcsmith 

Sorry for the bad description of the PR

This is a summary with some examples:

**1. Params and Types**
You can use this to send it as params to an endpoint or you can serialize it to json for other purposes on testing

```crystal
params = SomeBox.params
```

Here you can use PARAMS_TYPES to define the types when needed, useful in some contexts

```crystal
@params : SomeBox::PARAMS_TYPES = SomeBox.params
```

**2. Params, create, new receive a named_tuple to overwrite attributes easily**

```crystal
# Using a block would be big and hard to read
params  = SomeBox.params do |box|
  box.name = "overwrite name"
  box.other = "overwrite other"
end

# I don't really like this way either
SomeBox.params(&.name("not-a-sequence"))

# Sending a named_tuple would be clean and easy to read
params = SomeBox.params({ name: "overwrite name", other: "overwrite other"})

some = SomeBox.create({ name: "overwrite name", other: "overwrite other"})

some = SomeBox.new({ name: "overwrite name", other: "overwrite other"}) # would be like build?


```

Xosmond:
I think NameTuple can be a separate discussion then :). I think using blocks for small and simple boxes is fine, but for NamedTuples would be pretty useful for testing big and complex applications and when to reuse and change constantly things.

I have created this https://github.com/luckyframework/avram/pull/449 for that, since I can't change the branch here to merge here.

Thanks for your comments

PR #54, opened 2019-02-10
Add support for `OR`-joining of clauses in queries

HarrisonB:
Closes #31.

PR #442, opened 2020-08-21
Adding in typesafe WHERE OR

### LINKED ISSUE #31: Allow `or`-ing of where conditions
Forgive me if I am just not seeing the obvious! I would like to do

```crystal
PostQuery.new.title("A").or.content("B")
```
and expecting something like `WHERE title = "A" OR content = "B"`.

jwoertink:
Supersede #54
Fixes #31

Ok, a few things going on here....

This PR adds in BASIC `OR` query usage. As we all know, doing `OR` queries can get SUPER complicated. This implementation does not take in to account precedence in conditions. I do have a plan for that, but would like to do that in another PR. Right now this just changes the `AND` to `OR`, and that's it. 

We can now do things like:

```crystal
# SELECT * FROM users WHERE age > 20 OR age < 30 OR age = 52 AND name = 'Eilert'
UserQuery.new
  .age.gt(20)
  .or(&.age.lt(30))
  .or(&.age(52).name("Eilert"))
```

For more complex queries, you still have the escape hatch with a raw where `where("this = ? OR (that = ?)", 1, 2)`


matthewmcgarvey:
I've been giving this some thought and I really don't think this should go in as it is. There are so many ways this can produce invalid queries that I think it would put patterns into the codebase that maybe shouldn't be there. As you've already said, you know that precedence is an issue so I'm not going to point out any of that stuff. Here's a list of potential refactorings that could happen beforehand to make this change easier:

- unify wheres and raw wheres - right now they are two separate lists and and raw wheres are appended when making the sql statement (which I think there is a bug in this PR that you are updating the last where in both lists) but when `or` is introduced, the order that where clauses are added will matter
- Extract a generic type for the unified wheres list that allows us to represent the different type of where conditions without having to extend from the abstract `SqlClause` class. We'll need to introduce parenthesized conditions (`name = 'foo' AND (age = 1 OR age = 2)`) to allow effective use of `OR` and the parenthesized condition is not represented by `SqlClause`. The shared type probably only needs the `prepare` method. We could use a module with an abstract method instead of inheritance
- Update how equality between where conditions is found. Right now it is assumed that the where condition [will have a value](https://github.com/luckyframework/avram/blob/b002ead126de2e4e47c1ca9c2192ad86db91f844/src/avram/query_builder.cr#L326) but that will not be the case when the parenthesized condition is added. (Maybe we can switch to using Structs to get equality for free?)
- Right now query code is maintained in both `Queryable` and `QueryBuilder`. We could figure out a way to always try to move to a `QueryBuilder` so that `Queryable` just has the entry-level methods to start making a query
- Add precedence rule for `AND` instead of just doing a join on the where clauses. If a condition is a single expression or a group of expressions, my thinking would be that once we incorporate the ability in the codebase to parenthesize (or group) expressions, each where clause call is given the last condition and can either add to the last condition (representing an `AND`) or makes a new condition (which would be an `OR`) when that ability is added. Which this methodology `WHERE name = 'bill' or name = 'bob' and age = 42` would become `((name = bill), (name = bob, age = 42))`. This is definitely not clear but I'm pretty sure it works

jwoertink:
For the parenthesis, we will most likely add a new method that will wrap whatever query you add in to it. I'm not sure exactly how this looks, but there's a few other frameworks (I think Laravel is one) that already do this.

```crystal
# WHERE (name = 'Bill' OR name = 'Bob')
UserQuery.new.wrap_query(&.name("Bill").or(&.name("Bob")))
```

In #54 there was an abstraction in to a class for the Where stuff so they weren't floating in separate lists, but I couldn't get that PR to work without major refactors. My idea with this PR was write minimal code to get a minimal feature.

I think this PR as-is is probably fine as long as it doesn't bork any current existing queries. Once it's in, the next PR would add in the parenthesis part. Then the PR after that one could be a refactor of how it all works. But if there are any queries that work in master branch now that would break with this PR, let me know. I can add in a spec for that to ensure that doesn't happen.

I'll hold off with this until @paulcsmith has some time to review. In the meantime, if you have an idea for an easier implementation, feel free to open up a PR or you can comment in here with some code samples and I can try those out. Thanks @matthewmcgarvey!

paulcsmith:
> Right now query code is maintained in both Queryable and QueryBuilder. We could figure out a way to always try to move to a QueryBuilder so that Queryable just has the entry-level methods to start making a query

I think this is what Queryable is already doing. It is mostly a thin type-safe wrapper over QueryBuilder. So QueryBuilder is the low-level part and Queryable is the one that you typically use from your query object.

I think the other points are good, especially about how wheres and raw_wheres are split up.

I think I tend to agree with @jwoertink though that this can go in as-is now and then we can build off of it. We can hold off on documenting it in the guides until we feel comfortable with it. That way people *can* use this now, but with the caveat that it might be changing very soon!

Also I like your ideas so once this is in if you've got some PRs or something that'd be great. My time has been almost zero lately so don't have as much time to implement those suggestions quite yet

src/avram/query_builder.cr
```diff
@@ -309,6 +309,17 @@ class Avram::QueryBuilder
...
   end
 
+  def or(&block : Avram::QueryBuilder -> Avram::QueryBuilder)
+    if @wheres.empty? && @raw_wheres.empty?
+      raise Avram::InvalidQueryError.new("Cannot call `or` before calling a `where`")
+    end
+
+    @wheres.last.conjunction = Avram::Where::Conjunction::Or if @wheres.any?
+    @raw_wheres.last.conjunction = Avram::Where::Conjunction::Or if @raw_wheres.any?
+    block.call(self)
```
paulcsmith, line 319: I think there is a bug here that @matthewmcgarvey described. In that there could be doulbe `ors` on accident. For example `MyQuery.new.where("age > 30").name("Bob").or(&.name("Jane"))` it would add the `or` to *both* wheres.

I'm almost thinking we should make `or` ONLY work for `@wheres`. If you want to use an `or` with a raw where, then you can use a raw where to do it `.where("name $1 OR $2")`! What do you think? I think that is less bug-prone and kinda makes sense anyway since the `or` is on queryable anyway

paulcsmith (commented):
I think if this removes `or` for `raw_wheres` then this should be ok. I think `raw_wheres` should be special and basically untouched. I think that'll remove some of the bugs and edge-cases but still be useful for regular `wheres`

Thoughts?

src/avram/query_builder.cr
jwoertink: I like the idea of making the `or` only work on where and not raw_where. Though, I saw #460 which looks nice and would fix the issue of the ordering. But if we merged that one in, then I'm not sure how that would affect us being able to do different things on `where` vs `raw_where` since that PR would make `raw_where` deprecated 🤔

src/avram/query_builder.cr
```diff
@@ -298,17 +298,35 @@ class Avram::QueryBuilder
...
   end
 
+  def or(&block : Avram::QueryBuilder -> Avram::QueryBuilder)
+    if @wheres.empty?
+      raise Avram::InvalidQueryError.new("Cannot call `or` before calling a `where`")
+    end
+
+    @wheres.last.conjunction = Avram::Where::Conjunction::Or
+
+    block.call(self)
```
jwoertink, line 308: @matthewmcgarvey maybe you can see what's up here, but I'm not sure what I'm missing? 

If you look at the failing spec, you'll see that it's expecting `users.age` to be there 3 times, but it's only there once. This is the line where it calls the block which should hook in to the existing QueryBuilding to append `age` the second and third time. This worked prior to merging in master, but now it doesn't 🤔

src/avram/query_builder.cr
```diff
@@ -298,17 +298,35 @@ class Avram::QueryBuilder
...
   @_wheres_sql : String?
 
   private def wheres_sql
     @_wheres_sql ||= joined_wheres_queries
   end
 
   private def joined_wheres_queries
-    if wheres.any?
-      statements = wheres.map(&.prepare(->next_prepared_statement_placeholder))
+    if @wheres.any?
```
jwoertink, line 319: this should actually call the method instead of instance variable, but I was trying to eliminate the `uniq` in case that's what was causing the spec to fail. That doesn't seem to be the case though....

src/avram/query_builder.cr
matthewmcgarvey: @jwoertink I guess this PR was made after the update to immutable QueryBuilder/Queryable. You need to return this line, not the one below it.  The same in the other or method.

src/avram/query_builder.cr
matthewmcgarvey: The two uses of `@wheres` in here are causing one of the specs to fail. Once this and the other changes are made, all specs should be passing.

matthewmcgarvey (changes requested):

src/avram/query_builder.cr
jwoertink: Ah! You're right, I forgot about that. I'll give that a shot. Thanks!

matthewmcgarvey (approved):

PR #448, opened 2020-08-24
Fixed passing an empty string to a nilable type through params

### LINKED ISSUE #20: Issues with blank strings as params

### LINKED ISSUE #447: Empty field for type Time is invalid.
I have a table with this column `column date_certif : Time?` and action with a save query `agency = SaveAgency.create!(params)`

When I pass a param like `agency:date_certif=2020-08-01`, everything works great but, when I try to empty the column like this `agency:date_certif=""` I have the following error `date_certif: is invalid`.
It’s invalid because date_certif only accepts `Time | Nothing | Nil`.

For now to fix this I must do:
```
date_certif = params.get?("agency:date_certif")
date_certif = date_certif && date_certif.present? ? Time.parse_local(date_certif, "%F") : nil
agency = SaveAgency.create!(params, date_certif: date_certif)
```

jwoertink:
Fixes #447
Fixes #20 

If you pass a permitted column that's nilable through params as an empty string, then it will fail to cast that like in the case of `Time?` being parsed https://github.com/luckyframework/avram/blob/master/src/avram/charms/time_extensions.cr#L27. So in this case, if the type is nilable, and we send an empty value, just return nil instead of assigning a value. 

One case to note here that we may need to handle separately would be if you defined a `String?` and were ok with this being an empty string. In this case, you should just make your column NOT nilable, but as of right now we don't tell you that.

### DIFF
--- a/spec/save_operation_spec.cr
+++ b/spec/save_operation_spec.cr
@@ -79,16 +79,13 @@ describe "Avram::SaveOperation" do
     bucket.names.should eq([] of String)
   end
 
-  it "ignores empty params on nilable fields" do
+  it "treats empty strings as nil for Time? types instead of failing to parse" do
     avram_params = Avram::Params.new({"title" => "Test", "published_at" => ""})
 
-    SavePost.create(avram_params) do |operation, post|
-      operation.errors.should be_empty
-      operation.valid?.should eq true
-      post.should_not be_nil
-      post.not_nil!.published_at.should eq nil
-      post.not_nil!.title.should eq "Test"
-    end
+    post = SavePost.create!(avram_params)
+
+    post.published_at.should eq nil
+    post.title.should eq "Test"
   end
 
   describe ".create" do
--- a/spec/save_operation_spec.cr
+++ b/spec/save_operation_spec.cr
@@ -53,6 +53,10 @@ private class OverrideDefaults < ModelWithDefaultValues::SaveOperation
   permit_columns :greeting, :drafted_at, :published_at, :admin, :age, :money
 end
 
+private class SavePost < Post::SaveOperation
+  permit_columns :title, :published_at
+end
+
 describe "Avram::SaveOperation" do
   it "allows overriding the param_key" do
     ParamKeySaveOperation.param_key.should eq "custom_param"
@@ -73,6 +77,18 @@ describe "Avram::SaveOperation" do
     bucket = Bucket::SaveOperation.update!(bucket, names: [] of String)
 
     bucket.names.should eq([] of String)
+  end
+
+  it "ignores empty params on nilable fields" do
+    avram_params = Avram::Params.new({"title" => "Test", "published_at" => ""})
+
+    SavePost.create(avram_params) do |operation, post|
+      operation.errors.should be_empty
+      operation.valid?.should eq true
+      post.should_not be_nil
+      post.not_nil!.published_at.should eq nil
+      post.not_nil!.title.should eq "Test"
+    end
   end
 
   describe ".create" do
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -153,6 +153,11 @@ abstract class Avram::SaveOperation(T) < Avram::Operation
       end
 
       def set_{{ attribute[:name] }}_from_param(_value)
+        {% if attribute[:nilable] %}
+          # The type is nilable, but if the value is blank, it will return a FailedCast
+          # in types like `Time?` or `Int32?`, etc...
+          return nil if _value.blank?
+        {% end %}
         {% if attribute[:type].is_a?(Generic) %}
           # Pass `_value` in as an Array. Currently only single values are supported.
           # TODO: Update this once Lucky params support Arrays natively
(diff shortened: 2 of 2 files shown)


spec/save_operation_spec.cr
```diff
@@ -75,6 +79,18 @@ describe "Avram::SaveOperation" do
     bucket.names.should eq([] of String)
   end
 
+  it "ignores empty params on nilable fields" do
```
paulcsmith, line 82: I'm not 100% sure I understand the "ignores" part. Does this mean that it treats empty params as nil without erroring? Just trying to make sure I get the behavior

spec/save_operation_spec.cr
```diff
@@ -75,6 +79,18 @@ describe "Avram::SaveOperation" do
...
+    avram_params = Avram::Params.new({"title" => "Test", "published_at" => ""})
+
+    SavePost.create(avram_params) do |operation, post|
+      operation.errors.should be_empty
+      operation.valid?.should eq true
+      post.should_not be_nil
+      post.not_nil!.published_at.should eq nil
+      post.not_nil!.title.should eq "Test"
+    end
```
paulcsmith, line 91:
```suggestion
    post = SavePost.create!
    post
    post.published_at.should eq nil
    post.title.should eq "Test"
    end
```

src/avram/save_operation.cr
```diff
@@ -153,6 +153,11 @@ abstract class Avram::SaveOperation(T) < Avram::Operation
       end
 
       def set_{{ attribute[:name] }}_from_param(_value)
+        {% if attribute[:nilable] %}
+          # The type is nilable, but if the value is blank, it will return a FailedCast
+          # in types like `Time?` or `Int32?`, etc...
```
paulcsmith, line 158: Nice. This is a super helpful comment

src/avram/save_operation.cr
```diff
@@ -153,6 +153,11 @@ abstract class Avram::SaveOperation(T) < Avram::Operation
...
+          return nil if _value.blank?
```
paulcsmith, line 159: Actually I think we should change this slightly. If `_value.blank?` then I think it should always return `nil` and never attempt parsing. Otherwise when you send an empty string for a `Time` it will have both a "is missing" and a "is invalid". 

If we always return `nil` then it would just have "is required" which is the expected behavior IMO

spec/save_operation_spec.cr
jwoertink: yeah, it's bad wording... Maybe `it "doesn't fail with empty string params on nilable fields"`?

src/avram/save_operation.cr
jwoertink: oh, interesting point. Yeah, I think that will work. I'll try that

spec/save_operation_spec.cr
paulcsmith: I like that I think adding what it does might clarify it even more. `" treats empty strings as `nil` for Time types instead of failing to parse"`?

src/avram/save_operation.cr
jwoertink: Actually... Doing that causes 2 of the specs to fail. This kind of makes me think there may be a behavior change that's not expected... Think I should just fix those specs? or just roll with what I have now?

```
1) Avram::SaveOperation .update on failure yields the operation and nil
     Failure/Error: operation.save_failed?.should be_true

       Expected: true
            got: false

     # spec/save_operation_spec.cr:531

  2) Avram::SaveOperation .update! on failure raises an exception
     Failure/Error: expect_raises Avram::InvalidOperationError do

       Expected Avram::InvalidOperationError but nothing was raised

     # spec/save_operation_spec.cr:584
```

paulcsmith (approved):
Looks good!

PR #449, opened 2020-08-29
Add build_attributes to box model

Xosmond:
Discussion on https://github.com/luckyframework/avram/pull/439

This PR is for adding `build_attributes` to box model.


spec/box_spec.cr
```diff
@@ -5,6 +5,18 @@ describe Avram::Box do
     PlainModelBox.create.id.should_not be_nil
   end
 
+  describe "build_attributes" do
+    it "generate a named_tuple with attributes" do
+      attributes = TagBox.build_attributes
+      attributes.should eq({custom_id: nil, created_at: nil, updated_at: nil, name: "name-8"})
```
jwoertink, line 11: Does this spec still pass if this is the only spec you run? Since this attribute uses `sequence`, I would think that it wouldn't always return 8.

Xosmond: You are completely right, I'm not much familiar with sequences feature, btw I just added a reset to avoid failure if somebody add a new test for tag box.

jwoertink:
Awesome! good call on that sequences. This is gonna be a great addition. 🚀

PR #451, opened 2020-09-05
Handle duplicate joins

### LINKED ISSUE #370: Querying with a join table can produce invalid SQL
## Summary

In this instance I'm working with 3 models
- Business
- User
- Contributor

The contributors table is a join table between the other two. I'm writing a query to load a contributor but I want it limited to the current business and make sure that the current user has access to that business to be able to view the contributor.

My avram code looks like:

```crystal
business_query = BusinessQuery.new
  .id(business_id)
  .where_users(UserQuery.new.id(current_user.id))
ContributorQuery.new
  .where_businesses(business_query)
  .find(contributor_id)
```

When I run this code, it fails with the error `table name "contributors" specified more than once`

You will notice that I have a has many through set up on Business to be able to call `where_users` but I've tried it with `where_contributors` and it doesn't change anything.

## Generated SQL

```sql
SELECT contributors.id, contributors.created_at, contributors.updated_at, contributors.user_id, contributors.business_id
FROM contributors
INNER JOIN businesses ON contributors.business_id = businesses.id
INNER JOIN contributors ON businesses.id = contributors.business_id
INNER JOIN users ON contributors.user_id = users.id
WHERE businesses.id = $1
  AND users.id = $2
  AND contributors.id = $3
ORDER BY contributors.id ASC
LIMIT 1
```

## Potential Solution

When a query does `.where_x`, maybe everything in that should be a sub-select?

Doing something like:

```crystal
TodoQuery.new.where_users(UserQuery.new.id(1))
```

Right now would make:

```sql
SELECT todos.*
FROM todos
INNER JOIN users
ON todos.user_id = users.id
WHERE users.id = 1
```

But with subselects would make:

```sql
SELECT todos.*
FROM todos
INNER JOIN (
  SELECT users.*
  FROM users
  WHERE users.id = 1
) as users_temp
ON todos.user_id = users_temp.id
```

(I am not ABSOLUTELY sure those produce the same results as I'm making them up)

matthewmcgarvey:
Fixes #370 

## Summary

Querying with joins can sometimes produce invalid sql. The two cases that this PR handles are:
1. When they potential join is to the table being queried
2. When the potential join has already been joined

## Description

These situations are rather easy to get to when users attempt to make composable methods that limit the scope of queries. Say an app has the concept of a `Business` and a `User` and if the user belongs to a business they are considered an `Employee` of that business. If the app allowed users to access the api, it might make sense to limit their API access to businesses they are an employee of. Rather than having to remember how to scope their access correctly every time you have a new query, you could make helper methods like (granted this is convoluted, I'm sure there are more realistic ways to get in this situation):

```crystal
def user_scoped(query)
  scoped_user_query = UserQuery.new.id(current_user.id)
  query.where_users(scoped_user_query)
end

def business_scoped(query)
  scoped_business_query = user_scoped(BusinessQuery.new.id(current_business.id))
  query.where_businesses(scoped_business_query)
end
```

(Those methods would require that `Business` has a has many through association to `User` through `Employee`)
The problem arises when you want to provide an endpoint that returns all the employees for the user's current business. The action would do something like:
```crystal
employees = business_scoped(EmployeeQuery.new)
```

You end up with an error: `table name "employees" specified more than once (PQ::PQError)` because the BusinessQuery adds a join to limit by the user id and the main query is on the employees table.

## Note

I'm sure there are situations this doesn't handle and this might even cause issues so _PLEASE_ poke holes in this change.

### DIFF
--- a/spec/query_associations_spec.cr
+++ b/spec/query_associations_spec.cr
@@ -1,5 +1,11 @@
 require "./spec_helper"
 
+class LineItemProductQuery < LineItemProduct::BaseQuery
+end
+
+class ProductQuery < Product::BaseQuery
+end
+
 # Ensure it works with inherited query classes
 class CommentQuery < Comment::BaseQuery
   def body_eq(value)
@@ -139,5 +145,20 @@ describe "Query associations" do
 
     staff = NamedSpaced::Staff::BaseQuery.new
     staff.to_sql[0].should contain "named_spaced_staffs"
+  end
+
+  it "can query with potentially unnecessary joins" do
+    item = LineItemBox.create
+    product = ProductBox.create
+    line_item_product = LineItemProductBox.create &.line_item_id(item.id).product_id(product.id)
+
+    line_item_query = LineItemQuery.new
+      .id(item.id)
+      .where_products(ProductQuery.new.id(product.id))
+    result = LineItemProductQuery.new
+      .where_line_items(line_item_query)
+      .find(line_item_product.id)
+
+    result.should eq(line_item_product)
   end
 end
--- a/spec/query_associations_spec.cr
+++ b/spec/query_associations_spec.cr
@@ -161,4 +161,19 @@ describe "Query associations" do
 
     result.should eq(line_item_product)
   end
+
+  it "can query with potentially unnecessary joins again" do
+    item = LineItemBox.create
+    product = ProductBox.create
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Looks good! Just needs a rebase or pull from master. After everything passes feel free to merge 👍

PR #452, opened 2020-09-05
Allow using fill_existing_with and nilable columns

### LINKED ISSUE #421: fill_existing_with Overrides Nullable Column Unexpectedly
I have a scenario where I have a nillable column, but would like to populate it for the existing records. I'm adding a `confirmed_at` timestamp to users, and don't want to force users who have already signed up and begun using the app to have to confirm their email.

I wrote this migration:
```crystal
class AddConfirmationToUsers::V20200711174331 < Avram::Migrator::Migration::V1
  def migrate
    alter table_for(User) do
      add confirmed_at : Time?, fill_existing_with: Time.utc
      add confirmation_token : String, fill_existing_with: "PRE_TOKEN_AUTH"
    end
  end
end
```

This created the column as `not null` rather than allowing nulls, as intended by `Time?`. It did populate the values as expected

I think that we should either:

1. Warn the user with something like "Using fill_with_existing will override your optional `confirmed_at` column. Please populate nullable columns manually", and update the website documentation [here](https://luckyframework.org/guides/database/migrations#using-fill_existing_with-and-default-values) to call out that it happens.
1. Allow `fill_existing_with` to work with nullable columns, which I think makes sense from a database perspective.

matthewmcgarvey:
Fixes #421 

This allows users to add optional columns and choosing to fill any existing columns with a value. Before, if you provided `fill_existing_with` it forced the column to be non null.


paulcsmith (approved):
Love it! Thank you :D

Issue #6, opened 2018-06-13
Tables with no primary key


jwoertink:
With some schemas there might be join tables that don't need a primary key.

```
user_roles
  user_id : integer
  role_id : integer
```

With a setup like this, LuckyRecord currently doesn't like this table. It would be cool to be able to specify a model like this for doing joins through this table when you'd never need to query on this table directly.

lesderid:
Is there a solution for this yet? This is a pretty major issue for anything more than the simplest database schemas.

paulcsmith:
Right now there is not. Can you explain what your use case is? This should not be an issue for any new app as a primary key is added for all tables. Are you using an existing Schema with Avram that does not have primary keys? I’ve personally never worked on a database where tables don’t have a primary key so I’m trying to understand the use case so I can prioritize this

lesderid:
Having no primary key at all is usually not very useful, but it can make sense to have a composite primary key (i.e. a primary key over more than one column).

An example would be the 'tag map' table of a toxi-style tagging system (see http://howto.philippkeller.com/2005/06/19/Tagsystems-performance-tests/). In many cases having a composite key over the two (foreign key) IDs is sufficient, and a separate ID serving as a primary key doesn't serve much of a purpose.

paulcsmith:
@lesderid that makes sense. Thanks for sharing. In the meantime can you still use Avram and add the primary key, or is this currently a show stopper for you?

lesderid:
Adding the primary key is fine for now. 

(FWIW, Laravel devs ignored this issue, which annoyed a lot of people: https://github.com/laravel/framework/issues/5355#issuecomment-72261848.)

jwoertink:
closing in favor of 96

jwoertink:
This is actually not done. You still always have to call `primary_key`. https://github.com/luckyframework/avram/pull/121 added in the ability to use `skip_default_columns` which remove the `primary_key` and `timestamps`; however, if you don't manually call `primary_key` you'll get this error:

```
in lib/avram/src/avram/model.cr:100: undefined constant PRIMARY_KEY_TYPE

        primary_key_type: {{ PRIMARY_KEY_TYPE }},
                             ^~~~~~~~~~~~~~~~
```

Because the `primary_key` macro sets up those constants.

jwoertink:
[...]

PR #454, opened 2020-09-07
Provide better error when primary key is missing

matthewmcgarvey:
Pointed out in [this issue comment](https://github.com/luckyframework/avram/issues/6#issuecomment-515215447)

If a model uses `skip_default_columns`, they are required to call `primary_key` in the `table` block. If they forget, the error is cryptic. This updates the process to actually check for a primary key and provide a more clear error.

## Before

```
Showing last frame. Use --error-trace for full trace.

In src/avram/model.cr:170:30

 170 | primary_key_type: {{ PRIMARY_KEY_TYPE }},
                            ^
Error: undefined constant PRIMARY_KEY_TYPE
```

## After

```
Showing last frame. Use --error-trace for full trace.

There was a problem expanding macro 'table'

Code in spec/support/line_item_product.cr:4:3

 4 | table :line_items_products do
     ^
Called macro defined in src/avram/model.cr:112:3

 112 | macro table(table_name = nil)

Which expanded to:

 >  8 |  end
 >  9 | 
 > 10 |     validate_primary_key
            ^
Error: No primary key was specified.

Example:

  table do
    primary_key id : Int64
    ...
  end
```

### DIFF
--- a/src/avram/model.cr
+++ b/src/avram/model.cr
@@ -118,6 +118,8 @@ abstract class Avram::Model
 
     {{ yield }}
 
+    validate_primary_key
+
     setup({{table_name}})
     {% MACRO_CHECKS[:setup_complete] = true %}
   end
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Nice! Yeah, this is a much nicer error.

paulcsmith (approved):
Love it!

Issue #453, opened 2020-09-07
Add support for database views


matthewmcgarvey:
Database views are a great way to let Postgres do some of the data manipulation work before pulling the data into the app. There are several things in the way of allowing us to use Avram to interact with views:

- [ ] views will not have primary keys (if they do, they are from other tables)
- [ ] views will not have timestamp columns (if they do, they are from other tables)
- [ ] views are read-only but we always generate Operation logic
- [ ] SchemaEnforcer only verifies database tables

Connected:
- [SchemaEnforcer doesn't know about views](https://github.com/luckyframework/avram/issues/420)
- [Tables with no primary key](https://github.com/luckyframework/avram/issues/6)

jwoertink:
I'm actually using [Materialzed views](https://medium.com/@stefan.poeltl/views-v-s-materialized-views-v-s-rollup-tables-with-postgresql-2b3824b45330) already with Avram. Even though views don't need a primary key, if you just set one anyway, then the rest sets up like a normal model including the query class.

```crystal
  class Release < BaseModel
    macro default_columns
      primary_key id : Int32
    end

    table :releases do
      column released : Bool
      column released_at : Time?
    end
end

ReleaseQuery.new.released(false).select_count
```

The tricky part with views is that generating the migration to create them.

```sql
DROP MATERIALIZED VIEW IF EXISTS releases;
CREATE MATERIALIZED VIEW releases AS
   SELECT a bunch of stuff
```

And then you create a function to refresh the data (which we can do in crystal now)

```sql
CREATE FUNCTION public.refreshallmaterializedviewsconcurrently(schema_arg text DEFAULT 'public'::text) RETURNS integer
    LANGUAGE plpgsql SECURITY DEFINER
    AS $$
      DECLARE
        r RECORD;
      BEGIN
        RAISE NOTICE 'Refreshing materialized view in schema %', schema_arg;
        FOR r IN SELECT matviewname FROM pg_matviews WHERE schemaname = schema_arg
        LOOP
          RAISE NOTICE 'Refreshing %.%', schema_arg, r.matviewname;
          EXECUTE 'REFRESH MATERIALIZED VIEW ' || schema_arg || '.' || r.matviewname || ' with data';
        END LOOP;
        RETURN 1;
      END
    $$;
```

Then somewhere we call something like `select refreshallmaterializedviewsconcurrently();` I think, but I forget how we're currently running that.... Anyway, hopefully this gives some more clarity and context for anyone coming across this

[...]

PR #455, opened 2020-09-07
Extract PrimaryKeyQueryable from Queryable

matthewmcgarvey:
Related to #453 

If we want to support database views in the future, one step in getting there is removing the requirement of the database table having a primary key.

When the time comes to allow not setting a primary key, the `BaseQueryTemplate` not include `PrimaryKeyQueryable` which will remove any ability to query by the primary key for that model.

### DIFF
--- a/src/avram/primary_key_queryable.cr
+++ b/src/avram/primary_key_queryable.cr
@@ -1,13 +1,11 @@
 require "./errors"
-require "./queryable"
 
 module Avram::PrimaryKeyQueryable(T)
   abstract def id
   abstract def id(id_val)
+  abstract def query
 
   macro included
-    include Avram::Queryable(T)
-
     def self.find(id)
       new.find(id)
     end
--- a/src/avram/primary_key_queryable.cr
+++ b/src/avram/primary_key_queryable.cr
@@ -1,10 +1,6 @@
 require "./errors"
 
 module Avram::PrimaryKeyQueryable(T)
-  abstract def id
-  abstract def id(id_val)
-  abstract def query
-
   macro included
     def self.find(id)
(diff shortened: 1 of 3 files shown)


paulcsmith (approved):
Smart! I like this approach

paulcsmith:
Looks like there are some conflicts. Ping me when you're ready and I'll take another look + merge

matthewmcgarvey:
@paulcsmith This is ready to be reviewed again. I did tweak it so that `PrimaryKeyQueryable` does not include `Queryable`. Seemed like the wrong choice to do that inheritance type structure.

src/avram/primary_key_queryable.cr
```diff
@@ -0,0 +1,25 @@
+require "./errors"
+
+module Avram::PrimaryKeyQueryable(T)
+  abstract def id
+  abstract def id(id_val)
+  abstract def query
+
```
paulcsmith, line 7: I think these should be removed since the module itself implements the methods. 
```suggestion
```

paulcsmith (commented):
I like this new approach of having them implement just the methods they own and including both modules 👍 

Just one thought about removing the abstract def

src/avram/primary_key_queryable.cr
matthewmcgarvey: @paulcsmith done

jwoertink:
@matthewmcgarvey is there anything else you need to do on this? If not, I think it's ready to merge in.

jwoertink (approved):

matthewmcgarvey:
@jwoertink it's ready, thanks

Issue #368, opened 2020-05-15
Change how `validate_uniqueness_of` query works


paulcsmith:
> See https://gitter.im/luckyframework/Lobby?at=5ebdf71bf3ce603074bc908a for some background on the issue

So it uses a query and doesn't end with a criteria. When I originally wrote this I didn't know how else to do it. But I figured out a better way to do it in AvramSlugify.

## Now

```crystal
validate_uniqueness_of name, query: UserQuery.new.project_id(1).name # Need to end with `name` criteria
```

## After

```crystal
validate_uniqueness_of name, query: UserQuery.new.project_id(1) # Just use a query
```

Do this by using the passed in attribute's name and a raw where. Still (mostly) type-safe because the attribute is typed and auto-generated by the model

matthewmcgarvey:
Switching to this style would remove the ability to change how the value is searched by. If you wanted to check for the uniqueness of a string case insensitively you can do `validate_uniqueness_of name, query: UserQuery.new.name.lower` currently. We could add additional parameters for certain column types to compare in different ways though String might be the only one that needs it.

paulcsmith:
That’s a good point. Maybe we allow either one? So you can return a criteria if you want to customize but it is not required. What do you think?

matthewmcgarvey:
This is the link to the avram_slugify that should be referenced https://github.com/luckyframework/avram_slugify/blob/be75b34e358e6dc2f6d9ad8d9ed079754b945ad2/src/avram_slugify.cr#L29

PR #458, opened 2020-09-09
Remove duplicate validate_uniqueness_of from Avram::Validations

matthewmcgarvey:
The code that was in `Avram::Validations` would have failed if it wasn't always being overwritten by `Avram::DatabaseValidations` by nature of the way `macro included` works (which means `Avram::Validations` could not be included in a different module if that module doesn't overwrite the removed methods). Since it is only for code that saves to the database and there's explicit documentation that the code will fail if the methods are used outside of `SaveOperation`s, there's no need for it to be in `Avram::Validations`. .

I also changed `Avram::DatabaseValidations` to take in a generic `T` so that it can stop depending on the "magic" of `macro included`.

This change will allow completing #368 with only changes in `Avram::DatabaseValidations` instead of keeping both in sync.


paulcsmith (approved):
Looks great! Good idea on extracting it and using `T` 👍 

And thanks for the thorough PR description

PR #460, opened 2020-09-12
Represent wheres and raw wheres in the same way

### LINKED ISSUE #459: SQL where clauses are not kept in the order they are called if using raw wheres
Failing spec:

```crystal
it "keeps the sql in order that it was called when raw where methods used with non-raw" do
  query = UserQuery.new.where("first_name = ?", "Paul").name("Smith").query

  query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE first_name = 'Paul' AND users.name = $1"
  query.args.should eq ["Smith"]
end
```

Failure message:

```
Failure/Error: query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE first_name = 'Paul' AND users.name = $1"

       Expected: "SELECT users.id, users.created_at, users.updated_at, users.name, users.age, users.year_born, users.nickname, users.joined_at, users.total_score, users.average_score, users.available_for_hire FROM users WHERE first_name = 'Paul' AND users.name = $1"
            got: "SELECT users.id, users.created_at, users.updated_at, users.name, users.age, users.year_born, users.nickname, users.joined_at, users.total_score, users.average_score, users.available_for_hire FROM users WHERE users.name = $1 AND first_name = 'Paul'"
```

Raw wheres are currently stored separately from the regular where clauses ([source](https://github.com/luckyframework/avram/blob/b002ead126de2e4e47c1ca9c2192ad86db91f844/src/avram/query_builder.cr#L10)). The raw wheres are appended to the end where statement when make the full sql string. This is fine when our api only supports `AND` statements but is a problem when `OR` is added to the API.

### Potential Solutions

- Do not allow further "where" methods if a raw where method is used
- Unify the where and raw where list so that the order is preserved
- ???

matthewmcgarvey:
Fixes #459

`Avram::Where::Raw` and `Avram::Where::SqlClause` did not have overlapping ancestors so they were stored in separate lists in `Avram::QueryBuilder` and had different APIs. When the sql string was put together, raw wheres were appended to the where clause. When "OR" support is added, this will no longer be acceptable.

It simplifies the API to have them be used in the same way. This adds a common ancestor of `Avram::Where::Condition` that has the one method of `#prepare`. The only problem of doing this is that `SqlClause`s where being passed a placeholder string that had an incrementing number. Rather than keeping with the push style and checking which types were save to give the placeholder to, I switched the where conditions to take in a supplier of those placeholders so that it could get one if it wanted or just ignore it and not cause the number to increment.

PS: Naming is hard, any ideas on how to name the different abstract classes better?

### DIFF
--- a/src/avram/where.cr
+++ b/src/avram/where.cr
@@ -1,9 +1,8 @@
 module Avram::Where
   abstract class SqlClause
-    getter :column
-    getter :value
+    getter column : Symbol | String
 
-    def initialize(@column : Symbol | String, @value : String | Array(String) | Array(Int32))
+    def initialize(@column)
     end
 
     abstract def operator : String
@@ -18,7 +17,7 @@ module Avram::Where
     end
 
     def ==(other : SqlClause)
-      (prepare(->{"unusued"}) + value.to_s) == (other.prepare(->{"unused"}) + other.value.to_s)
+      prepare(->{"unusued"}) == other.prepare(->{"unused"})
     end
 
     def ==(other)
@@ -26,14 +25,24 @@ module Avram::Where
     end
   end
 
-  abstract class NullSqlClause < SqlClause
-    @value = "NULL"
+  abstract class ValueHoldingSqlClause < SqlClause
+    getter value : String | Array(String) | Array(Int32)
 
-    def initialize(@column : Symbol | String)
+    def initialize(@column, @value)
     end
 
+    def ==(other : ValueHoldingSqlClause)
+      (prepare(->{"unusued"}) + value.to_s) == (other.prepare(->{"unused"}) + other.value.to_s)
+    end
+
+    def ==(other : SqlClause)
+      false
+    end
+  end
+
+  abstract class NullSqlClause < SqlClause
     def prepare(_placeholder_supplier : Proc(String))
-      "#{column} #{operator} #{@value}"
+      "#{column} #{operator} NULL"
     end
   end
 
@@ -43,7 +52,7 @@ module Avram::Where
     end
 
     def negated : NotNull
-      NotNull.new(@column)
+      NotNull.new(column)
     end
   end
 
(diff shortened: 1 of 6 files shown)


jwoertink:
Ah, I see what you're doing here. That's nice. I'll come back for a review again later, but initial glance over I like the direction. I'll also try and think of a good name for that abstract class. Thanks for doing this!

src/avram/where.cr
```diff
@@ -1,31 +1,52 @@
...
+      (prepare(->{ "unused" }) + value.to_s) == (other.prepare(->{ "unused" }) + other.value.to_s)
+    end
+
+    def ==(other : Condition)
+      false
+    end
+  end
+
+  abstract class NullSqlClause < SqlClause
+    def prepare(_placeholder_supplier : Proc(String))
```
jwoertink, line 48:
```suggestion
    def prepare(_placeholder_supplier : Proc(String)) : String
```

This should probably have the return type to keep it consistent with the abstract.

src/avram/where.cr
```diff
@@ -45,139 +66,139 @@ module Avram::Where
...
     end
 
     def negated : In
-      In.new(@column, @value)
+      In.new(column, value)
     end
 
-    def prepare(prepared_statement_placeholder : String)
-      "#{column} #{operator} (#{prepared_statement_placeholder})"
+    def prepare(placeholder_supplier : Proc(String))
```
jwoertink, line 196:
```suggestion
    def prepare(placeholder_supplier : Proc(String)) : String
```

src/avram/where.cr
```diff
@@ -1,31 +1,52 @@
...
+    abstract def operator : String
+    abstract def negated : SqlClause
+
+    def prepare(placeholder_supplier : Proc(String)) : String
+      "#{column} #{operator} #{placeholder_supplier.call}"
+    end
+  end
+
+  abstract class ValueHoldingSqlClause < SqlClause
+    getter value : String | Array(String) | Array(Int32)
```
jwoertink, line 33: I know this type comes from what was defined in Condition before, but any idea why these types specifically? I ask because I'm wondering if with the latest pg and db, will this blow up when someone uses UUID? 🤔 Also, Int64 I guess... I'm assuming this is the value of a DB column which could be many different types... Or maybe I'm wrong and this value is something else?

jwoertink (commented):
Ok, so for the most part I like what this is doing, and I think I'll need it before being able to merge in #442. I think my only hesitation here is if will cause any type-safe issues by these being merged in.... For example, when you query, you never actually call `Avram::Where` directly... you'll do `UserQuery.new.username()` or you'll do `UserQuery.new.where("")`.... I may just be overthinking it now that I write it out 😝  But what are your thoughts?

src/avram/where.cr
matthewmcgarvey: I'm not able to hunt down why we convert everything to these types

ultimately, we call out to here https://github.com/crystal-lang/crystal-db/blob/bd456028644853f852f5adfe053142e6ad885e52/src/db/statement.cr#L69-L71
which does not have type signatures so it's not very helpful.  Our process ultimately delegates to the `#to_db` method for each type which in most things converts to a string (Ints, UUIDs, etc).

jwoertink: It'll probably be ok, but lets keep this in mind if we notice something weird popup. I feel like this should have more types on it, but if it works as is then we can just roll with it.

matthewmcgarvey:
@jwoertink I don't think any type safety was lost/gained. Most queries will never interact with the `#where` that takes in these base classes, but when they do they won't have to remember whether they should use `#where` or `#raw_where`.

jwoertink (approved):
👍

PR #461, opened 2020-09-14
Fixing issue with required bool fields

### LINKED ISSUE #457: Updating a record fails if a column has original value of `false`
The following fails, with an error:

```crystal
SaveUserOptions.update(
  user.options!, # user.options! has `password_notify` = false, `login_notify` = true
  params(login_notify: false) # `password_notify` is not updated here
) do |operation, updated_user_options|
  # ...
end
```

The error is `{:password_notify => ["is required"]}`

Seems the operation fails because `password_notify` has original value of `false`. If you explicitly pass `password_notify: false` to the operation, it works OK. Otherwise, if `password_notify` has original value of `true`, it works OK.

This seems to be a regression in #424. My guess is this line:

https://github.com/luckyframework/avram/blob/fedd1bef805ca3c18d1b8042bec68f800be0dcf3/src/avram/save_operation.cr#L126

jwoertink:
Fixes #457

Thanks to @akadusei for catching this and the suggested fix.

If you had a Boolean field and the default value was already set to `false`, but your operation had a `validate_required` on that field, the update would fail. This wasn't the case if you removed the validation. This PR fixes that.


spec/support/boxes/model_with_default_values_box.cr
```diff
@@ -0,0 +1,4 @@
+class ModelWithDefaultValuesBox < BaseBox
```
paulcsmith, line 1: I'm having trouble understanding this part. Maybe you could use the operation directly without any args instead of having a box? Since boxes usually set fake data it seems a bit odd to have one that doesn't do anything

spec/save_operation_spec.cr
```diff
@@ -428,6 +436,17 @@ describe "Avram::SaveOperation" do
...
+
+      it "updates with a record that has defaults" do
+        model = ModelWithDefaultValuesBox.create
+        params = Avram::Params.new({"greeting" => "Hi"})
+        OverrideDefaults.update(model, params) do |_operation, record|
+          record.should_not eq nil
+          r = record.not_nil!
+          r.greeting.should eq "Hi"
+          r.admin.should eq false
+        end
```
paulcsmith, line 448:
```suggestion
        record = OverrideDefaults.update!(model, greeting: "Hi")
        record.greeting.should eq "Hi"
        record.admin.should eq false
```

spec/save_operation_spec.cr
```diff
@@ -553,6 +572,20 @@ describe "Avram::SaveOperation" do
         end
       end
     end
+
+    context "when the default is false, but the field is required" do
+      it "passes required validation as 'false' is a valid Boolean value" do
```
paulcsmith, line 577: What do you think about this?

```suggestion
    context "when the default is false and the field is required" do
      it "is valid since 'false' is a valid Boolean value" do
```

spec/save_operation_spec.cr
```diff
@@ -553,6 +572,20 @@ describe "Avram::SaveOperation" do
...
+        user = UserBox.create &.nickname("oopsie").available_for_hire(false)
+        params = Avram::Params.new({"nickname" => "falsey mcfalserson"})
+        SaveUserWithFalseValueValidations.update(user, params) do |operation, record|
+          record.should_not eq nil
+          r = record.not_nil!
+          operation.errors.should eq({} of Symbol => Array(String))
```
paulcsmith, line 583:
```suggestion
          operation.valid?.should be_true
```

paulcsmith (approved):
A few small comments. LGTM

PR #465, opened 2020-09-16
Use parent's primary key name for the belongs to join foreign key

### LINKED ISSUE #464: join generated for belongs_to association uses its own primary key name rather than association's
The join sql generated for `belongs_to` uses the primary key name for the current model, not the model it is joining to.

This causes a bug when one side has a custom primary key name and the other doesn't.

## Example

```crystal
class Post < Avram::Model
  skip_default_columns

  table do
    primary_key custom_id : UUID
    has_many comments : Comment
  end
end

class Comment < Avram::Model
  table do
    belongs_to post : Post
  end
end

# given that post exists and has 
Comment::BaseQuery.new.join_posts.first
```

That should return a comment with a post but instead you get an error:
```
column posts.id does not exist (PQ::PQError)
...stacktrace
```

matthewmcgarvey:
Fixes #464 

It was using the model's own primary key name for the foreign key of the belongs to association which causes problems when one side has a different primary key name than the other.

I also discovered that the `PRIMARY_KEY_NAME` constant was a [MacroId](https://crystal-lang.org/api/0.35.1/Crystal/Macros/MacroId.html) rather than a usable representation of the primary key name which caused a compilation error when attempting to use it. I made it a symbol because that seems to be the standard set in the library.


db/migrations/20180115194734_create_taggings.cr
```diff
@@ -7,7 +7,7 @@ class CreateTaggings::V20180115194734 < Avram::Migrator::Migration::V1
     end
 
     create :taggings do
-      primary_key custom_id : Int64
+      primary_key id : Int64
```
jwoertink, line 10: Why did these need to change? Does it fail if you keep them as `custom_id`? I think the idea was that we wanted to test using custom ids here. 🤔

matthewmcgarvey: This was to cause failures that would be fixed when the change was made (which is why I didn't add any tests). There are other associations that test the custom primary key name. I chose to change this one because it had an association with a custom key name and was not widely used in the specs so I wouldn't have to update much

jwoertink: Ah, ok. That's cool then 👍

jwoertink (approved):

Issue #462, opened 2020-09-15
Refactor SchemaEnforcer to avoid re-querying the database for each model


matthewmcgarvey:
Right now, `Avram::Model` includes `Avram::SchemaEnforcer` and then each model makes sql requests to check that its table name and listed columns exist. So, having 10 models ends up making 20 queries. Right now, this check only runs once when starting an app in development, but if the check [ever gets moved to a middleware](https://github.com/luckyframework/lucky/issues/1142) it would make the sql queries on every request. And, if we want wanted to add more types of database checks (warn if belongs_to relationship doesn't have a foreign key relationship set up for example) we would also have to add yet another sql query.

## Proposal

I propose that we update models to include a `Avram::SchemaEnforceable` module and that runs all the checks. We also add an `Avram::DatabaseInfo` class that is accessible from the class-level of every `Avram::Database`. That class would have many `Avram::DatabaseTable`'s and each of those would have many `Avram::DatabaseColumn`'s. Rather than the checks getting passed a database class and using it to make a query, they will be passed the `Avram::DatabaseInfo` and have rich classes to compare to. The `Avram::DatabaseInfo` class should only run one query to get all the database tables and could also join to get the column info first or subsequent queries could be made (lazily or not). This `Avram::DatabaseInfo` should be a singleton so that it is only "new-ed" up once at application start.

I expect this to make adding database view support to the schema enforcement process easier than it would be right now.

### Pseudo-Code

```crystal
class Avram::Model
  include Avram::SchemaEnforceable
  ...
end

module Avram::SchemaEnforceable
  DATABASE_VALIDATIONS = [
    EnsureExistingTable,
    EnsureMatchingColumns
  ]
  DATABASE_INFO = Avram::DatabaseInfo.new(
  def self.run_database_validations!
    DATABASE_VALIDATIONS.each do |validation|
      validation.new(self, database.database_info).validate!
    end
  end
end

class Avram::Database
  memoize def self.database_info
    ..
  end
end

class Avram::DatabaseInfo
  has many DatabaseTable
end

...etc
```

jwoertink:
The sounds like a pretty good direction. Since I've never actually been able to use the SchemaEnforcer, I'm not sure how it works now. I just have a few points I think this would need to know about, so maybe you can add these in to your thoughts on the structure:

* Need to know if the model is using a view/materialized view
* Need to know if any columns were defined on the table that aren't defined in the model
* Need to know if any of the columns in the model have a default value, and the table has those same default values
[...]

PR #467, opened 2020-09-19
Add DatabaseInfo for SchemaEnforcer use

matthewmcgarvey:
Connected #462 

This is part 1 of refactoring `SchemaEnforcer`. This adds `Avram::Database::DatabaseInfo` to every `Avram::Database` which pulls all of the tables and columns from that database for application use and memoizes it. This means that we only run one query for all the database information instead of a query each time we need the information.

It also opens up the query to pull more than just the most basic information. It now has all view table information as well as regular tables so that views can be handled. Future schema enforcers could be added to check data types of columns, as well.

table and column table info:
- https://www.postgresql.org/docs/11/infoschema-tables.html
- https://www.postgresql.org/docs/11/infoschema-columns.html

### DIFF
--- a/src/avram/database.cr
+++ b/src/avram/database.cr
@@ -177,67 +177,23 @@ abstract class Avram::Database
     transactions.delete(Fiber.current.object_id)
   end
 
-  def table_names
-    tables_with_schema(excluding: "migrations")
-  end
-
-  def tables_with_schema(excluding : String)
-    select_rows <<-SQL
-    SELECT table_name
-    FROM information_schema.tables
-    WHERE table_schema='public'
-    AND table_type='BASE TABLE'
-    AND table_name != '#{excluding}';
-    SQL
-  end
-
-  def select_rows(statement)
-    rows = [] of String
-
-    run do |db|
-      db.query statement do |rs|
-        rs.each do
-          rows << rs.read(String)
-        end
-      end
-    end
-
-    rows
-  end
-
-  def table_columns(table_name)
-    statement = <<-SQL
-    SELECT column_name as name, is_nullable::boolean as nilable
-    FROM information_schema.columns
-    WHERE table_schema = 'public'
-      AND table_name = '#{table_name}'
-    SQL
-
-    run { |db| db.query_all statement, as: TableColumn }
-  end
-
-  class TableColumn
-    DB.mapping({
-      name:    String,
-      nilable: Bool,
-    })
-  end
-
   class DatabaseCleaner
-    private getter database
+    private getter database : Avram::Database
+    private getter database_info : Avram::Database::DatabaseInfo
 
-    def initialize(@database : Avram::Database)
+    def initialize(@database)
+      @database_info = database.class.database_info
     end
 
     def truncate
-      table_names = database.table_names
+      table_names = database_info.table_names
       return if table_names.empty?
       statement = ("TRUNCATE TABLE #{table_names.map { |name| name }.join(", ")} RESTART IDENTITY CASCADE;")
       database.exec statement
     end
 
     def delete
-      table_names = database.table_names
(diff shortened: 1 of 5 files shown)


src/avram/database.cr
```diff
@@ -38,6 +38,12 @@ abstract class Avram::Database
     new.delete
   end
 
+  @@database_info : DatabaseInfo?
+
+  def self.database_info : DatabaseInfo
+    @@database_info ||= DatabaseInfo.load(self)
```
wontruefree, line 44: is `self` in this context the concrete class of `Avram::Database`?

matthewmcgarvey: Yes. I had everything in `DatabaseInfo.load` in this method but moved it in effort to clean up

matthewmcgarvey:
@jwoertink Could you take a look at this PR?

src/avram/database/column_info.cr
```diff
@@ -0,0 +1,24 @@
+module Avram
+  struct Database::ColumnInfo
```
jwoertink, line 2: Nice use of the struct 👍

src/avram/database/database_info.cr
```diff
@@ -0,0 +1,45 @@
+module Avram
+  struct Database::DatabaseInfo
+    def self.load(database)
```
jwoertink, line 3:
```suggestion
    def self.load(database : Avram::Database)
```

The type signature will make it easier to know what's being passed in

src/avram/database/database_info.cr
```diff
@@ -0,0 +1,45 @@
...
+    def table?(name : String)
+      table_names.includes?(name)
+    end
+
+    def table_names
+      table_infos.map(&.table_name)
+    end
+
+    def table(name : String) : TableInfo?
+      table_infos.find { |table_info| table_info.table_name == name }
```
jwoertink, line 42:
```suggestion
      table_infos.find(&.table_name.==(name))
```

In `TableInfo` you wrote a column method that looked like this. Either way is fine, but they should look the same.

src/avram/database/database_info.cr
```diff
@@ -0,0 +1,45 @@
+module Avram
+  struct Database::DatabaseInfo
+    def self.load(database)
+      sql = <<-SQL
```
wontruefree, line 4: This is a super small thing.  Would this be better as a constant?

src/avram/database.cr
```diff
@@ -171,67 +177,23 @@ abstract class Avram::Database
...
-    tables_with_schema(excluding: "migrations")
-  end
-
-  def tables_with_schema(excluding : String)
-    select_rows <<-SQL
-    SELECT table_name
-    FROM information_schema.tables
-    WHERE table_schema='public'
-    AND table_type='BASE TABLE'
-    AND table_name != '#{excluding}';
```
jwoertink, line 184: Do we need to account for this somewhere to avoid having the migrations table included in any of this?

jwoertink (commented):
It looks pretty good, but I did have a few questions and suggestions.

src/avram/database.cr
matthewmcgarvey: So, I'm not entirely sure why this condition was there to begin with since all the checks are pulling out specific database tables. So I guess there's two differences between master and this right now.
1. I am no longer excluding the migrations database
2. I am pull tall tables (which includes views) and not just `BASE TABLE`'s

jwoertink: Cool with me. Maybe it was just to save on doing an extra query? But with this all being in one big query it probably doesn't matter.

jwoertink (approved):
Ok, it all looks good now 👍  Feel free to merge when ready.

PR #468, opened 2020-09-23
Use DB::Serializable instead of DB.mapping

### LINKED ISSUE #364: Investigate moving away from DB.mapping
Possibly in the next Crystal release, `JSON.mapping` and `YAML.mapping` will be deprecated https://github.com/crystal-lang/crystal/pull/9272 in favor of using Serializable

In the crystal-db shard, `DB.mapping` was made to mimic those APIs. To keep up with that, a `DB::Serializable` was added https://github.com/crystal-lang/crystal-db/blob/master/src/db/serializable.cr

I haven't seen an issue yet, but I'd imagine `DB.mapping` could go the same way. Right now it's used [here](https://github.com/luckyframework/avram/blob/master/src/avram/model.cr#L212). Maybe this could clean up our code?

matthewmcgarvey:
Fixes #364 

In order to investigate moving to `DB::Serializable`, it makes sense to have a working branch with it. This PR in no way means that we _should_ move to it. This was the minimum changes needed to get the specs to pass.

I do not see much benefit to moving to this new way other than `DB.mapping` might be deprecated like `JSON.mapping`. There is also the side effect of other instance variables being picked up which had to be tracked down. I am not sure if this would cause any bugs for end-users, either.


jwoertink:
Nice! On the one hand, I feel like this code reads better, but on the other, that `@[DB::Field(ignore: true)]` seems like it could get in the way real quick. 🤔  Thanks for giving this a shot! 

Hey @bcardiff, you wouldn't happen to know if `DB.mapping` will go the way of `JSON.mapping` and `YAML.mapping`, would you? 😬

matthewmcgarvey:
Going down this path, instead of having one method that cycles over the columns and creates the property, we could make the property in the `column`/`belongs_to`/etc. macros.  I was running into weird bugs with it though

jwoertink:
oh, that's a neat idea! That would make the code even cleaner if we could do that.

bcardiff:
I have no plans of removing `DB.mapping` for now. I would like to refactor it so the expanded code use the annotation, so by construction, they would not be able to offer different behavior. But db needs more stuff before it can goes into a clean-up cycle IMO.

jwoertink:
Cool. Thanks for the inside info! So no major rush on this then, but at least we now know *how* to do this if we decide to go this route 😄

PR #469, opened 2020-09-23
New Interface for Operations

### LINKED ISSUE #69: Method for working with VirtualForms more easily

jwoertink:
Fixes #69 

This PR is part of a series which breaks out #369 in to much smaller (and easier to digest) PRs. I'll try to best explain everything going on here.

In this PR `SaveOperation` no longer inherits from `Operation`, and `Operation` now has an interface which makes it feel a lot closer to how `SaveOperation` works. This means we can now do this:

```crystal
class SignInUser < Avram::Operation
  needs http_context : HTTP::Context
  attribute email : String
  attribute password : String
  attribute remember_me : Bool
  file_attribute :biometric_retinal_scan

  before_run do
    check_login_count(http_context)
    validate_required email, password
  end

  after_run do |user|
    save_successful_login(user)
  end

  def run
    # do some fancy login logic
    new_current_user
  end
end

post "/login" do
  SignInUser.run(params, http_context: context) do |operation, user|
    if user
      store_user_and_redirect(user)
    else
       you_get_the_idea
    end
  end
end
```

All of the `SaveOperation` work exactly the same as before.


src/avram/callbacks.cr
```diff
@@ -24,6 +24,18 @@ module Avram::Callbacks
     end
   end
 
+  macro before_run(method_name)
```
jwoertink, line 27: This does the exact same thing as `before_save`, but it feels weird calling `before_save` in an Operation. To clean this up, we could just have `before_run` call `before_save` internally... but then if we ever needed `before_save` to act different, we'd end up having to break them out again.

src/avram/callbacks.cr
```diff
@@ -80,6 +104,55 @@ module Avram::Callbacks
...
+  #   after_run do |value|
+  #     value == "some report"
+  #   end
+  #
+  #   def run
+  #     "some report"
+  #   end
+  # end
+  # ```
+  macro after_run(&block)
```
jwoertink, line 134: The `after_save` equivalent doesn't exist yet, but will in a following PR to keep this as small as I can.

src/avram/needy_initializer.cr
```diff
@@ -0,0 +1,102 @@
+module Avram::NeedyInitializer
```
jwoertink, line 1: This module covers everything the `needy_initializer_with_save_methods` does, but without the save methods since those are related to SaveOperation specifically. I made a separate one because I had a lot of trouble with the compiler trying to tell "if it's SaveOp, then call these macros, else skip them". It was easier to just break them apart.

src/avram/operation.cr
```diff
@@ -1,23 +1,82 @@
...
+require "./needy_initializer"
 
-class Avram::VirtualForm
-  macro inherited
-    {% raise "Avram::VirtualForm has been renamed to Avram::Operation. Please inherit from Avram::Operation." %}
-  end
-end
-
-class Avram::Operation
+abstract class Avram::Operation
```
jwoertink, line 8: If we wanted an extra level of type-safety, we could make this a Generic, and ensure `run` returns `T`. 🤔  Some people may just want it to run a task and be void at the end though..

src/avram/operation.cr
```diff
@@ -1,23 +1,82 @@
...
+  include Avram::NeedyInitializer
   include Avram::DefineAttribute
   include Avram::Validations
   include Avram::SaveOperationErrors
```
jwoertink, line 12: This isn't really `SaveOperation` specific. I could rename it, or that could come in a future PR I'm planning with some cleanup stuff.

src/avram/callbacks.cr
```diff
@@ -80,6 +104,55 @@ module Avram::Callbacks
...
+    {%
+      if block.args.size != 1
+        raise <<-ERR
+        The 'after_run' callback requires only 1 block arg to be passed.
```
matthewmcgarvey, line 138: "only" sounds like it assumes that the error happened because the block given had multiple parameters but they could also have given a block with no params and this error would be worded awkwardly.

src/avram/callbacks.cr
```diff
@@ -80,6 +104,55 @@ module Avram::Callbacks
...
+    %}
+    def after_run(object)
+      {% if @type.methods.map(&.name).includes?(:after_run.id) %}
+        previous_def
+      {% else %}
+        super
+      {% end %}
+
+      {{ block.args.first }} = object
+      {{ block.body }}
```
matthewmcgarvey, line 152: in the before run I think the code actually yields to the given block but this inlines it. Should it be consistent between the two?

src/avram/needy_initializer.cr
```diff
@@ -0,0 +1,102 @@
...
+
+    macro inherited
+      inherit_needs
+    end
+  end
+
+  macro needs(type_declaration)
+    {% OPERATION_NEEDS << type_declaration %}
+    @{{ type_declaration.var }} : {{ type_declaration.type }}
+    property {{ type_declaration.var }}
```
matthewmcgarvey, line 13:
```suggestion
    property {{ type_declaration.var }} : {{ type_declaration.type }}
```
Works right?

src/avram/needy_initializer.cr
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  macro inherit_needs
+    \{% if !@type.constant(:OPERATION_NEEDS) %}
+      OPERATION_NEEDS = [] of Nil
+    \{% end %}
```
matthewmcgarvey, line 19: This doesn't seem necessary...

matthewmcgarvey (commented):
I was nervous that you were going to try to merge the monster PR 🤣

src/avram/callbacks.cr
jwoertink: Maybe `The 'after_run' callback requires exactly 1 block arg to be passed.`?

src/avram/callbacks.cr
jwoertink: No, because this is a macro yield and the other is an method yield. These actually have to work different.

src/avram/needy_initializer.cr
jwoertink: 🤷  lol. I know it's bad, but I literally copied and just ripped out code I assumed wasn't needed. I guess I should actually read what it is doing 😂

src/avram/needy_initializer.cr
jwoertink: Pretty sure!

jwoertink:
@confact and @stephendolan since both of you have Lucky apps out in the wild, would you mind taking a glance over this to see if something is missing? When you get a free moment, of course.

I'm mainly looking for examples of regular `Avram::Operation` you might have in your apps where the code is a bit messy and see if this would clean that up and allow you to do what you need still. Or if you see something that makes you say "if this got merged, my app would break with no way to fix it"... Thanks! 😄

src/avram/callbacks.cr
```diff
@@ -24,7 +24,28 @@ module Avram::Callbacks
...
+    def before_run
+      {% if @type.methods.map(&.name).includes?(:before_run.id) %}
+        previous_def
+      {% else %}
+        super
+      {% end %}
+
+      {{ method_name.id }}
+    end
+  end
```
matthewmcgarvey, line 46: This and the identical `before_save` macro can delegate to the other macro that takes a block so that the code doesn't have to be duplicated
```suggestion
  macro before_run(method_name)
    before_run { {{ method_name.id }} }
  end
```

src/avram/callbacks.cr
```diff
@@ -80,6 +120,65 @@ module Avram::Callbacks
...
+    def after_run(object)
+      {% if @type.methods.map(&.name).includes?(:after_run.id) %}
+        previous_def
+      {% else %}
+        super
+      {% end %}
+
+      {{ method_name.id }}(object)
+    end
+  end
```
matthewmcgarvey, line 143: This can delegate to the other macro like
```suggestion
  macro after_run(method_name)
    after_run { |object| {{ method_name.id }}(object) }
  end
```

src/avram/callbacks.cr
```diff
@@ -80,6 +120,65 @@ module Avram::Callbacks
...
+    {%
+      if block.args.size != 1
+        raise <<-ERR
+        The 'after_run' callback requires exactly 1 block arg to be passed.
+        Example:
+          after_run { |value| some_method(value) }
+        ERR
+      end
+    %}
+    def after_run(object)
```
matthewmcgarvey, line 170:
```suggestion
    def after_run(%object)
```

If users pass a block like`{ |object| do_something(object) }` it fails to compile (the variable usage will have to be updated as well)

src/avram/operation.cr
```diff
@@ -1,23 +1,94 @@
...
+    operation.before_run
+    value = operation.run
+    if operation.valid?
+      operation.after_run(value)
+    else
+      value = nil
+    end
+    raise Avram::FailedOperation.new("The operation failed to return a value") unless value
+    operation
+  end
```
matthewmcgarvey, line 82: We can delegate to `.run`
```suggestion
  def self.run!(params : Avram::Paramable, *args, **named_args)
    run(params, *args, **named_args) do |operation, value|
      raise Avram::FailedOperation.new("The operation failed to return a value") unless value
      operation
    end
  end
```

By the way, should this be returning the operation or the value?

matthewmcgarvey (changes requested):

src/avram/callbacks.cr
wontruefree: This is some tricky naming.  What about `before_execution`?

src/avram/operation.cr
jwoertink: My original intention was to return the operation itself in case you needed to use that object for whatever. You'd still have access to the value. But now that I think about it, doing `SaveOperation.create!` returns the model (the value), so I guess this may as well follow suit. If you need access to the operation, you just have to use the block form. I think that's fine, right?

src/avram/callbacks.cr
jwoertink: I'm not sure that I follow here. Also, doesn't `%object` just turn it in to a fresh var so you wouldn't have a `object` local variable available to you?

src/avram/callbacks.cr
jwoertink: The method that you call in these is called `run`. So like `LoginUser.run(params)`. This was actually sort of modeled after the mutations gem in Ruby 😅  (though, that doesn't have callbacks).

src/avram/callbacks.cr
jwoertink: Ohh.. ok.. I'm waking up 😅  I see what you mean now.

src/avram/needy_initializer.cr
jwoertink: I took a look at this again, and without it, things start to explode, so I'm leaving it in for now.

jwoertink:
Thanks for the review @matthewmcgarvey! Those changes definitely help to make this code easier to read. If you have any ideas for more specs I can add in, let me know!

matthewmcgarvey (approved):
Looks good 👍

jwoertink:
Sweet! I'm gonna merge and start prepping the next PR for this series.

src/avram/callbacks.cr
paulcsmith: I think I'd flip it. So `before_save` calls `before_run` under the hood. That way if `before_save` needs to change something we can modify just `before_save` and leave `before_run` alone

src/avram/operation.cr
paulcsmith: Future PR sounds good! `Avram::OperationsErrors` seems like a good name

paulcsmith (commented):
This is really awesome!! Thank you

PR #471, opened 2020-09-25
Add support for optional arrays

### LINKED ISSUE #470: Compilation errors when trying to make/add/use an optional array column
Right now, code is expecting columns to be either an array or an optional, but not both.

matthewmcgarvey:
Fixes #470 

There are a couple of bugs that are fixed by these changes:
- You could not add an optional array column in a migration
- Optional array fields on models were not converted from database queries correctly
- Adding an array column to a table was adding "NOT NULL" to the add column sql statement even if it was not intended

### DIFF
--- a/spec/array_column_spec.cr
+++ b/spec/array_column_spec.cr
@@ -1,5 +1,9 @@
 require "./spec_helper"
 
+private class SaveBucket < Bucket::SaveOperation
+  permit_columns numbers
+end
+
 describe "Array Columns" do
   it "fails when passing a single value to an array query" do
     BucketBox.new.numbers([1, 2, 3]).create
@@ -22,6 +26,10 @@ describe "Array Columns" do
   it "handles optional Array" do
     BucketBox.create &.numbers(nil)
     bucket = BucketQuery.new.last
+    bucket.numbers.should be_nil
+    bucket = SaveBucket.update!(bucket, numbers: [1, 2, 3])
+    bucket.numbers.should eq([1, 2, 3])
+    bucket = SaveBucket.update!(bucket, numbers: nil)
     bucket.numbers.should be_nil
   end
 end
--- a/spec/array_column_spec.cr
+++ b/spec/array_column_spec.cr
@@ -18,4 +18,10 @@ describe "Array Columns" do
     bucket = BucketQuery.new.last
     bucket.floaty_numbers.should eq([1.1, 2.2, 3.3, 4.4])
   end
+
+  it "handles optional Array" do
+    BucketBox.create &.numbers(nil)
+    bucket = BucketQuery.new.last
+    bucket.numbers.should be_nil
+  end
 end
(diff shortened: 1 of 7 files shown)


src/avram/model.cr
```diff
@@ -247,13 +247,8 @@ abstract class Avram::Model
...
       {% db_type = column[:type].is_a?(Generic) ? column[:type].type_vars.first : column[:type] %}
-      def {{column[:name]}} : {% if column[:nilable] %}::Union({{db_type}}, ::Nil){% else %}{{column[:type]}}{% end %}
-        %from_db = {{ db_type }}::Lucky.from_db!(@{{column[:name]}})
-        {% if column[:nilable] %}
-          %from_db.as?({{db_type}})
-        {% else %}
-          %from_db.as({{column[:type]}})
-        {% end %}
+      def {{column[:name]}} : {{column[:type]}}{{(column[:nilable] ? "?" : "").id}}
+        {{ db_type }}::Lucky.from_db!(@{{column[:name]}})
```
jwoertink, line 251: I just realized not all types have a `from_db!` method. [bool](https://github.com/luckyframework/avram/blob/master/src/avram/charms/bool_extensions.cr) and [uuid](https://github.com/luckyframework/avram/blob/master/src/avram/charms/uuid_extensions.cr). I wonder if these are bugs that just haven't been caught yet 🤔

src/avram/migrator/alter_table_statement.cr
```diff
@@ -112,18 +112,17 @@ class Avram::Migrator::AlterTableStatement
   end
 
   macro add(type_declaration, index = false, using = :btree, unique = false, default = nil, fill_existing_with = nil, **type_options)
-    {% if type_declaration.type.is_a?(Union) %}
-      {% type = type_declaration.type.types.first %}
+    {% type = type_declaration.type %}
```
jwoertink, line 115: Yeah, this is much cleaner!

spec/array_column_spec.cr
```diff
@@ -18,4 +18,10 @@ describe "Array Columns" do
     bucket = BucketQuery.new.last
     bucket.floaty_numbers.should eq([1.1, 2.2, 3.3, 4.4])
   end
+
+  it "handles optional Array" do
+    BucketBox.create &.numbers(nil)
+    bucket = BucketQuery.new.last
+    bucket.numbers.should be_nil
```
jwoertink, line 25: I'm pretty sure it works, but just for peace of mind, can you add to this spec where we take this bucket and update it to have an empty array? Or maybe add another spec in here that just tests a SaveOperation can have a nilable array but be updated to an empty array?

jwoertink (commented):
Good call on this! I think the implementation is definitely much cleaner this way. I'd like to see just 1 more spec for a little extra safety, and then I think we're good 👍

src/avram/model.cr
matthewmcgarvey: I think they just use the default `from_db!` coming from `Avram::Type` https://github.com/luckyframework/avram/blob/master/src/avram/type.cr#L6

jwoertink: ohh... sneaky 😂

jwoertink (approved):

PR #472, opened 2020-09-26
Add Avram::Type module directly to base types

matthewmcgarvey:
By monkey patching `Avram::Type` onto the base types instead of a nested `Lucky` module, we can take advantaged of the `self` type and remove a bunch of unnecessary methods.

This has the disadvantage of adding methods to Crystal (which is why I updated the methods to start with underscore) so feel free to say no.

### DIFF
--- a/spec/support/custom_email.cr
+++ b/spec/support/custom_email.cr
@@ -1,4 +1,22 @@
 class CustomEmail
+  extend Avram::Type
+
+  def self._parse_attribute(value : CustomEmail)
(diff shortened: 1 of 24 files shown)


confact:
@matthewmcgarvey it feels like you are not done here yet, I can be wrong ^^

But I do like it as I had problems understanding the Avram::Type myself at first when doing Enum support. This does look nicer in that way.

What do you think of instead having method names like `_parse_attribute` to have a `lucky`-prefix? So we know it is actually from Lucky?

like:
```
lucky_parse
```
Or similar.

What do you think about that?

PR #474, opened 2020-10-03
overwrite to_s and to_i for Enums

confact:
making it possible to get int and the string directly, good for showing the enum on pages and such.

Not it will return class instance on `to_s` and not work at all on `to_i`

now it will return say `Ongoing` on to_s and `0` on to_i

not sure if maybe this will work instead:
```
delegate :to_s, :to_i, to: @enum
```

I feel it is a personal taste that feels best. I can change to that if you want.

### DIFF
--- a/spec/type_extensions/enum_spec.cr
+++ b/spec/type_extensions/enum_spec.cr
@@ -25,7 +25,7 @@ describe "Enum" do
     issue.status.opened?.should eq(true)
     issue.status.value.should eq(0)
   end
-  
+
   it "access enum to_s and to_i" do
     issue = IssueBox.create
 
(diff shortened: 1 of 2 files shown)


jwoertink (commented):
We do use some delegation in Lucky as well https://github.com/luckyframework/lucky/search?q=delegate so I'd be ok with that. Also, right below this is the `forward_missing_to @enum`. Does that not already forward `to_s` and `to_i`? Or is it because all objects kind of respond to `to_s` that it's not really forwarded?

confact:
@jwoertink as `to_s` and `to_i` is not missed as it exists in more or less every object it will not forward those. 

I moved it to delegate now.

jwoertink (approved):
Sweet. Looks good to me! Thanks 😄

PR #478, opened 2020-10-13
migrator can handle Int64

wontruefree:
This is because coachroachdb stores the version as an Int64.

This seems to me it should be an int64 because it should be a bigint.
https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L114

### DIFF
--- a/src/avram/migrator/migration.cr
+++ b/src/avram/migrator/migration.cr
@@ -62,7 +62,7 @@ abstract class Avram::Migrator::Migration::V1
 
   def migrated?
     DB.open(Avram::Migrator::Runner.database_url) do |db|
-      db.query_one? "SELECT id FROM migrations WHERE version = $1", version, as: Int32
+      db.query_one? "SELECT id FROM migrations WHERE version = $1", version, as: Int
     end
   end
 


src/avram/migrator/migration.cr
```diff
@@ -62,7 +62,7 @@ abstract class Avram::Migrator::Migration::V1
 
   def migrated?
     DB.open(Avram::Migrator::Runner.database_url) do |db|
-      db.query_one? "SELECT id FROM migrations WHERE version = $1", version, as: Int32
+      db.query_one? "SELECT id FROM migrations WHERE version = $1", version, as: Int
```
jwoertink, line 65: I think you're right.. This should probably be `Int64`. I'm not sure how well supported using the generic `Int` is. I've had issues with it a few times. Let's try `Int64` and see if that works out.

wontruefree: Int64 seems to cause some problems for me on postgres.  I will have to look into it more.

jwoertink: I'm actually surprised that crystal-db didn't blow up on that before. I wonder if that's an issue? 🤔  Let me know what you find out.

jwoertink (commented):
Just coming back to this, I realized I read the SQL wrong the first time.

This is only selecting `id` not `version`. We actually define `version` as `bigint` which is `Int64` already. https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L113-L114

The `id` is defined as `serial` https://github.com/luckyframework/avram/blob/master/src/avram/migrator/columns/primary_keys/int32_primary_key.cr#L9 which is just `Int32`. 

I'm gonna close this PR since there's actually no change to be made. I would like to know about the issue you were having with cockroach though. I wonder if there's an issue somewhere else? 🤔

PR #481, opened 2020-10-15
Better callbacks for SaveOperation

### LINKED ISSUE #480: before_save callbacks may run multiple times
This is related to https://github.com/luckyframework/avram/pull/469

In this case, some `before_save` callbacks may run twice because they are ran when calling `valid?` which could get called multiple times. 

A fix will be in the next PR series for https://github.com/luckyframework/avram/pull/369

jwoertink:
This PR is the next in the series for https://github.com/luckyframework/avram/pull/369 which covers a few things related to callbacks.

Fixes #480 

* Fixes using `after_save` and `after_commit` with blocks which previously raised errors.
* Added new `skip_before_save`, `skip_after_save`, and `skip_after_commit` methods that allow for skipping named callbacks. Not able to skip block callbacks.
* Breaks calling the `valid?` method to run the `before_save` callbacks. 

For a little more info on that last item, the current implementation allows you to manually call `valid?` which will [run before_save callbacks](https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L220). Right now it's possible that before_save could be called [here](https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L317), and [here](https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L338) if the first condition failed.

By moving the `before_save` to the actual `save` method, it now only runs once. We call `save` internally when calling `SaveOp.create` or `SaveOp.update`, and the `valid?` method is called by the `save` method. This means that as long as you use the SaveOperations like normal, then everything should be fine, but if you need to instantiate a SaveOperation, and call `valid?` manually, then things may get a little weird. 

Another one I wanted to fix was https://github.com/luckyframework/avram/issues/432 but there's a bit to discuss on this one as well. Currently if you call `SaveOp.update(thing)`, but the `thing` doesn't have any updates, then no SQL is ran. This is nice because it faster, and saves on query. The downside is that your `after_save` and `after_commit` callbacks will never run. As the issue indicates, some people may want these to run, but we document our `after_save` callback that it's save to run other saves since the whole thing could be rolled back. If no DB query is actually made, then there's no rollback that happens.

We have a few options that I can think of here:

* Run the `after_save` in that second branch, but wrap it in some catch to tell this parent SaveOp that it failed if the after_saves fail or whatever... Then if it succeeds we run the after_commit.
* Just add in the after_save and after_commit but let the user figure out their own specifics (this just sounds like asking for trouble)
* Technically, `SaveOperation` has access to `after_run`, it's just never called. We could add that in as a "hey, no DB stuff happened, but you can still run this other callback".
* Leave as is and just document that this is a side affect

The first option seems like the most plausible to me as far as fixing the issue, but it also makes me think things may get a bit more complicated with that conditional branch.

### DIFF
--- a/spec/save_operation_callbacks_spec.cr
+++ b/spec/save_operation_callbacks_spec.cr
@@ -1,11 +1,20 @@
 require "./spec_helper"
 
+module TestableOperation
+  macro included
+    @callbacks_that_ran = [] of String
+    getter callbacks_that_ran
+
+    def mark_callback(callback_name : String)
+      @callbacks_that_ran << callback_name
+    end
+  end
+end
+
 private class CallbacksSaveOperation < Post::SaveOperation
+  include TestableOperation
   needs rollback : Bool = false
   needs skip_set_required : Bool = false
-
-  @callbacks_that_ran = [] of String
-  getter callbacks_that_ran
 
   before_save setup_required_attributes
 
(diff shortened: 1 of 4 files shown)


spec/save_operation_callbacks_spec.cr
```diff
@@ -74,14 +83,49 @@ private class SaveLineItemSub < SaveLineItemBase
...
+private class UpdateOperationWithSkipCallbacks < SaveOperationWithCallbacks
+  permit_columns :title
+  skip_before_save :set_title
+  skip_after_save :notify_save_complete
+  skip_after_commit :notify_commit_complete
+end
+
 describe "Avram::SaveOperation callbacks" do
-  it "does not run after_* callbacks if just validating" do
+  it "does not run any callbacks if just validating" do
```
jwoertink, line 124: This is because `save` must be called now to kick off the `before_save` callbacks.

src/avram/callbacks.cr
```diff
@@ -176,36 +225,51 @@ module Avram::Callbacks
...
+  #   after_commit do |comment|
+  #     NewCommentNotificationEmail.new(comment, to: comment.author!).deliver_now
+  #   end
+  # end
+  # ```
+  macro after_commit(&block)
+    {%
+      if block.args.size != 1
+        raise <<-ERR
+        The 'after_commit' callback requires only 1 block arg to be passed.
```
jwoertink, line 254: oops. Forgot to update this error to match the others.

matthewmcgarvey (approved):

matthewmcgarvey:
I'm not super fond of redefining methods, though I'm not blocking the PR. I wonder if there is a way we could define an instance-level hash of method names to a block so that we would just add or remove from it and iterate over it in the callbacks.

jwoertink:
Yeah, this does have the drawback that if you try to skip a callback that's defined later, then it's not actually skipped. But my thought was why would anyone do that? It wouldn't make sense to define a callback and a skip at the same level, unless we updated the skips to be conditional. I think conditional callbacks could be added later though. For now we could just document that skipping should only be used at a subclass level especially since a future PR will make subclassing SaveOperations more common.

paulcsmith:
> The downside is that your after_save and after_commit callbacks will never run. As the issue indicates, some people may want these to run, but we document our after_save callback that it's save to run other saves since the whole thing could be rolled back. If no DB query is actually made, then there's no rollback that happens

I think this may be a feature not bug. Running afer save and after commit callbacks if nothing was saved or committed seems surprising to me. I'd say that we keep it as-is for now, and maybe provide some other mechanism or some other set of callbacks that explicitly run no matter what. Something like `after_successful_validation` or something. So it is clear it runs regardless of whether it was actually "saved" or not

PR #484, opened 2020-10-16
Operation cleanup

jwoertink:
This PR is the next in the series for https://github.com/luckyframework/avram/pull/369 which doesn't really add anything. This is more of a general cleanup start. I didn't want to get too crazy.

* Remove all the remnants of the old Forms and past Operation stuff that's been gone for a few versions now
* Moved all the operation related specs to its own folder for a bit of organization.

I didn't move the attribute or the validation specs. I was on the fence, but If anyone has a preference I could go either way. 

On a side note, this is just related to Operations, but I'd eventually like to do this for all sections (i.e. models, queries, etc...)

### DIFF
--- a/spec/operations/define_attribute_spec.cr
+++ b/spec/operations/define_attribute_spec.cr
@@ -1,4 +1,4 @@
-require "./spec_helper"
+require "../spec_helper"
 
 private class OperationWithAttributes < Avram::Operation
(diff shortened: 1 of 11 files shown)


matthewmcgarvey (approved):
+1 for organization! 🎉🎉🎉

PR #485, opened 2020-10-16
Add compile-time error when passing a raw Hash in to SaveOperation.create/update

### LINKED ISSUE #313: When you try to pass a hash to operation methods like create/update, show how to use named args instead

jwoertink:
Fixes #313 

A part of the https://github.com/luckyframework/avram/pull/369 series. This is just an extra development level catch in case you pass a raw hash in to a SaveOperation create or update. I tested this by adding a spec that does it. 

On a side note, testing compile-time errors is impossible. I wonder if we should create a spec directory with commented out tests that would be designed for just having a way to make sure they work the way we think... 🤔  it would be a little weird and messy, but something to think about.

### DIFF
--- a/src/avram/needy_initializer_and_save_methods.cr
+++ b/src/avram/needy_initializer_and_save_methods.cr
@@ -80,7 +80,30 @@ module Avram::NeedyInitializerAndSaveMethods
     generate_save_methods({{ attribute_method_args }}, {{ attribute_params }})
   end
 
+  macro hash_is_not_allowed_helpful_error(method)
+    {% raise <<-ERROR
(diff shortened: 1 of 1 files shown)


matthewmcgarvey (approved):
Looks good.  Check out https://github.com/icy-arctic-fox/spectator/issues/15 we might actually be able to test compilation errors

jwoertink:
oh that's nice. That would be great for us to have something like that. Maybe we use that in a separate shard we can include in all our projects for testing that stuff?

PR #486, opened 2020-10-18
Remove register_setup_step macro

matthewmcgarvey:
Connected to #453 

In order to support views well, we need to not generate the operation code for view classes. The way that the code was setup, there was no easy way to avoid setting it up. Since this was not a documented API, I believe it is safe to remove.


jwoertink (approved):
I'm not sure what the original intention of this was. Maybe just a way to ensure that everything runs? I think it's fine to remove too since this isn't something that needs to be user accessible. It's more like an internal setup deal.

paulcsmith (commented):
Sorry for looking at this so late. The point of register_setup_step was that third party libraries could register new setup steps. Maybe setting up new methods, or generating classes like the Acram:SaveOperationTemplate.setup.

I just never got around to documenting it, but it would be pretty powerful. 

My thought would be that we add `register_setup_step` again, but only register the steps where needed. That may mean we need to rework register_setup_step to work on aper-model basis though.

So summary is this is fine as-is since I don't think anyone was using it, but ideally we'd add some version of it that works per-model and document it so it can be used by third party libs. For example you might `include Shrine::Model` and it'll add some code to your operation automatically, or add some methods that are for handling uploads.

This is not a high priority but just wanted to share the background and a possible solution

PR #492, opened 2020-10-21
Clarify which method is preferred + fix update example

paulcsmith:
Make it clearer which one is preferred. The hash conversion may be useful in some spots but named args are better in the vast majority of cases because it can catch errors with column names, or value types. The hash on the other hand can have incorrect keys or values

Also adds `record` to the update example

Screen Shot 2020-10-21 at 3 26 26 PM

### DIFF
--- a/src/avram/needy_initializer_and_save_methods.cr
+++ b/src/avram/needy_initializer_and_save_methods.cr
@@ -86,8 +86,8 @@ module Avram::NeedyInitializerAndSaveMethods
 
       Try this...
 
-        * Pass named arguments - #{@type}.#{method.id}(title: "My Title")
-        * Convert hash to params - Avram::Params.new({"title" => "My Title"})
+        * Use named arguments (recommended)  - #{@type}.#{method.id}(title: "My Title")
+        * Convert hash to params (not as safe) - Avram::Params.new({"title" => "My Title"})
 
       ERROR
     %}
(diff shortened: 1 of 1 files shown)


spec/operations/save_operation_spec.cr
```diff
@@ -91,6 +91,7 @@ describe "Avram::SaveOperation" do
     avram_params = Avram::Params.new({"title" => "Test", "published_at" => ""})
 
     post = SavePost.create!(avram_params)
+    SavePost.update!(post, {"foo" => "bar"})
```
jwoertink, line 94: This probably has to be removed, right?

src/avram/needy_initializer_and_save_methods.cr
```diff
@@ -142,7 +142,7 @@ module Avram::NeedyInitializerAndSaveMethods
       {% else %}
         yield nil, nil
       {% end %}
-      hash_is_not_allowed_helpful_error(:update{% if with_bang %}!{% end %})
+      hash_is_not_allowed_helpful_error(:update{% if with_bang %}!{% end %}, additional_args: "record, ")
```
jwoertink, line 145: ah, good call! I wasn't sure how to best do that.

jwoertink (commented):
Yeah, this is much nicer. Thanks!

spec/operations/save_operation_spec.cr
paulcsmith: Ooops lol

PR #494, opened 2020-10-21
Add ability to use custom association names

### LINKED ISSUE #473: Naming associations something other than the model causes issues
It sometimes makes sense to name an association something other than the model name it points to. For example, you might have a `Book` model and it makes more sense for it to have and `author` than just a `user`. When this is done with Avram models, it can cause issues because we assume the table from the variable name in some places and use the underlying table name to generate code in others which creates a confusing and sometimes broken api.

It is possible to pass a `table_name` flag with this appropriately named table to a `belongs_to` relationship but it is undocumented and confusing. I believe that each association should use the given association name where ever possible and delegate to class methods on the provided type where the association name won't work.

- [ ] Fix belongs_to
- [ ] Fix has_many
- [ ] Fix has_one

### has_many

When declaring a has_many relationship with a custom name, the generated join method is still using the table name instead of the given association name.

```crystal
class Boss < BaseModel
  table do
    has_many employees : User
  end
end
```

The generated join should result in queries being made like this:

```crystal
BossQuery.new.where_employees(UserQuery.new.name("asdfasf"))
```

but it produces a compilation error because the underlying `inner_join_xxx` methods use the table name.

matthewmcgarvey:
Fixes #473 

This updates Avram associations to always use the given association name rather than the underlying types table name where ever possible.

## Breaking Changes

We always used the table name when generating `where_xxx` methods, now we always use the association name. The most common way this will break will be for `belongs_to` relationships.

With a model like

```crystal
class Book
  table do
    belongs_to user : User
  end
end
```

### Before

```crystal
# notice the pluralization because it used the table name
BookQuery.new.where_users(...)
```

### After

```crystal
# singular because it used the association name given in the model
BookQuery.new.where_user(...)
```

### DIFF
--- a/src/avram/associations/has_many.cr
+++ b/src/avram/associations/has_many.cr
@@ -1,7 +1,13 @@
 module Avram::Associations::HasMany
-  macro has_many(type_declaration, through = nil, foreign_key = nil)
+  macro has_many(type_declaration, through = nil, source = nil, foreign_key = nil)
     {% if !through.is_a?(NilLiteral) && !through.is_a?(SymbolLiteral) %}
       {% through.raise "The association name for 'through' must be a Symbol. Instead, got: #{through}" %}
+    {% end %}
+    {% if !source.is_a?(NilLiteral) && !source.is_a?(SymbolLiteral) %}
+      {% source.raise "The value for 'source' must be a Symbol. Instead, got: #{source}" %}
+    {% end %}
+    {% if !source.is_a?(NilLiteral) && through.is_a?(NilLiteral) %}
+      {% source.raise "Source must only be specified when defining a has many through relationship." %}
     {% end %}
     {% assoc_name = type_declaration.var %}
 
@@ -16,6 +22,7 @@ module Avram::Associations::HasMany
       type: {{ type_declaration.type }},
       foreign_key: :{{ foreign_key }},
       through: {{ through }},
+      source: {{ source }},
       relationship_type: :has_many
 
     {% model = type_declaration.type %}
--- a/src/avram/associations/has_many.cr
+++ b/src/avram/associations/has_many.cr
@@ -12,7 +12,7 @@ module Avram::Associations::HasMany
     {% foreign_key = foreign_key.id %}
 
     association \
-      table_name: :{{ assoc_name }},
+      assoc_name: :{{ assoc_name }},
       type: {{ type_declaration.type }},
       foreign_key: :{{ foreign_key }},
       through: {{ through }},
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -113,8 +113,15 @@ class Avram::BaseQueryTemplate
               {{ join_type.downcase.id }}_join_{{ assoc[:through].id }}
                 .__yield_where_{{ assoc[:through].id }} do |join_query|
                   {% if assoc[:source].is_a?(NilLiteral) %}
-                    {% singular_association_name = run("../run_macros/singularize_word.cr", assoc[:assoc_name]) %}
-                    join_query.{{ join_type.downcase.id }}_join_{{ singular_association_name.id }}
+                    # check if the join query has a has_many joins method (default)
+                    if join_query.responds_to?(:{{ join_type.downcase.id }}_join_{{ assoc[:assoc_name].id }})
+                      # through a has_many association
+                      join_query.{{ join_type.downcase.id }}_join_{{ assoc[:assoc_name].id }}
+                    else
+                      # through a belongs_to association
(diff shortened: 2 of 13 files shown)


src/avram/base_query_template.cr
```diff
@@ -112,13 +112,14 @@ class Avram::BaseQueryTemplate
             {% elsif assoc[:through] %}
               {{ join_type.downcase.id }}_join_{{ assoc[:through].id }}
                 .__yield_where_{{ assoc[:through].id }} do |join_query|
-                  join_query.{{ join_type.downcase.id }}_join_{{ assoc[:table_name] }}
+                  {% singular_association_name = run("../run_macros/singularize_word.cr", assoc[:assoc_name]) %}
```
matthewmcgarvey, line 115: This needs to be conditional. Only associations where the "through" class has a "belongs_to" or a "has_one" association should be singularized, otherwise we should use the association name

We will also have to add in an optional property to specify the association name when it's something totally different.

src/avram/associations/has_many.cr
```diff
@@ -1,8 +1,14 @@
 module Avram::Associations::HasMany
-  macro has_many(type_declaration, through = nil, foreign_key = nil)
+  macro has_many(type_declaration, through = nil, source = nil, foreign_key = nil)
```
matthewmcgarvey, line 2: Source is a property that can be passed to dictate the exact association we look for on the connecting model. If it is not specified we singularize the association name. That means that source is required if you want to create a has_many through a has_many relationship because the relationship name would be plural.

I can't think of a good way to automatically handle calling the correct methods on a has_many AND a belongs_to. I also can't find any examples of a compiled, type-safe, Rails-like ORM to learn from.

src/avram/base_query_template.cr
```diff
@@ -112,13 +112,18 @@ class Avram::BaseQueryTemplate
             {% elsif assoc[:through] %}
               {{ join_type.downcase.id }}_join_{{ assoc[:through].id }}
                 .__yield_where_{{ assoc[:through].id }} do |join_query|
-                  join_query.{{ join_type.downcase.id }}_join_{{ assoc[:table_name] }}
+                  {% if assoc[:source].is_a?(NilLiteral) %}
+                    {% singular_association_name = run("../run_macros/singularize_word.cr", assoc[:assoc_name]) %}
+                    join_query.{{ join_type.downcase.id }}_join_{{ singular_association_name.id }}
+                  {% else %}
+                    join_query.{{ join_type.downcase.id }}_join_{{ assoc[:source].id }}
+                  {% end %}
```
matthewmcgarvey, line 120: This is where we make the choice about the has_many through join. If a source is not specified we singularize the association name (we assume it is a belongs_to relationship)

vjousse:
I've got the same problem, an example from my own code below.

## Model
```crystal
class Article < BaseModel
  table do
    column slug : String
    column title : String
    column description : String
    column body : String
    belongs_to author : User
    has_many taggings : Tagging
    has_many tags : Tag, through: :taggings
    has_many favoritings : Favoriting
    has_many favoriting_users : User, through: :favoritings
  end
end
```

## Expected query
```crystal
# Article has only one author, so I expect it to be singular
ArticleQuery.new.where_author(…)
```

## Current query
```crystal
# where_users doesn't match any of my association names
ArticleQuery.new.where_users(…)
```

Not sure if it's related, but when I try to join my `has_many` relationship to `User` through my `:favoriting` table, I have a compile time error

## Code of the query
```crystal
articles_query = articles_query.where_favoriting_users(UserQuery.new.username(favorited_filter))
```

## Compile time error
```crystar
web          | There was a problem expanding macro 'setup'
web          |
web          | Code in macro 'setup'
web          |
web          |  49 | Avram::BaseQueryTemplate.setup(
web          |       ^
web          | Called macro defined in lib/avram/src/avram/base_query_template.cr:2:3
web          |
web          |  2 | macro setup(type, columns, associations, table_name, primary_key_name, *args, **named_args)
web          |
web          | Which expanded to:
web          |
web          |  > 576 | inner_join_favoritings
web          |  > 577 | __yield_where_favoritings do |join_query|
web          |  > 578 |   join_query.inner_join_favoriting_users
web          |                       ^--------------------------
web          | Error: undefined method 'inner_join_favoriting_users' for Favoriting::BaseQuery
```

paulcsmith:
/sparkle for thorough example and explanation. 

@vjousse I think the compile error is related to this so once we figure this out it should fix the issue you're describing!

matthewmcgarvey:
I was not able to add a spec around a `has_many through` a `has_many` relationship because of https://github.com/luckyframework/avram/issues/496

matthewmcgarvey:
Replaced by https://github.com/luckyframework/avram/pull/525

PR #495, opened 2020-10-22
Adding conditional callbacks

### LINKED ISSUE #483: Conditional callbacks
With the new addition of https://github.com/luckyframework/avram/pull/481 we can now skip some callbacks. However, this implementation does have a drawback where it redefines the callback method to do nothing which means that if the skip callback is defined before the actual callback, then it won't actually be skipped.

```crystal
  before_save :set_thing
  skip_before_save :set_thing

  private def set_thing
    # This is still executed
  end
end

# you'd have to do something funky like
  before_save :set_thing

  private def set_thing
    # This is still executed
  end

  # this would redefine the method later allowing you to actually skip it
  skip_before_save :set_thing
end
```

Now, I don't think anyone would actually write code like either of these examples, but if we had conditional callbacks they might. Imagine something like this:

```crystal
  before_save :set_thing
  skip_before_save :set_thing, if: ->(user) { user.admin? }

  private def set_thing
    # This is still executed
  end
end
```

In this case, we may need to think about a system to register callbacks at the instance level. I kinda like what @matthewmcgarvey [suggested here](https://github.com/luckyframework/avram/pull/481#issuecomment-709638338). 

Now, I could have sworn we had some sort of conditional on the `before_save` before like where you could say `before_save :whatever, on: :create`... but I couldn't find it. I wanted to see if there was a reason we removed it so if anyone remembers if we had that before, let me know. 

Another idea I just had, which may be a bit a crazy, is to change the callbacks to be segmented in to a more type-safe way... Just thinking out loud here, but what about this?

```crystal
  before_save UserSetThingCallback.new

end

class UserSetThingCallback
  include Avram::CallbackBeforePipe

  def call
    # runs the before save
  end

  def skip_call(user)
    # if this is defined
    # and returns true, it would skip running the call.
    # you could just have it return `true` to always skip...
  end
end

module Avram::CallbackBeforePipe
  macro included
     abstract def call
     # whatever else
     
  end
end
```

Obviously a "rough" sketch, and would be another major refactor, but worth a shot possibly...

jwoertink:
Fixes #483 

This PR does 2 things.
* It removes the `skip_*` macros that I just put in the other day
* It adds in the ability to conditionally run callbacks

You can read through the issue for more info, but a quick TL;DR: You can still skip callbacks, it's just done by override a method yourself instead of a macro doing it for you.

This allows us to now do things like this:

```crystal
abstract class SaveTransaction < Transaction::SaveOperation
  before_save :validate_card_number, if: :new_transaction?
  def validate_card_number
    card_number.value == 42
  end
end

class CreateTransaction < SaveTransaction
  permit_columns :card_number, :exp_date
  private def new_transaction?
    true
  end
end

class UpdateTransaction < SaveTransaction
  permit_columns :exp_date
  private def new_transaction?
    false
  end
end
```

### DIFF
--- a/spec/operations/save_operation_callbacks_spec.cr
+++ b/spec/operations/save_operation_callbacks_spec.cr
@@ -128,16 +128,16 @@ end
 private class UpdateOperationWithSkipCallbacks < SaveOperationWithCallbacks
   permit_columns :title
 
-  before_save(unless: :skip_callback) { mark_callback("before_save_in_a_block2") }
-  after_save(unless: :skip_callback) do |saved_post|
+  before_save(unless: :skip_callback?) { mark_callback("before_save_in_a_block2") }
+  after_save(unless: :skip_callback?) do |saved_post|
     mark_callback("after_save_in_a_block with #{saved_post.title}2")
   end
-  after_commit(unless: :skip_callback) do |saved_post|
+  after_commit(unless: :skip_callback?) do |saved_post|
     mark_callback("after_commit_in_a_block with #{saved_post.title}2")
   end
 
   # skip all the blocks
-  private def skip_callback
+  private def skip_callback?
     true
   end
 
--- a/spec/operations/save_operation_callbacks_spec.cr
+++ b/spec/operations/save_operation_callbacks_spec.cr
@@ -86,17 +86,29 @@ end
 private class SaveOperationWithCallbacks < Post::SaveOperation
   include TestableOperation
 
-  before_save :set_title
+  before_save :set_title, if: :title_is_needed?
   before_save { mark_callback("before_save_in_a_block") }
 
-  after_save :notify_save_complete
+  after_save :notify_save_complete, unless: :already_notified?
   after_save do |saved_post|
     mark_callback("after_save_in_a_block with #{saved_post.title}")
   end
 
-  after_commit :notify_commit_complete
+  after_commit :notify_commit_complete, if: :finalize_commit?
   after_commit do |saved_post|
     mark_callback("after_commit_in_a_block with #{saved_post.title}")
+  end
+
+  private def title_is_needed?
+    true
+  end
+
+  private def already_notified?
+    false
+  end
+
+  private def finalize_commit?
+    true
   end
 
   private def set_title
@@ -115,6 +127,34 @@ end
 
 private class UpdateOperationWithSkipCallbacks < SaveOperationWithCallbacks
   permit_columns :title
+
+  before_save(unless: :skip_callback) { mark_callback("before_save_in_a_block2") }
+  after_save(unless: :skip_callback) do |saved_post|
+    mark_callback("after_save_in_a_block with #{saved_post.title}2")
+  end
+  after_commit(unless: :skip_callback) do |saved_post|
+    mark_callback("after_commit_in_a_block with #{saved_post.title}2")
+  end
+
+  # skip all the blocks
+  private def skip_callback
+    true
+  end
+
+  # skip before_save :set_title
+  private def title_is_needed?
+    false
+  end
+
+  # skip after_save :notify_save_complete
+  private def already_notified?
+    true
+  end
+
+  # skip after_commit :notify_commit_complete
+  private def finalize_commit?
+    false
+  end
 end
 
 describe "Avram::SaveOperation callbacks" do
--- a/spec/operations/save_operation_callbacks_spec.cr
+++ b/spec/operations/save_operation_callbacks_spec.cr
@@ -115,9 +115,6 @@ end
 
 private class UpdateOperationWithSkipCallbacks < SaveOperationWithCallbacks
   permit_columns :title
-  skip_before_save :set_title
-  skip_after_save :notify_save_complete
-  skip_after_commit :notify_commit_complete
 end
 
 describe "Avram::SaveOperation callbacks" do
--- a/src/avram/callbacks.cr
+++ b/src/avram/callbacks.cr
@@ -5,15 +5,24 @@ module Avram::Callbacks
   # You can set defaults, validate, or perform any other setup necessary for
   # saving.
   #
+  # Optionally you can specify a symbol to `if` or `unless` which allows you to
+  # run this conditionally. The symbol should reference a method you've defined
+  # that returns a truthy/falsey value
+  #
   # ```
   # before_save :run_validations
+  # before_save :run_validations, unless: :admin_bypass?
   #
   # private def run_validations
   #   validate_required name, age
   # end
+  #
+  # private def admin_bypass?
+  #   respond_to?(:user) && user.admin?
+  # end
   # ```
-  macro before_save(method_name)
-    before_save do
+  macro before_save(method_name, if _if = nil, unless _unless = nil)
+    before_save(if: {{ _if }}, unless: {{ _unless }}) do
       {{ method_name.id }}
     end
   end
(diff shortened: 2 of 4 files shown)


src/avram/callbacks.cr
```diff
@@ -45,20 +48,57 @@ module Avram::Callbacks
...
+  # private def skip_callback?
+  #   false
+  # end
   # ```
-  macro before_save
+  macro before_save(if _if = nil, unless _unless = nil)
+    {% if _if != nil && _unless != nil %}
+      {% raise "Your before_save callbacks should only specify `if` or `unless`, but not both." %}
+    {% end %}
+    {% unless _if.is_a?(SymbolLiteral) || _if.is_a?(NilLiteral) %}
```
matthewmcgarvey, line 68: Need to add this for unless as well, and in the other macros that are missing it

matthewmcgarvey (changes requested):
Just missing the error for an invalid unless, otherwise it looks good and I'm glad we can remove those skips.

spec/operations/save_operation_callbacks_spec.cr
```diff
@@ -115,9 +127,34 @@ end
...
+    mark_callback("after_save_in_a_block with #{saved_post.title}2")
+  end
+  after_commit(unless: :skip_callback) do |saved_post|
+    mark_callback("after_commit_in_a_block with #{saved_post.title}2")
+  end
+
+  # skip all the blocks
+  private def skip_callback
+    true
+  end
```
paulcsmith, line 142: The diff is hard to read, but I changed `skip_callback` to `skip_callback?`

```suggestion
  before_save(unless: :skip_callback?) { mark_callback("before_save_in_a_block2") }
  after_save(unless: :skip_callback?) do |saved_post|
    mark_callback("after_save_in_a_block with #{saved_post.title}2")
  end
  after_commit(unless: :skip_callback?) do |saved_post|
    mark_callback("after_commit_in_a_block with #{saved_post.title}2")
  end

  # skip all the blocks
  private def skip_callback?
    true
  end
```

src/avram/callbacks.cr
```diff
@@ -5,25 +5,28 @@ module Avram::Callbacks
...
   # before_save :run_validations
+  # before_save :run_validations, unless: :admin_bypass?
   #
   # private def run_validations
   #   validate_required name, age
   # end
+  #
+  # private def admin_bypass?
+  #   respond_to?(:user) && user.admin?
+  # end
```
paulcsmith, line 22:
```suggestion
  # before_save :validate_can_internet, unless: :too_cool_for_school?
  #
  # private def run_validations
  #   validate_required name, age
  # end
  #
  # private def validate_can_internet
  #   validate_size_of age, min: 13
  # end
  #
  # private def too_cool_for_school?
  #   [true, false].sample
  # end
```

This is probably a bad example, but something simpleish

paulcsmith: And maybe a more advanced example could go below. Like "Advanced if/unless" that shows using it with field values and passed in `needs`. But that can come later. Not a blocker

src/avram/callbacks.cr
```diff
@@ -45,20 +48,57 @@ module Avram::Callbacks
...
   # ```
-  macro before_save
+  macro before_save(if _if = nil, unless _unless = nil)
+    {% if _if != nil && _unless != nil %}
+      {% raise "Your before_save callbacks should only specify `if` or `unless`, but not both." %}
+    {% end %}
+    {% unless _if.is_a?(SymbolLiteral) || _if.is_a?(NilLiteral) %}
+      {%
+        raise <<-ERROR
+      You must pass a Symbol to `if`. This is a reference to a method you define.
```
paulcsmith, line 71:
```suggestion
      You must pass a Symbol to `if`. The Symbol will reference a method you define.
```

src/avram/callbacks.cr
```diff
@@ -83,6 +123,10 @@ module Avram::Callbacks
 
   # Run the given method after save, but before transaction is committed
   #
+  # Optionally you can specify a symbol to `if` or `unless` which allows you to
```
paulcsmith, line 126:
```suggestion
  # Optionally you can pass an  `if` or `unless` argument which allows you to
```

src/avram/callbacks.cr
```diff
@@ -102,20 +146,18 @@ module Avram::Callbacks
...
 
-  # Redefines the given `method_name` method to do nothing
-  macro skip_after_save(method_name)
-    def {{ method_name.id }}(_object)
-    end
-  end
-
   # Run the given block after save, but before transaction is committed
   #
+  # Optionally you can specify a symbol to `if` or `unless` which allows you to
```
paulcsmith, line 157:
```suggestion
  # Optionally you can pass an  `if` or `unless` argument which allows you to
```

src/avram/callbacks.cr
```diff
@@ -212,6 +284,10 @@ module Avram::Callbacks
 
   # Run the given method after save and after successful transaction commit
   #
+  # Optionally you can specify a symbol to `if` or `unless` which allows you to
```
paulcsmith, line 287:
```suggestion
  # Optionally you can pass an  `if` or `unless` argument which allows you to
```

paulcsmith (approved):
So stoked for this!

jwoertink:
I'm moving these compile errors to look like this: 

Screen Shot 2020-10-22 at 3 27 58 PM

jwoertink:
weird... no clue why ameba failed randomly... Probably a fluke though.

jwoertink:
Ok, turns out it's doing this locally as well

Screen Shot 2020-10-22 at 3 59 42 PM

matthewmcgarvey:
@jwoertink it seems to be the Callbacks module. Here is the most minimal I'm able to make the file and still cause that error

```crystal
module Avram::Callbacks
  macro conditional_error_for_inline_callbacks
    \{%
      raise ""
    %}
  end

  macro before_save(if _if = nil)
  end
end
```

jwoertink:
It seems I've unearthed some ameba bugs... https://github.com/crystal-ameba/ameba/issues/169 😂  oops..

veelenga:
@jwoertink the buggy rule is disabled for a while. I think this work is not blocked anymore by Ameba :) Thanks for reporting.

jwoertink:
@veelenga cool, thanks! I was mainly keeping it open so we could reference it and use it for a "test ground" if we needed.

src/avram/callbacks.cr
```diff
@@ -1,29 +1,77 @@
 module Avram::Callbacks
+  # :nodoc:
+  macro conditional_error_for_inline_callbacks(callback, method_name, condition)
+    \{%
```
veelenga, line 4: @jwoertink could you please elaborate on why the leading slash is needed here? I think this is causing the compiler to report the incorrect location.

As I can see from the doc it is used to prevent evaluating a macro **the outer macro**, which is not the case here.

https://crystal-lang.org/reference/syntax_and_semantics/macros.html#nested-macros

jwoertink: I have the macro interpolation in this, and without the leading slashes, the compile-time error produces the curly brace literals. 

```
# with the leading slash, I get
You must pass a Symbol to `if`

# without the leading slash, I get
You must pass a Symbol to `{{ condition.id }}`
```

matthewmcgarvey (approved):
Woops! didn't realize I was still blocking

jwoertink:
Sweet! Thanks for the update @matthewmcgarvey. Now that it's using Ameba master, it's all passing. Also looks like the crystal issue was fixed in master, so next release of those will be good again 👍

Issue #496, opened 2020-10-22
Compiler error when defining has_many through a has_many relationship


matthewmcgarvey:
https://github.com/luckyframework/avram/blob/dddcf7d4002ac51cbc12b36d15574ca5ee11989a/src/avram/associations/has_many.cr#L90

https://github.com/luckyframework/avram/blob/dddcf7d4002ac51cbc12b36d15574ca5ee11989a/src/avram/associations/has_many.cr#L114

Both of these lines (and maybe more) expect the target association to have a `#preload_xxx` method that matches the "through" name. So in a code example like

```crystal
class Sale < BaseModel
  table do
    belongs_to employee : Employee
  end
end

class Employee < BaseModel
  table do
    has_many sales : Sale
    belongs_to manager : Manager
  end
end

class Manager < BaseModel
  table do
    has_many employees : Employee
    has_many sales : Sale, through: :employees
  end
end
```

Trying to call `manager.sales` with lazy loading enabled causes a compilation error like

```
Showing last frame. Use --error-trace for full trace.

There was a problem expanding macro 'define_has_many_lazy_loading'

Code in macro 'has_many'

 19 | define_has_many_lazy_loading(sales, Sale, manager_id, :employees)
      ^
Called macro defined in src/avram/associations/has_many.cr:72:11

 72 | private macro define_has_many_lazy_loading(assoc_name, model, foreign_key, through)

Which expanded to:

 > 39 |   through_query.manager_id(id)
 > 40 | end
 > 41 | .preload_employees
         ^----------------
Error: undefined method 'preload_employees' for Sale::BaseQuery

Did you mean 'preload_employee'?
```

**NOTE**: this is the first error I ran into when trying to define a has_many through a has_many association. There could be more after this oen.

jwoertink:
related: https://github.com/luckyframework/avram/issues/196

Issue #196, opened 2019-08-01
Has many through a has many through is broke


jwoertink:
This is probably more of an edge case, but if you need a model to associate to another one but through 2 layers of joins, you'll end up with something like this:

Let's say the app is a multi-tenant blog app where a user can post on many different blogs, and each of these posts have comments, so the User has many comments but through the posts that are through the blos

```crystal
table :users do
  has_many blogs : Blog
  has_many posts : Post, through: :blogs
  has_many comments : Comment, through: :posts
end
```

When you try to use `user.comments`, you'll get an error that says: 

> undefined method 'user_id' for Post::BaseQuery

The error would be correct because `posts` doesn't have `user_id`, but `blogs` does. The proper SQL for this would be:

```SQL
SELECT *
FROM users
INNER JOIN blogs ON users.id = blogs.user_id
INNER JOIN posts ON blogs.id = posts.blog_id
INNER JOIN comments ON posts.id = comments.post_id
```

jwoertink:
It seems at some point this may have been fixed. We're using it in an app, and it's not breaking... I'd like to see a spec for this though just to be sure.

PR #499, opened 2020-10-23
Fix bugs in has_many through has_many associations

matthewmcgarvey:
Fixes: #496
Related: #196
Related: #494

There were three places that using a `has_many` through a `has_many` relationship was broken.

1. When lazy loading was on, accessing the relationship resulted in a compilation error
2. When calling `preload_xxx`, it was only handling a `belongs_to` flow
3. When calling `xxx_count`, there was a compilation error

All 3 of these suffered the same issue while number 2 had an extra problem.  They were all assuming that the association pointed back to the intermediary "through" association in a `has_many` way because the intermediary belonged to the association. That would not be the case in a `has_many` through a `has_many` because the association would belong to the intermediary.

## Solution

The solution was to singularize the association ("products" becomes "product") and start using the `responds_to?` method. We have rarely used this method in the codebase because it means we are determining the appropriate method to call at run time instead of compile time, but I cannot find a better way to accomplish this.

### DIFF
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -51,7 +51,7 @@ describe Avram::Model do
       post.tags.should eq [tag]
     end
 
-    it "count has_many through belongs_to associations" do
+    it "counts has_many through belongs_to associations" do
       tag = TagBox.create
       post = PostBox.create
       TagBox.create
@@ -60,7 +60,7 @@ describe Avram::Model do
       post.tags_count.should eq 1
     end
 
-    it "count has_many through has_many associations" do
+    it "counts has_many through has_many associations" do
       manager = ManagerBox.create
       employee = EmployeeBox.new.manager_id(manager.id).create
       BusinessSaleBox.new.employee_id(employee.id).create
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -15,11 +15,14 @@ describe Avram::Model do
     it "gets the related records for nilable association that exists" do
       manager = ManagerBox.create
       employee = EmployeeBox.new.manager_id(manager.id).create
+      business_sale = BusinessSaleBox.new.employee_id(employee.id).create
 
       manager = Manager::BaseQuery.new.find(manager.id)
 
       manager.employees.to_a.should eq [employee]
       employee.manager.should eq manager
+      employee.business_sales.should eq [business_sale]
+      manager.business_sales.should eq [business_sale]
     end
 
     it "returns nil for nilable association that doesn't exist" do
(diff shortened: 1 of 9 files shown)


spec/associations_spec.cr
```diff
@@ -48,14 +51,22 @@ describe Avram::Model do
       post.tags.should eq [tag]
     end
 
-    it "count through associations" do
+    it "count has_many through belongs_to associations" do
```
jwoertink, line 54:
```suggestion
    it "counts has_many through belongs_to associations" do
```

spec/associations_spec.cr
```diff
@@ -48,14 +51,22 @@ describe Avram::Model do
...
       tag = TagBox.create
       post = PostBox.create
       TagBox.create
       TaggingBox.new.tag_id(tag.id).post_id(post.id).create
 
       post.tags_count.should eq 1
     end
+
+    it "count has_many through has_many associations" do
```
jwoertink, line 63:
```suggestion
    it "counts has_many through has_many associations" do
```

spec/support/manager.cr
```diff
@@ -2,5 +2,6 @@ class Manager < BaseModel
   table do
     column name : String
     has_many employees : Employee
+    has_many business_sales : BusinessSale, through: :employees
```
jwoertink, line 5: In #196 I had also run in to an issue of a has_many through a has_many through.. Just to ensure that's also covered, can we maybe add in another `has_many customers : Customer, through: :business_sales`? or something along those lines.

jwoertink (commented):
This will be great to get in, but it's a pretty intense PR so I'd like to have @paulcsmith review as well. Overall it looks good, but only main suggestion would be to also add in a has_many through a has_many through just to cover the specs for #196

src/avram/associations/has_many.cr
```diff
@@ -19,12 +19,16 @@ module Avram::Associations::HasMany
       relationship_type: :has_many
 
     {% model = type_declaration.type %}
+    {% singular_through = nil %}
+    {% if !through.is_a?(NilLiteral) %}
+      {% singular_through = run("../../run_macros/singularize_word.cr", through) %}
```
wontruefree, line 24: I am curious why do we need to use the `run_macro`?

matthewmcgarvey: Because of the way Avram works, we're not able to look at the associations of other models, so there's no way to know if the association we are referencing is a `has_many` (plural) or a `belongs_to` (singular). So a library could have many authors through books, but from library how do we know if books belongs to an author, which would be singular, or have many authors plural? So I added the singularize word macro to turn `authors` into `author` so that we can check which name for the association that book has.

I do not like this solution AT ALL, but with the current limitations of Avram, I don't see a better way of jumping associations.

spec/support/manager.cr
matthewmcgarvey: @jwoertink There are definitely still bugs when doing a has_many through a has_many through but I don't have a good solution for them and it would add a lot more changes to this PR which is already confusing enough. I updated the description to not mark that issue as fixed

matthewmcgarvey:
Replaced by https://github.com/luckyframework/avram/pull/525

PR #500, opened 2020-10-23
Adding new has_key_for? method to params

### LINKED ISSUE #72: Add method to params object to check if it has key for a Form

jwoertink:
Fixes #72
This is another in the https://github.com/luckyframework/avram/pull/369 series.

This doesn't really do much in Avram itself since most of the implementation comes from Lucky::Params.

### DIFF
--- a/src/avram/paramable.cr
+++ b/src/avram/paramable.cr
@@ -3,5 +3,8 @@ module Avram::Paramable
   abstract def nested(key : String) : Hash(String, String)
   abstract def get?(key : String)
   abstract def get(key : String)
-  abstract def has_key_for?(operation : Avram::Operation.class | Avram::SaveOperation.class) : Bool
+
+  def has_key_for?(operation : Avram::Operation.class | Avram::SaveOperation.class) : Bool
+    nested?(operation.param_key).any?
+  end
 end
--- a/src/avram/paramable.cr
+++ b/src/avram/paramable.cr
@@ -3,4 +3,5 @@ module Avram::Paramable
   abstract def nested(key : String) : Hash(String, String)
   abstract def get?(key : String)
   abstract def get(key : String)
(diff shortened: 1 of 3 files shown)


src/avram/params.cr
```diff
@@ -38,4 +38,8 @@ class Avram::Params
   def nested_file(key) : Hash(String, String)
     @hash
   end
+
+  def has_key_for?(operation : Avram::Operation.class | Avram::SaveOperation.class) : Bool
+    nested?(operation.param_key).any?
```
matthewmcgarvey, line 43: Seems like this implementation can go in `Avram::Paramable`?

jwoertink: I thought about that too. I actually wrote the code in Lucky first, then came and did this. I think I'll do that.

src/avram/paramable.cr
```diff
@@ -3,4 +3,8 @@ module Avram::Paramable
   abstract def nested(key : String) : Hash(String, String)
   abstract def get?(key : String)
   abstract def get(key : String)
+
+  def has_key_for?(operation : Avram::Operation.class | Avram::SaveOperation.class) : Bool
+    nested?(operation.param_key).any?
```
jwoertink, line 8: Since Avram::Params are mainly used to circumvent some testing stuff, the actual specs for this will go in Lucky itself.

paulcsmith: I think specs should definitely be here so we know we don't break Avram. 

I think we can still test this by making a fake class that implements this `Paramable` and then tests that it returns the right thing. 

See example here: https://github.com/luckyframework/avram/blob/79668a93943c7706bb775c01b5fd34d02c342c3f/spec/operations/nested_save_operation_spec.cr#L17-L58

PR #505, opened 2020-10-28
Adding in specs for params has_key_for? method

### LINKED ISSUE #503: Add specs for has_key_for? method
https://github.com/luckyframework/avram/pull/500#discussion_r512787102
https://github.com/luckyframework/lucky/pull/1287#issuecomment-717323997

jwoertink:
Fixes #503

### DIFF
--- a/spec/params_spec.cr
+++ b/spec/params_spec.cr
@@ -1,0 +1,96 @@
+require "./spec_helper"
+
+private class TestOperationWithDefaultParamKey < Avram::Operation
+  attribute title : String
+
+  def run
+  end
+end
+
+private class TestOperationWithCustomParamKey < Avram::Operation
+  param_key :test_op
+  attribute title : String
+
+  def run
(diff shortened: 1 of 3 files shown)


spec/params_spec.cr
```diff
@@ -0,0 +1,96 @@
...
+    raise "Not implemented"
+  end
+
+  def nested_file?(key)
+    raise "Not implemented"
+  end
+
+  def nested_file(key)
+    raise "Not implemented"
+  end
```
paulcsmith, line 55: Not necessary to do now at all, but since we do these "fake" params in at least 2 places now, maybe we could have an `Avram::FakeParams` abstract class in `spec_support` that adds these methods in so you don't have to duplicate it.

Up to you though!

paulcsmith (approved):

jwoertink:
Thanks for the review!

PR #506, opened 2020-10-28
General core cleanup

matthewmcgarvey:
I've been doing some experiments and using the "core" of Avram to avoid having to start from scratch. There is quite a bit of code I've had to add that I realized I shouldn't have to. In a bunch of places we were setting things up in a macro when we could move it down into the abstract classes and modules because they (should) take in the generic class it's working with (the `Avram::Model`) and we can expect that to have a lot of helpful class methods. By doing this, we remove a fair bit of code from macros which I consider a win.

### DIFF
--- a/src/avram/query.cr
+++ b/src/avram/query.cr
@@ -2,10 +2,12 @@ require "./queryable"
 
 abstract class Avram::Query
   abstract def database : Avram::Database.class
+  abstract def table_name
 
   # runs a SQL `TRUNCATE` on the current table
   def self.truncate
-    new.database.exec "TRUNCATE TABLE #{@@table_name}"
+    query = new
+    query.database.exec "TRUNCATE TABLE #{query.table_name}"
   end
 
   private def escape_sql(value : Int32)
--- a/src/avram/query.cr
+++ b/src/avram/query.cr
@@ -7,20 +7,10 @@ abstract class Avram::Query(T)
     query.database.exec "TRUNCATE TABLE #{query.table_name}"
   end
 
+  delegate :database, :table_name, :primary_key_name, to: T
+
   def schema_class
     T
-  end
-
-  def database : Avram::Database.class
-    schema_class.database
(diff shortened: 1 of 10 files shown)


src/avram/query.cr
```diff
@@ -1,11 +1,26 @@
 require "./queryable"
 
-abstract class Avram::Query
-  abstract def database : Avram::Database.class
-
+abstract class Avram::Query(T)
   # runs a SQL `TRUNCATE` on the current table
   def self.truncate
-    new.database.exec "TRUNCATE TABLE #{@@table_name}"
+    query = new
```
wontruefree, line 6: I'm curious why assign an variable here?

matthewmcgarvey: table_name is an instance method. This is the only place that needs it outside of instance methods so I didn't want to add table_name as a class method as well.

src/avram/save_operation_template.cr
```diff
@@ -19,22 +19,10 @@ class Avram::SaveOperationTemplate
         end
       {% end %}
 
-      def database
-        {{ type }}.database
```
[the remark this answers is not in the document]
matthewmcgarvey: No, because we call `.database` on the generic now, if the generic doesn't have that class method it will fail to compile.

src/avram/save_operation.cr
```diff
@@ -59,6 +51,18 @@ abstract class Avram::SaveOperation(T)
...
+    T.database
+  end
+
+  def table_name
+    T.table_name
+  end
+
+  def primary_key_name
+    T.primary_key_name
+  end
```
paulcsmith, line 64: I think this might help clarify what it is doing and take a bit less space 

```suggestion
  delegate :database, :table_name, :primary_key_name, to: T
```

paulcsmith (approved):
This is a cool refactor. looks good 👍

PR #509, opened 2020-10-30
Don't preload through association of has_many on count

### LINKED ISSUE #508: has_many through associations preload through association on select count call
https://github.com/luckyframework/avram/blob/255d4f1f59ac39ce050ab55d619ceaece9df1179/src/avram/associations/has_many.cr#L88-L95

This definitely looks like a copy/paste error but it is not necessary since all we hold on to is the count of the association.

matthewmcgarvey:
Fixes #508


jwoertink (approved):
oh yeah! 😂  that makes sense. We don't need those records since we're just returning the count.

Issue #510, opened 2020-11-03
There's no reason to have both Avram::Query and Avram::Queryable


matthewmcgarvey:
Right now the `Avram::Query` abstract class does almost nothing but `Avram::Queryable` depends on the few methods that it provides. I believe it makes sense to merge all of the functionality in it into `Avram::Queryable` and to remove the class.

paulcsmith:
I think since only internal code references this it would not be difficult to do post 1.0 so I'll remove that, but feel free to add it back if I missed something and users/external libs do depend on these classes

PR #517, opened 2020-11-10
Remove Avram::Query in favor of Avram::Queryable

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/510

Queries were extending `Avram::Query` _and_ including `Avram::Queryable` but there was no reason to do both since there wasn't a case where we used one and not the other. Since most of the logic was in `Avram::Queryable` it makes sense to move the little code in `Avram::Query` over and delete it.


wontruefree:
didn't even know this. I ❤️ this.

jwoertink (approved):
Sweet 👍 

Yeah, I think there was some talk about how external shards would hook in to Avram and using the inheritence, but this probably makes that easier if they can just `include Avram::Queryable` if needed.

Issue #487, opened 2020-10-18
Ensuring uniqueness of SQL where conditions won't work when "or" conditions are added


matthewmcgarvey:
When the only possible SQL queries were "AND" conditions, determining the uniqueness of sql conditions was possible. `SELECT * FROM users WHERE name = 'Paul' AND name = 'Paul'` is the same as `SELECT * FROM users WHERE name = 'Paul'`. But when the queries start to look like `SELECT * FROM users WHERE (name = 'Paul' AND color = 'red') OR (name = 'Paul' AND color = 'green')`, it would yield different results to remove the duplicate name conditions.

I believe this feature is most likely unused and my hope is that it would cause no issues if we removed it.

PR #518, opened 2020-11-10
Remove unique filter of where clauses

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/487

With the addition of `OR` to the query builder, where clauses can no longer be uniquely filtered because two clauses are not equal just because they have they same condition, they could be separated by an `OR` and affect the outcome if removed. For example, these two clauses are not the same but the current query builder would turn the first into the second because of the way it checks for uniqueness:
```sql
SELECT *
FROM users
WHERE name = 'Paul'
    AND age = 18
    OR name = 'Paul'
    AND age = 100;
```

```sql
SELECT *
FROM users
WHERE name = 'Paul'
    AND age = 18
    OR age = 100;
```

This will not affect any queries negatively as truly duplicate conditions are handled as expected by Postgres.

### DIFF
--- a/spec/query_builder_spec.cr
+++ b/spec/query_builder_spec.cr
@@ -1,22 +1,25 @@
 require "./spec_helper"
 
 describe Avram::QueryBuilder do
-  it "ensures uniqueness for where, orders, and joins" do
+  it "ensures uniqueness for orders, and joins" do
     query = new_query
-      .where(Avram::Where::Equal.new(:name, "Paul"))
-      .where(Avram::Where::Equal.new(:name, "Paul"))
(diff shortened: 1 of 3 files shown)


jwoertink (approved):
This looks good. I wonder if we can come up with some sort of query analyzer deal at some point to help sort of bring this back in. Since one of the main goals of Avram is to help catch bugs in development, and part of that being poorly written SQL (i.e. slow queries, duplicate where clauses, etc...), it would be nice to figure out a way to add something to help catch things like this.

For now, I don't think there's much we can do, so this is probably ok.

Issue #463, opened 2020-09-15
Associations don't respect namespaces when defining PrimaryKeyType


jwoertink:
Given:

```crystal
module MyApp
  class Post < BaseModel
    table do
      belongs_to user : User
    end
  end
end
```

If I try to define a SaveOperation for this:

```crystal
class SavePost < MyApp::Post::SaveOperation
  #... blah
end
```

I'll get the error:

```
web    |  > 32 | user_id : User::PrimaryKeyType | Nothing = Nothing.new,
web    |                     ^
web    | Error: undefined constant User::PrimaryKeyType
```

The current work-around is if your model is namespaced, then the associations must also be defined with the namespace.

```crystal
module MyApp
  class Post < BaseModel
    table do
      belongs_to user : MyApp::User
    end
  end
end
```

matthewmcgarvey:
Is it fixed if you require the user file from the post file?

jwoertink:
No, it doesn't seem to. At least not in my case it's not.

matthewmcgarvey:
@jwoertink I've made a branch of Avram with a class moved into a module and the specs all pass. Can you show me what I'm missing to cause the problem? https://github.com/matthewmcgarvey/avram/tree/namespaced-models

jwoertink:
You have to make a custom SaveOperation. Try adding a 

```crystal
class SavePrice < Billing::Price::SaveOperation
end
```

matthewmcgarvey:
I'm still not able to replicate. The branch has a custom operation but I'm sure I'm missing something

jwoertink:
I forget which app I had this issue on. I'll have to see if it's still an issue on master branch. It's possible we fixed it without knowing 🤷‍♂️

PR #519, opened 2020-11-10
Organize spec models

matthewmcgarvey:
In order to test https://github.com/luckyframework/avram/issues/463 we have to make a model in a namespace and that is difficult in Avram's specs currently because the files aren't in folders than can be required in a particular order.


jwoertink (approved):
🙏  👏  👍

PR #520, opened 2020-11-10
Refactor SchemaEnforcer validations

matthewmcgarvey:
Related to https://github.com/luckyframework/avram/issues/462

Migrates `Avram::SchemaEnforcer` validations to a common API called `Avram::SchemaEnforcer::Validation`. The goal is to move these to a list so that anyone can add to it and take advantage of the schema enforcer flow.

There might be a further refactor in the future where validations won't need any arguments because they could be made generic (`Avram::SchemaEnforcer::Validation(T)`) since we rely on class methods for everything already. Not exactly sure how that will work right now so I didn't want to move there yet.

A change that had to be made in order to support this refactor was adding a `Avram::Model.columns` method to get column info.  The constant (`Avram::Model::COLUMNS`) contains information that would be difficult to work with at runtime, for example I stringified the column type.


jwoertink (approved):
I love the concept of having a base set of validations, and allowing people to hook in to it so they can add additional validations.

Issue #501, opened 2020-10-24
Add helpful error for using belongs_to in migration


matthewmcgarvey:
Without thinking, I accidentally used `belongs_to` in a migration instead of `add_belongs_to` and got this error:

```
Showing last frame. Use --error-trace for full trace.

Unhandled exception: Negative argument (ArgumentError)
  from raise<ArgumentError>:NoReturn
  from String#*<Int32>:String
  from Crystal::TypeException+@Crystal::ErrorFormat#append_error_indicator<String::Builder, Int32, Int32, Int32>:(String::Builder | Nil)
  from Crystal::TypeException+@Crystal::ErrorFormat#error_body<Nil, (String | Nil)>:(String | Nil)
  from Crystal::TypeException+@Crystal::TypeException#append_to_s<IO::FileDescriptor, Nil>:(IO::FileDescriptor | Nil)
  from Crystal::Command#run:(Bool | Nil)
  from __crystal_main
  from main
```

I'm not able to reproduce it in Avram's migrations but, either way, we should set up a macro for `belongs_to` in migrations that recommends they switch to using `add_belongs_to`

paulcsmith:
Great idea 👍

PR #522, opened 2020-11-10
Add helpful error when using belongs_to in migration

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/501

The error looks like:

```
Error: Unexpected call to `belongs_to` in a migration.

Found in /Users/matthew/workspace/crystal/avram/db/migrations/20201110211223_bad_migration.cr:7:18.

Did you mean to use 'add_belongs_to'?

'add_belongs_to foo : Foo, ...'
```

### DIFF
--- a/src/avram/migrator/create_table_statement.cr
+++ b/src/avram/migrator/create_table_statement.cr
@@ -133,4 +133,16 @@ class Avram::Migrator::CreateTableStatement
 
     add_index :{{ foreign_key_name }}
   end
+
+  macro belongs_to(type_declaration, *args, **named_args)
+    {% raise <<-ERROR
+      Unexpected call to `belongs_to` in a migration.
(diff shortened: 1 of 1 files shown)


src/avram/migrator/create_table_statement.cr
```diff
@@ -133,4 +133,16 @@ class Avram::Migrator::CreateTableStatement
 
     add_index :{{ foreign_key_name }}
   end
+
+  macro belongs_to(type_declaration, *args, **named_args)
+    {% raise <<-ERROR
+      Unexpected call to `belongs_to` in a migration.
+      Found in #{type_declaration.filename.id}:#{type_declaration.line_number}:#{type_declaration.column_number}.
```
jwoertink, line 140: I love this! There's a ton of spots we should be using these path and line numbers 😄

jwoertink (approved):
This is a good error 👍

PR #524, opened 2020-11-11
Improve alter migrations error msg

shortly-portly:
Improve error reporting for migration where a user accidently includes the type alongside the var name for rename, rename_belongs_to, remove, remove_belongs_to. 

For example:

      remove name : String

produces the following error:

    Error:

    "remove" expected a symbol like ':user', instead got: 'name : String'.

    in: "/data/db/migrations/20170127143149_create_users.cr":19:14

    Try replacing...

      ▸ 'name : String' with 'name'

### DIFF
--- a/src/avram/migrator/alter_table_statement.cr
+++ b/src/avram/migrator/alter_table_statement.cr
@@ -175,12 +175,20 @@ class Avram::Migrator::AlterTableStatement
     @fill_existing_with_statements << "ALTER TABLE #{@table_name} ALTER COLUMN #{column} SET NOT NULL;" unless nilable
   end
 
-  {% symbol_expected_message = "%s expected a symbol like ':user', instead got: '%s'" %}
-
   macro rename(old_name, new_name)
     {% for name in {old_name, new_name} %}
       {% unless name.is_a?(SymbolLiteral) %}
-        {% raise symbol_expected_message % {"rename", name} %}
+        {% raise <<-ERROR
+
+        rename expected a symbol like ':user', instead got: '#{name}'.
+
+        in: #{name.filename}:#{name.line_number}:#{name.column_number}
+
+        Try replacing...
+
+          ▸ '#{name}' with '#{name.var}'
+        ERROR
+        %}
       {% end %}
     {% end %}
     renamed_rows << "RENAME COLUMN #{{{old_name}}} TO #{{{new_name}}}"
@@ -189,7 +197,17 @@ class Avram::Migrator::AlterTableStatement
   macro rename_belongs_to(old_association_name, new_association_name)
     {% for association_name in {old_association_name, new_association_name} %}
       {% unless association_name.is_a?(SymbolLiteral) %}
-        {% raise symbol_expected_message % {"rename_belongs_to", association_name} %}
+        {% raise <<-ERROR
+
+        rename_belongs_to expected a symbol like ':user', instead got: '#{name}'.
+
+        in: #{name.filename}:#{name.line_number}:#{name.column_number}
+
+        Try replacing...
+
+          ▸ '#{name}' with '#{name.var}'
+        ERROR
+        %}
       {% end %}
     {% end %}
     rename {{old_association_name}}_id, {{new_association_name}}_id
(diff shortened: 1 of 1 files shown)


src/avram/migrator/alter_table_statement.cr
```diff
@@ -175,12 +175,25 @@ class Avram::Migrator::AlterTableStatement
     @fill_existing_with_statements << "ALTER TABLE #{@table_name} ALTER COLUMN #{column} SET NOT NULL;" unless nilable
   end
 
-  {% symbol_expected_message = "%s expected a symbol like ':user', instead got: '%s'" %}
+  macro symbol_expected_error(action, name)
+
```
jwoertink, line 179: We may want to check if `name` is a `TypeDeclaration` here. It could be a `String` in some cases...

src/avram/migrator/alter_table_statement.cr
```diff
@@ -175,12 +175,25 @@ class Avram::Migrator::AlterTableStatement
...
+    {% raise <<-ERROR
+
+    #{action} expected a symbol like ':user', instead got: '#{name}'.
```
jwoertink, line 182: If we have a check for it being a `TypeDeclaration`, then we can say `like ':{{ name.var }}'`.

src/avram/migrator/alter_table_statement.cr
```diff
@@ -175,12 +175,25 @@ class Avram::Migrator::AlterTableStatement
...
+
+    in: #{name.filename}:#{name.line_number}:#{name.column_number}
+
+    Try replacing...
+
+      ▸ '#{name}' with '#{name.var}'
```
jwoertink, line 188:
```suggestion
      ▸ '#{name}' with ':#{name.var}'
```

jwoertink (changes requested):
I love this change! Any time we can have better error messages is always a win. Just left a few comments on it, but I think it's on the right direction!

src/avram/migrator/alter_table_statement.cr
shortly-portly: OK, so I made the change (locally)  (but also accepted your smaller changes in Github. When I try and push my changes back up I get an error because "the tip of your branch is behind its remote counterpart. I think this might be because I accepted your changes. The error also says git pull to remedy the situation but is it a git pull from Avram or from my local github copy (as you can tell I'm a bit new to all of this :-)). cheers.

src/avram/migrator/alter_table_statement.cr
shortly-portly: Hi Jeremy - done the change but see my earlier comment about Git. Cheers.

src/avram/migrator/alter_table_statement.cr
jwoertink: You should be able to just run `git pull` and be good, or `git pull origin shortly-portly:improve-alter-migrations-error-msg` may be needed. But I'd say just try `git pull` first.

shortly-portly: Thanks Jeremy - I've just got it sorted and pushed a newer version that caters for Strings as you suggested. Hopefully we are good to go but if you have any further suggestions then just shout! cheers.

src/avram/migrator/alter_table_statement.cr
```diff
@@ -175,12 +175,31 @@ class Avram::Migrator::AlterTableStatement
...
+
+    {% if name.is_a?(TypeDeclaration) %}
+      {% example = name.var %}
+    {% else %}
+      {% example = name.id %}
+    {% end %}
+
+    {% raise <<-ERROR
+
+    #{action} expected a symbol like #{example}, instead got: #{name}.
```
jwoertink, line 188:
```suggestion
    #{action} expected a symbol like :#{example}, instead got: #{name}.
```

I think just this bit and we're good!

jwoertink (commented):
Sweet! I think just the one missing colon from the example, but it's looking great! Thanks for doing this.

jwoertink (approved):
Awesome! Great work 👍

PR #525, opened 2020-11-12
Model association updates

### LINKED ISSUE #473: Naming associations something other than the model causes issues

### LINKED ISSUE #523: Update how we define has many through relationships
Related: https://github.com/luckyframework/avram/issues/496
Related: https://github.com/luckyframework/avram/issues/196

## Intro

When we define a has many through relationship, we assume that the intermediary "through" relationship has a foreign key pointing at the final table. So a `has_many customers : Customer, through: :employees` comes with the assumption that the employee model has a belongs to relationship to customers. That could be a sensible requirement. [AdonisJS's ORM](https://preview.adonisjs.com/guides/model-relations/has-many-through/) requires has many through relationships to be defined in this way.

## Problem

The problem is that we would like to support all has many through relationships in Avram and has many through relationships can also be defined where the intermediary has a has many relationship to the final table. The problem with has many through associations is that the association name given to them on the base model is not necessarily the same name as the name given to it on the intermediary model. In the example above, `has_many customers : Customer, through: :employees`, the intermediary `Employee` model could have a `has_many customers : Customer` relationship or a `belongs_to customer : Customer` relationship. The difference between the two is pluralization. The question is, though, how do we know in the base/source model which association type it is on the intermediary model? We can't inspect the associations to find out because of our current way of code generation which means there's no clear compile time way of determining how to interact with the intermediary model.

### Use a macro to pluralize/singularize the association and use `responds_to?` at runtime

This is definitely my least favorite solution and will make quite a mess in the codebase. The idea would be that we would pass around the association name of "customers" and we would singularize it into "customer" and use if blocks where we call `if @association.responds_to?(singular_name)` to determine the right methods to call.

I don't like it because of how likely we are to get the wrong name and that it creates runtime checks where we have had compile time safety.

### Remove the ability to define has many through a has many relationship

This could be an option, but I don't know how we would implement it. As I said above, we cannot inspect other models to determine at compile-time what the associations are.

### Remove the ability to define has many through relationships

While this might sound silly, Diesel, the most popular rust ORM, [only allows belongs to relationships](https://docs.diesel.rs/diesel/associations/index.html) so it is not unheard of to limit associations.

### Separate Queries from Models

This would allow us to inspect models when generating the code for queries and determine relationships more confidently. This would be a significant change to the codebase and have a lot of implications outside of this narrow issue.

### Require the associations names to be defined for the relationship

[This is how Ecto (elixir) does their has many through relationships](https://www.notion.so/Avram-Associations-Notes-ff0e880e9c5e492aaef68443e26776ba#b84cad0e1fa84ee4bdf5bcbf56c64484). Rather than declaring `has_many customers : Customer, through: :employees` and Avram has to guess on how to go from Employee to Customer, users would declare it like `has_many customers : Customer, through: [:employees, :customers]` and the array is the method list that we use to get to customers.

I definitely like this solution this best. While it would still be a breaking change, it would be a simple one and it would solve this issue entirely.

matthewmcgarvey:
Fixes #523 
Fixes #473
Fixes: #496
Related (but does not fix): #196

Had to bring in the changes in #494 and #499 in order to get it working.

## Changes

- Updates "has many through"'s to take in an array in the `through` field.
- Belongs to associations now have singular "where_xxx" queries

### Old

```crystal
class Manager < BaseModel
  table do
    has_many employees : Employee
    has_many customers : Customer, through: :employees
  end
end
```

### New

```crystal
class Manager < BaseModel
  table do
    has_many employees : Employee
    has_many customers : Customer, through: [:employees, :customers]
  end
end
```

The array is the association route to get to the customers. So Manager will need to have an association with the name of `employees` and Employee is expected to have an association called `customers` that returns Customer.

### Old

```crystal
class Employee < BaseModel
  table do
    belongs_to manager : Manager
  end
end

# notice the plural where, we were using the table name
Employee::BaseQuery.new.where_managers(...)
```

### New

```crystal
class Employee < BaseModel
  table do
    belongs_to manager : Manager
  end
end

# notice the singular where, we are using the association name now
Employee::BaseQuery.new.where_manager(...)
```

This also means that belongs to associations can now be named something other than the class and still work

```crystal
class Employee < BaseModel
  table do
    belongs_to boss : Manager
  end
end

Employee::BaseQuery.new.where_boss(...)
```

### DIFF
--- a/src/avram/associations/has_many.cr
+++ b/src/avram/associations/has_many.cr
@@ -1,10 +1,28 @@
 module Avram::Associations::HasMany
   macro has_many(type_declaration, through = nil, foreign_key = nil)
     {% if !through.is_a?(NilLiteral) && (!through.is_a?(ArrayLiteral) || through.any? { |item| !item.is_a?(SymbolLiteral) }) %}
-      {% through.raise "'through' must be given an Array of Symbols. Instead, got: #{through}" %}
+      {% through.raise <<-ERROR
+      'through' on #{@type.name} must be given an Array(Symbol). Instead, got: #{through}
+
+      Example...
+         has_many comments : Comment
+         has_many related_authors : User, through: [:comments, :author]
+
+      Learn more about associations: https://luckyframework.org/guides/database/models#model-associations
+      ERROR
+      %}
     {% end %}
     {% if !through.is_a?(NilLiteral) && through.size < 2 %}
-      {% through.raise "'through' must have at least two items. Instead, got: #{through}" %}
+      {% through.raise <<-ERROR
+      'through' on #{@type.name} must be given at least two items. Instead, got: #{through}
+
+      Example...
+         has_many comments : Comment
+         has_many related_authors : User, through: [:comments, :author]
+
+      Learn more about associations: https://luckyframework.org/guides/database/models#model-associations
+      ERROR
+      %}
     {% end %}
     {% assoc_name = type_declaration.var %}
 
--- a/src/avram/associations/has_many.cr
+++ b/src/avram/associations/has_many.cr
@@ -1,7 +1,10 @@
 module Avram::Associations::HasMany
   macro has_many(type_declaration, through = nil, foreign_key = nil)
-    {% if !through.is_a?(NilLiteral) && !through.is_a?(SymbolLiteral) %}
-      {% through.raise "The association name for 'through' must be a Symbol. Instead, got: #{through}" %}
+    {% if !through.is_a?(NilLiteral) && (!through.is_a?(ArrayLiteral) || through.any? { |item| !item.is_a?(SymbolLiteral) }) %}
+      {% through.raise "'through' must be given an Array of Symbols. Instead, got: #{through}" %}
+    {% end %}
+    {% if !through.is_a?(NilLiteral) && through.size < 2 %}
+      {% through.raise "'through' must have at least two items. Instead, got: #{through}" %}
     {% end %}
     {% assoc_name = type_declaration.var %}
 
@@ -21,7 +24,7 @@ module Avram::Associations::HasMany
     {% model = type_declaration.type %}
 
     define_has_many_lazy_loading({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{ through }})
-    define_has_many_base_query({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{through}})
+    define_has_many_base_query({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{ through }})
   end
 
   private macro define_has_many_base_query(assoc_name, model, foreign_key, through)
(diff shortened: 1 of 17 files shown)


src/avram/associations/has_many.cr
```diff
@@ -69,6 +61,31 @@ module Avram::Associations::HasMany
...
+
+      def preload_{{ assoc_name }}
+        modified_query = yield {{ model }}::BaseQuery.new
+        preload_{{ assoc_name }}(modified_query)
+      end
+
+      def preload_{{ assoc_name }}(preload_query : {{ model }}::BaseQuery)
+        preload_{{ through.first.id }} do |through_query|
+          through_query.preload_{{ through[1].id }}(preload_query)
+        end
```
matthewmcgarvey, line 78: We force the preload of the through association and in the query we tell that association to preload the next one.

This is reversing the way we were doing it before. This is required because we don't know the method names in the old direction, we were making a guess using the foreign key which doesn't handle both the has many through a belongs to and a has many through a has many scenarios.

I _think_ this ends up making more sql calls and could potentially be N+1 but I'm not sure.

src/avram/associations/has_many.cr
```diff
@@ -85,13 +102,10 @@ module Avram::Associations::HasMany
...
-          .new
-          .join_{{ through.id }}
-          .__yield_where_{{ through.id }} do |through_query|
-            through_query.{{ foreign_key.id }}(id)
-          end
-          .select_count
+        {{ through.first.id }}_query
+          .{{ foreign_key.id }}(id)
+          .map(&.{{ through[1].id }}_count)
+          .sum
```
matthewmcgarvey, line 108: This is definitely making more sql calls than before. Like I said above, we can't start at the opposite end of the association, we have to start from base and travel through the "throughs"/intermediary tables because those are the only names we know.

This is why I had to add the `def {{...}}_count` method to belongs to associations.

src/avram/model.cr
```diff
@@ -223,6 +224,14 @@ abstract class Avram::Model
     })
   end
 
+  macro setup_association_queries(associations, *args, **named_args)
+    {% for assoc in associations %}
+      def {{ assoc[:assoc_name] }}_query
+        {{ assoc[:type] }}::BaseQuery.new
+      end
+    {% end %}
+  end
```
matthewmcgarvey, line 233: This is necessary because we don't have access to the types of the intermediary associations in a has many through

src/avram/associations/has_many.cr
```diff
@@ -1,7 +1,10 @@
 module Avram::Associations::HasMany
   macro has_many(type_declaration, through = nil, foreign_key = nil)
-    {% if !through.is_a?(NilLiteral) && !through.is_a?(SymbolLiteral) %}
-      {% through.raise "The association name for 'through' must be a Symbol. Instead, got: #{through}" %}
+    {% if !through.is_a?(NilLiteral) && (!through.is_a?(ArrayLiteral) || through.any? { |item| !item.is_a?(SymbolLiteral) }) %}
+      {% through.raise "'through' must be given an Array of Symbols. Instead, got: #{through}" %}
```
jwoertink, line 4: What about an error like this?

```crystal
      {% through.raise <<-ERROR
      'through' on #{@type.name} must be given an Array of Symbols. Instead, got: #{through}
      
      Example...
         has_many comments : Comment
         has_many related_authors : User, through: [:comments, :author]
      
      Learn more about associations: https://luckyframework.org/guides/database/models#model-associations
      ERROR
      %}
```

We could also see if it's possible to put the exact file and such like we've done in some other errors. I'm just thinking "what goes in this array?" type deals.

src/avram/associations/has_many.cr
```diff
@@ -1,7 +1,10 @@
...
+    {% end %}
+    {% if !through.is_a?(NilLiteral) && through.size < 2 %}
+      {% through.raise "'through' must have at least two items. Instead, got: #{through}" %}
```
jwoertink, line 7: Same?

src/avram/associations/has_many.cr
```diff
@@ -1,7 +1,10 @@
 module Avram::Associations::HasMany
   macro has_many(type_declaration, through = nil, foreign_key = nil)
-    {% if !through.is_a?(NilLiteral) && !through.is_a?(SymbolLiteral) %}
-      {% through.raise "The association name for 'through' must be a Symbol. Instead, got: #{through}" %}
+    {% if !through.is_a?(NilLiteral) && (!through.is_a?(ArrayLiteral) || through.any? { |item| !item.is_a?(SymbolLiteral) }) %}
+      {% through.raise "'through' must be given an Array of Symbols. Instead, got: #{through}" %}
+    {% end %}
+    {% if !through.is_a?(NilLiteral) && through.size < 2 %}
```
jwoertink, line 6: Do you expect this to work with more than 2? I tried a 3 level deep, but got errors.

```crystal
has_many things : Thing, through: [:first, :second, :things]
```

If that's not the case, then we should check that this is always exactly 2, and then make it a `Tuple(Symbol, Symbol)`. If we can allow more than 2, then we will need a whole test for that.

matthewmcgarvey: right now it only handles two, but I plan to add multiple layers. I don't want to have multiple breaking changes though

jwoertink: Ah ok. That's fine then

wontruefree:
Is the destination name of the through association required or optional now?

matthewmcgarvey:
@wontruefree I'm not sure I understand the question. Could you explain what you mean?

jwoertink (approved):
Ok, I gave this another shot after the update, and it's working. I can run specs back to back again, and my tests all passed. I think we should just merge this and get it in the hands of some people to really play around with it. So feel free to merge when you feel ready 👍

paulcsmith:
This is awesome. Thanks @matthewmcgarvey

PR #526, opened 2020-11-12
Update association property to assoc_name instead of table_name

matthewmcgarvey:
Breaking out changes from https://github.com/luckyframework/avram/pull/525 to reduce its size.

## Why make this change?

In all places except belongs_to, we are using the association name instead of trying to infer the table name. It isn't necessary to make this inference though. Anywhere we access the table name from the association, we can use the constant `::TABLE_NAME` or the class method `.table_name`. The belongs to macro is still inferring the table name for now because replacing it with the association name takes it from plural to singular and that leads to all the changes made in the PR this is coming from.

### DIFF
--- a/spec/support/models/comment.cr
+++ b/spec/support/models/comment.cr
@@ -16,9 +16,7 @@ class CommentForCustomPost < BaseModel
     primary_key custom_id : Int64
     timestamps
     column body : String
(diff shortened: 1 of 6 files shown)


jwoertink (approved):
👍

Issue #521, opened 2020-11-10
Catch duplicate migration names


jwoertink:
Right now you can do 

```
lucky gen.migration CreateUsers
lucky gen.migration CreateUsers
```

and this will generate 2 separate migrations called `_create_users.cr` but with different timestamps. We should raise an error on the second one saying that you've already done this. We currently do the same for generating models. 

Ref: https://discord.com/channels/743896265057632256/743896265057632259/775771902979080204

matthewmcgarvey:
It could also be copy/paste errors, I wonder if there's a way we could detect duplicate classes. I suspect we'd only be able to check file names?

jwoertink:
Yeah, the filenames minus the timestamps would probably be fine. I know I've done it in the past with rails where I've done "rails g migration add_files_to_users" then like 3 years later ran the same thing...  It's laziness, but maybe we can help catch that to force you to be a little more explicit.

wontruefree:
I think raising a warning might be more appropriate unless there could be unexpected behavior.

jwoertink:
This is fixed in #528

PR #528, opened 2020-11-12
Fix duplicate migration names

shortly-portly:
lucky gen.migrations now returns an error if the migration name already exists.

Fixes issues: https://github.com/luckyframework/avram/issues/521

Error message is:

        Migration name must be unique

         Migration name: repeat_migration.cr already exists as: 20201112200313_repeat_migration.cr.

Also added tests for the above and for camel case.


jwoertink (approved):
Love it!

matthewmcgarvey (approved):

PR #532, opened 2020-11-13
Some simple cleanup on the path to #369

jwoertink:
This just cleans up the operation error stuff.

Ref: https://github.com/luckyframework/avram/pull/469#discussion_r509677459

### DIFF
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -8,13 +8,13 @@ require "./param_key_override"
 require "./inherit_column_attributes"
 require "./validations"
 require "./define_attribute"
-require "./save_operation_errors"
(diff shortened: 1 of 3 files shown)


src/avram/save_operation.cr
```diff
@@ -228,11 +228,6 @@ abstract class Avram::SaveOperation(T)
     save_status == SaveStatus::SaveFailed
   end
 
-  # :nodoc:
-  macro fillable(*args)
```
jwoertink, line 232: This hasn't been a thing for a while, and I don't think it's a common enough term for people to accidentally use.

matthewmcgarvey (approved):

PR #534, opened 2020-11-13
Custom errors and valid property on Operations

### LINKED ISSUE #529: Allow for custom errors on operations
I want to be able to do my own custom validation and error against a non column.

```crystal
class SaveMessage < Message::SaveOperation
  needs author : User
  permit_columns text

  before do
    if author.banned?
      add_error(:author_banned, "You've been banned from posting messages.")
    end 
   end
end
```

jwoertink:
Fixes #529 

This allows you to set custom errors not related to attributes. When you set an error message, the whole operation is no longer valid.

The main use case is when your operations (or save operations) need some external object, but that object itself may also not be valid.

```crystal
class SaveMessage < Message::SaveOperation
  needs user : User

  before_save do
    if user.is_not_allowed_to_save_message?
       add_error(:user_not_welcome, "You don messed up A A ron.")
    end
  end
end

SaveMessage.create(user: current_user) do |op, _|
  op.errors[:user_not_welcome] == ["You don messed up A A ron"]
end
```

### DIFF
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -37,7 +37,6 @@ abstract class Avram::SaveOperation(T)
   @params : Avram::Paramable
   getter :record, :params
   property save_status : SaveStatus = SaveStatus::Unperformed
-  property valid : Bool = true
 
   abstract def attributes
 
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -37,6 +37,7 @@ abstract class Avram::SaveOperation(T)
   @params : Avram::Paramable
   getter :record, :params
   property save_status : SaveStatus = SaveStatus::Unperformed
+  property valid : Bool = true
 
   abstract def attributes
 
@@ -210,7 +211,9 @@ abstract class Avram::SaveOperation(T)
 
   # Runs required validation,
   # then returns `true` if all attributes are valid.
+  # Set `valid` to false to force invalid state.
   def valid? : Bool
+    return false unless valid
     # These validations must be ran after all `before_save` callbacks have completed
     # in the case that someone has set a required field in a `before_save`. If we run
     # this in a `before_save` ourselves, the ordering would cause this to be ran first.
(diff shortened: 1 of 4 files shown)


src/avram/save_operation.cr
```diff
@@ -210,7 +211,9 @@ abstract class Avram::SaveOperation(T)
 
   # Runs required validation,
   # then returns `true` if all attributes are valid.
+  # Set `valid` to false to force invalid state.
   def valid? : Bool
+    return false unless valid
```
jwoertink, line 216: don't even waste time running required validations since those may run some queries (i.e. uniqs and such). We are saying it's not valid so the others don't really matter.

matthewmcgarvey:
@jwoertink Calling `add_error` does not make the operation invalid, is that intentional?

This is something that has tripped me up in rails before, and this PR implements the same thing.

jwoertink:
hmm 🤔  I guess I assumed you'd never call `add_error` unless it was invalid... BUT.... you do make a good point. I think the ruby mutations shard does that where you just add an error and the whole thing fails. Do you think that would be a better way? So like, instead of setting `self.valid = false`, you just add an error message, and it knows it's invalid at that point.... I'm thinking maybe that is a better UX.

```crystal
def valid?
  return false if any_errors_detected?

   #... other stuff it does
end
```

matthewmcgarvey:
maybe the `valid?` check would just see if any errors are present?

jwoertink:
Yup, exactly. Ok. I'll make that change and push it up.

jwoertink:
Alright, so this doesn't really fix my original thought behind #529, but it does allow me to do some of what I intended by setting a custom error on operations.

I'm thinking maybe in a future PR we could have some way to "halt" the entire chain of callbacks and such. Something that says "don't run any more callbacks, don't run any other validations... nothing... just stop and exist the operation". For now, this will be a good improvement.

matthewmcgarvey (approved):

PR #536, opened 2020-11-15
Allow unique index on belongs to

shortly-portly:
Allow for:
 add_belongs_to author : User, unique: true

Updated tests.


matthewmcgarvey (approved):
Nice 👍

PR #538, opened 2020-11-18
Remove table_name from Avram::Model setup macro

matthewmcgarvey:
No connected issue, refactor only.

Inlined the only usage of `table_name` and removed it from the `setup` macro.


jwoertink (approved):
👍

Issue #539, opened 2020-11-18
Insert only sends changed values


matthewmcgarvey:
https://github.com/luckyframework/avram/blob/c5443e3c8101c7172454d2b43fb451979ab012f4/src/avram/save_operation.cr#L392-L394

https://github.com/luckyframework/avram/blob/c5443e3c8101c7172454d2b43fb451979ab012f4/src/avram/save_operation.cr#L279-L287

We shouldn't be using `SaveOperation#changes` to get the values for an insert. It filters out attributes that have the same value as before. An insert can have attributes filtered when they have the same value as the default.

# Reproducing the error

Create a table that has a non-nullable column and does not supply a default value. Create a model to write to that table and provide a default value to that column in the model. Either with a Box or a SaveOperation attempt to create a model by supplying it with the same value as the default. It will fail with error like:

```
null value in column "sales" violates not-null constraint (PQ::PQError)
         from lib/pg/src/pq/connection.cr:203:7 in 'handle_error'
         from lib/pg/src/pq/connection.cr:184:9 in 'handle_async_frames'
         from lib/pg/src/pq/connection.cr:162:7 in 'read'
         from lib/pg/src/pq/connection.cr:388:18 in 'expect_frame'
         from lib/pg/src/pq/connection.cr:372:9 in 'read_next_row_start'
         from lib/pg/src/pg/result_set.cr:39:8 in 'move_next'
         from lib/db/src/db/result_set.cr:39:13 in 'from_rs'
         from src/avram/save_operation.cr:374:17 in 'insert'
         from src/avram/save_operation.cr:356:7 in 'insert_or_update'
         from src/avram/save_operation.cr:308:9 in 'save'
         from src/avram/save_operation.cr:337:8 in 'save!'
         from spec/support/models/company.cr:2:3 in 'create!:sales:earnings'
         from spec/operations/save_operation_spec.cr:425:9 in '->'
         from ../../../.asdf/installs/crystal/0.35.1/src/primitives.cr:255:3 in 'internal_run'
         from ../../../.asdf/installs/crystal/0.35.1/src/spec/example.cr:33:16 in 'run'
         from ../../../.asdf/installs/crystal/0.35.1/src/spec/context.cr:18:23 in 'internal_run'
         from ../../../.asdf/installs/crystal/0.35.1/src/spec/context.cr:330:7 in 'run'
         from ../../../.asdf/installs/crystal/0.35.1/src/spec/context.cr:18:23 in 'internal_run'
[...]

PR #540, opened 2020-11-18
Allow insert giving default value

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/539

Does not filter out unchanged attributes, only nils, on inserting. This allows creating a model with the same value as the default.

### DIFF
--- a/spec/operations/save_operation_spec.cr
+++ b/spec/operations/save_operation_spec.cr
@@ -421,6 +421,14 @@ describe "Avram::SaveOperation" do
         end
       end
 
+      it "allows supplying a value that matches the default" do
+        result = Company::SaveOperation.create!(sales: 0_i64, earnings: 0_i64)
+        default_result = Company::SaveOperation.create!
+
+        result.sales.should eq default_result.sales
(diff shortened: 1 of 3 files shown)


jwoertink (approved):
Nice!

spec/operations/save_operation_spec.cr
```diff
@@ -421,6 +421,14 @@ describe "Avram::SaveOperation" do
         end
       end
 
+      it "allows supplying a value that matches the default" do
+        result = Company::SaveOperation.create!(sales: 0_i64, earnings: 0_i64)
```
wontruefree, line 425: shouldnt this be a float?

```
result = Company::SaveOperation.create!(sales: 0_i64, earnings: 0_f64)
```

matthewmcgarvey: Oya, I guess you're right. The type extensions have methods for doing the conversation, luckily!

wontruefree: TIL.  I am surprised that worked but it makes sense.

PR #541, opened 2020-11-20
Only truncate expected database tables

matthewmcgarvey:
We were truncating all table names in DatabaseInfo. This included:
- The migrations table
- Non-tables like views

This bug showed itself while running specs in generated Lucky apps because the database is truncated after every test.

We did not see this in Avram because the specs did not load the database migrations which it now does.

The bug started in https://github.com/luckyframework/avram/commit/361a9e84d4b1d3cf35346f6eb361c9388bcaccf0


jwoertink (approved):
Cool 👍

Issue #531, opened 2020-11-12
Issues with preloads


jwoertink:
I'm building an app against master, and I'm running in to a lot of issues with preloading. Way more than I've ran into before, so something must have changed recently, but I'm not sure what. I'm currently unable to use any method generated from an association because they all require preloading. The issue is that once these methods are created, you no longer have access to the underlying query to be able to preload them.

This is the most recent one I got:

```crystal
class MessageSerializer < BaseSerializer
  def initialize(@message : Message)
  end

  def render
    {
      text: @message.text,
      message_id: @message.id,
      username: @message.user.username
    }
  end
end
```

> user for Message must be preloaded with 'preload_user'

In this case `Message` `belongs_to user : User`, and `User` `has_many messages : Message`. Just calling `message.user` throws the error. 

To get around this, I have to now create a `user` method in my `Message` model that will build out the query with the preload.

```crystal
class Message < BaseModel
  table do
    column text : String
    column user_id : Int64
  end

  def user
    UserQuery.find(user_id)
  end
end
```

By doing this, it makes the `belongs_to` useless. I'm having this same issue when it comes to the `has_many`. 

I should also note that the `message` object I got comes back from a callback in the `SaveOperation` so I don't have access to the original query that created it.

```crystal
class SaveMessage < Message::SaveOperation

  after_save do |message|
      # I want to ship this data off somewhere
      # but can't because of the preload error
     MessageSerializer.new(message)
  end
end
```

matthewmcgarvey:
I've been referencing Elixir's Ecto library a bit lately and they have a way to add associations to an object after it has been loaded. You call `Repo.preload message, :user` to add the user association to the message object. I think we should be able to do something like that with our query objects.

I support the explicit-ness of Avram and I think the user method you added to Message would hide the database operation that is happening.

https://hexdocs.pm/ecto/Ecto.Repo.html#c:preload/3

jwoertink:
I actually totally forgot that I can skip the preload with `message.user!`, but we should definitely make this a lot easier to deal with. No error message pointed me in that direction, and I only remembered it was a thing about an hour ago 😅

PR #542, opened 2020-11-24
Add belongs_to preload methods for existing records

matthewmcgarvey:
Part 1 of https://github.com/luckyframework/avram/issues/531

Provides a way to preload associations onto existing records. This is especially useful when you have helper methods for your actions to load certain objects from the database but don't have associations preloaded for an association needed in a particular action. You could either update that helper method to preload the association you need, but that could be overkill for your other actions. You could also make a separate query for the association records but they won't be connected to the other record so you have to pass around both where they are needed.

## Usage

```crystal
user.organization # LazyLoadError
user = UserQuery.preload_organization(user)
user.organization # Success!
```

### DIFF
--- a/src/avram/associations/belongs_to.cr
+++ b/src/avram/associations/belongs_to.cr
@@ -55,13 +55,11 @@ module Avram::Associations::BelongsTo
       def self.preload_{{ assoc_name }}(records : Enumerable, preload_query = {{ model }}::BaseQuery.new)
         ids = records.compact_map(&.{{ foreign_key }})
         empty_results = {} of {{ model }}::PrimaryKeyType => Array({{ model }})
-        {{ assoc_name }} = ids.empty? ? empty_results  : preload_query.dup.id.in(ids).results.group_by(&.id)
+        {{ assoc_name }} = ids.empty? ? empty_results  : preload_query.id.in(ids).results.group_by(&.id)
         records.each do |record|
-          if (id = record.{{ foreign_key }})
-            record.__set_preloaded_{{ assoc_name }} {{ assoc_name }}[id]?.try(&.first?)
-          else
-            record.__set_preloaded_{{ assoc_name }} nil
-          end
+          id = record.{{ foreign_key }}
+          assoc = id.nil? ? nil : {{ assoc_name }}[id]?.try(&.first?)
+          record.__set_preloaded_{{ assoc_name }}(assoc)
         end
         records
       end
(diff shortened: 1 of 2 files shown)


src/avram/associations/belongs_to.cr
```diff
@@ -48,6 +48,24 @@ module Avram::Associations::BelongsTo
 
   private macro define_belongs_to_base_query(assoc_name, model, foreign_key)
     class BaseQuery
+      def self.preload_{{ assoc_name }}(record, preload_query = {{ model }}::BaseQuery.new)
+        preload_{{ assoc_name }}(records: [record], preload_query: preload_query).first
+      end
+
+      def self.preload_{{ assoc_name }}(records : Enumerable, preload_query = {{ model }}::BaseQuery.new)
+        ids = records.map(&.{{ foreign_key }}).compact
```
wontruefree, line 56: Would `compact_map` work here?
```suggestion
        ids = records.compact_map(&.{{ foreign_key }})
```

src/avram/associations/belongs_to.cr
```diff
@@ -48,6 +48,24 @@ module Avram::Associations::BelongsTo
 
   private macro define_belongs_to_base_query(assoc_name, model, foreign_key)
     class BaseQuery
+      def self.preload_{{ assoc_name }}(record, preload_query = {{ model }}::BaseQuery.new)
+        preload_{{ assoc_name }}(records: [record], preload_query: preload_query).first
+      end
+
+      def self.preload_{{ assoc_name }}(records : Enumerable, preload_query = {{ model }}::BaseQuery.new)
+        ids = records.compact_map(&.{{ foreign_key }})
```
jwoertink, line 56: I like this refactor 👍

jwoertink (commented):
This is pretty awesome, and I think it'll help a lot. Are you thinking about adding this in to the `has_many` and `has_one` as well? Or just keep it on `belongs_to` for now?

matthewmcgarvey:
@jwoertink SInce this was the first go at it, I wanted to keep it simple so I'll add follow-up prs for the other association types after this one is merged in.

src/avram/associations/belongs_to.cr
```diff
@@ -48,6 +48,24 @@ module Avram::Associations::BelongsTo
 
   private macro define_belongs_to_base_query(assoc_name, model, foreign_key)
     class BaseQuery
+      def self.preload_{{ assoc_name }}(record, preload_query = {{ model }}::BaseQuery.new)
+        preload_{{ assoc_name }}(records: [record], preload_query: preload_query).first
+      end
+
+      def self.preload_{{ assoc_name }}(records : Enumerable, preload_query = {{ model }}::BaseQuery.new)
```
matthewmcgarvey, line 55: Should this modify the passed in records or return new records? Right now the passed in records are being mutated, but I'm not sure if we have a way of getting an exact copy.

wontruefree: It seems to me mutating records would be more performant.  In general Avam seems to have moved to duping database interaction for safety.  I think updating the `records` themselves might mitigate some unexpected behavior.

wontruefree: here is an example of non mutating refactor.

```crystal
results = {} of {{ model }}::PrimaryKeyType => Array({{ model }})
ids = records.compact_map(&.{{ foreign_key }})

results if ids.empty?
{{ assoc_name }} = preload_query.dup.id.in(ids).results.group_by(&.id)

records.each do |record|
  if (id = record.{{ foreign_key }})
    rec = record.dup
    rec.__set_preloaded_{{ assoc_name }} {{ assoc_name }}[id]?.try(&.first?)
    results << rec
  else
    rec = record.dup
    rec.__set_preloaded_{{ assoc_name }} nil
    results << rec
  end
end

results
```

matthewmcgarvey: Unfortunately I had to split the implementations between the class level and the instance level because the instance level relies on the mutations but I decided to go down this route. Thanks for looking into this 👍

jwoertink (approved):
👍

PR #544, opened 2020-11-25
Adding in a new after_completed callback

### LINKED ISSUE #432: Callbacks do not run if no column attribute changed

jwoertink:
Fixes #432 

Also see comment: https://github.com/luckyframework/avram/pull/481#issuecomment-713881686

This is a new callback added for SaveOperation that will run after commits, but this also runs even if no changes are happening to the record. As where `after_commit` only runs if an actual DB commit happens, `after_completed` will run even when there's no changes to be made.

Naming is hard, so I'm down for a different name. I also thought about calling it `after_finalized` or `after_succesful_run`.

Another option I toyed with was doing `after_validated`; however, I felt that a name like that would imply that it would run after the `valid?` calls... This would allow us to call in both spots, but the original issue for #432 wanted to run things after stuff was saved, and that would not be the case here 🤔

### DIFF
--- a/spec/operations/save_operation_callbacks_spec.cr
+++ b/spec/operations/save_operation_callbacks_spec.cr
@@ -157,6 +157,19 @@ private class UpdateOperationWithSkipCallbacks < SaveOperationWithCallbacks
   end
 end
 
+private class UpdateOperationWithNoUpdates < Post::SaveOperation
+  include TestableOperation
+
+  after_completed :mark_operation_completed
+  after_completed do |_saved_post|
+    mark_callback("after_completed_in_a_block")
+  end
+
+  private def mark_operation_completed(_saved_post)
+    mark_callback("after_completed_called")
+  end
+end
+
 describe "Avram::SaveOperation callbacks" do
   it "does not run any callbacks if just validating" do
     operation = CallbacksSaveOperation.new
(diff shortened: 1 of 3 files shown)


matthewmcgarvey (approved):
👍 nice

PR #545, opened 2020-11-25
Refactor SchemaEnforcer validations into separate files

matthewmcgarvey:
I'm looking into adding another validation but the SchemaEnforcer file was feeling bloated so I moved the validations into their own files and refactored the specs as well.


jwoertink (approved):
Ah nice! Those specs are easier to read.

PR #546, opened 2020-11-25
UUID schema enforcer

matthewmcgarvey:
Related to https://github.com/luckyframework/avram/issues/393

If we ever wanted to support generating UUID primary keys from the database, this validation will aid in catching tables that need to be updated to set a default value. I did **not** add it to the list of validations that run by default.


jwoertink (approved):
I love this!

PR #548, opened 2020-12-01
Preloading specs refactor

matthewmcgarvey:
Related to https://github.com/luckyframework/avram/pull/542

 Some of the spec files are getting bloated because they have specs for each type of association. It makes it hard to tell what is tested and where to add specs. Breaking the large spec files into individual files for each association type makes them slimmer and easier to digest. I am updating the preloading specs first because that is where I will be working.


spec/support/query_spy.cr
```diff
@@ -0,0 +1,12 @@
+# NOTE: This is only for testing if this is called during a query
+# TODO: Remove once a proper mocking shard is built
```
jwoertink, line 2: Nice. I think we have a few other random "private" classes that do stuff like this that we could probably make a whole directory for this sort of stuff.

jwoertink (approved):
Much nicer!

PR #549, opened 2020-12-01
Remove primary key name and type from Model setup macro

matthewmcgarvey:
Related to: https://github.com/luckyframework/avram/issues/6
Related to: https://github.com/luckyframework/avram/issues/453

Another step on the way on breaking the requirements on having primary keys. If you did not provide a primary key in the model, it immediately failed when trying to run the `setup` macro. Now, anywhere that uses the primary key pulls it through the `Crystal::Macros::TypeNode#constant` which returns `NilLiteral` if the constant is not found.

### DIFF
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -1,5 +1,5 @@
 class Avram::BaseQueryTemplate
-  macro setup(type, columns, associations, primary_key_name, *args, **named_args)
+  macro setup(type, columns, associations, *args, **named_args)
     class ::{{ type }}::BaseQuery
       private class Nothing
       end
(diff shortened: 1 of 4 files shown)


jwoertink (commented):
I've been thinking about this a bit... Do you think by allowing models to not require a primary key, we lose out on some of the safety? What if we just said you can't have a table without a primary key? Existing tables that don't have them could run a migration to fill it with bunk data. Then for views, we create a separate class, or make a separate `view` macro instead of `table` which would circumvent calling the validate_primary_key? I'd hate to run in to an issue where you end up doing `thing.id.not_nil!` 🤔  

This change in itself seems ok since it's just a bit of refactor, but it just makes me a little nervous to start going down this route. What do you think?

matthewmcgarvey:
@jwoertink I'm fine with Avram requiring regular tables to have a primary key if we think that is best. In fact, there's still an explicit validation in the model that raises an error if no primary key is specified https://github.com/luckyframework/avram/blob/2788dff63119b14f64a010c1c2e67072685fa66e/src/avram/model.cr#L98

I think we will have a separate set up for views, but I think we would still call a lot of the same set up methods that the table macro calls which is why this change is necessary.

jwoertink (approved):
Ok. I think we should for sure just require a primary key on all tables. If you have an existing table that doesn't have one, then you just handle adding it in or whatever. We can document that. Then for views, we just don't call `table`, and they will work a little different. I'm gonna close out that other issue and make note there too. 👍

PR #551, opened 2020-12-03
Remove Avram::Model#to_param

matthewmcgarvey:

This is an undocumented method on models that is yet another hurdle to jump when adding supports for database views. This also doesn't provide any functionality over calling `#id` on the model directly. If anyone was using it, it would take two seconds to add it back to the models they use it on.


matthewmcgarvey:
Rolled into https://github.com/luckyframework/avram/pull/552

PR #552, opened 2020-12-03
Clean up Avram::Model to prepare for database views

matthewmcgarvey:

So close to being able to support database views which might not have primary keys! This PR updates the base query template to conditionally add the `Avram::PrimaryKeyQueryable` model and extracts a similar `Avram::PrimaryKeyMethods` module for methods in `Avram::Module` that depends on the primary key beings defined. It also moves the `Avram::Model.primary_key_name` to be nilable. `Avram::SchemaEnforcer::EnsureUUIDPrimaryKeyHasDefault` also needed to be updated to ignore models that are missing the primary key.

`Avram::Model#to_param` was also removed. This is an undocumented method on models that is yet another hurdle to jump when adding supports for database views. This also doesn't provide any functionality over calling `#id` on the model directly. If anyone was using it, it would take two seconds to add it back to the models they use it on.


src/avram/model.cr
```diff
@@ -8,82 +8,27 @@ abstract class Avram::Model
   include Avram::Polymorphic
   include Avram::SchemaEnforcer
 
-  abstract def id
```
jwoertink, line 11: Do you think we should move this to the `table` macro? Then if you call `table`, you have to have the `id` method defined.

src/avram/model.cr
```diff
@@ -8,82 +8,27 @@ abstract class Avram::Model
...
-  #
-  # Instead re-assign the variable. Now 'name' will return "Updated" since
-  # 'user' references the reloaded model.
-  # user = user.reload
-  # user.name
-  # ```
+  # Refer to `PrimaryKeyMethods#reload`
   def reload : self
-    base_query_class.find(id)
+    raise Avram::UnsupportedQueryError.new("Unable to call Avram::Model#reload on #{self.class.name} because no primary key has been defined.")
```
jwoertink, line 26: What do you think about making this a compile-time error? I think for it to work, the definitions in `PrimaryKeyMethods` need to be inside of the macro `included` 🤔 

```crystal
{% raise "Unable to call Avram::Model#reload on {{ @type.id }} because no primary key has been defined." %}
```

jwoertink (commented):
This is sweet that it's getting closer! A few small comments, but I think it's good 👍

src/avram/model.cr
matthewmcgarvey: At that point it isn't an expected method on `Avram::Model` it's just a method we define for models connected to tables which is what we already do. I don't think it's valuable to keep

jwoertink: ok 👍

src/avram/model.cr
matthewmcgarvey: Great idea! I had no idea you could do this. Now if only I could get my brain to remember it 🤦‍♂️🤦‍♂️🤦‍♂️

matthewmcgarvey:
@jwoertink is this ready to go?

jwoertink (approved):
I think this is good now 👍

PR #554, opened 2020-12-04
Revert "Fix issue with blank strings and type extensions (#381)"

matthewmcgarvey:
This reverts commit 2788dff63119b14f64a010c1c2e67072685fa66e.

See https://github.com/luckyframework/avram/pull/381#issuecomment-738443183 as to why


src/avram/where.cr
```diff
@@ -38,7 +38,7 @@ module Avram::Where
   end
 
   abstract class ValueHoldingSqlClause < SqlClause
-    getter value : String | Array(String) | Array(Int32) | Nil
```
jwoertink, line 41: All because of this... I'm still not sure why this messes with Lucky Params... 😝

jwoertink (approved):
Thanks for doing this. I think we still need it, but we may need to come up with a different implementation.

jwoertink:
Hey @bcardiff if you get the chance, this should revert that change that broke your tests. Hopefully they all compile and run 😬

bcardiff:
test-ecosystem is as happy as it can be regarding lucky 🎉  https://github.com/bcardiff/test-ecosystem/runs/1500476107?check_suite_focus=true . Thanks!

jwoertink:
Sweet! That's good to know. We are looking in to how we can better catch stuff like this.

PR #555, opened 2020-12-04
Add support for database views

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/453

To declare a view, instead of using the `table` macro, there is now a `view` macro. There are a few differences between the two:

- There are no default columns. If your view has a primary key or timestamps, they must be defined explicitly
- No `Model::SaveOperation` is defined
- If the view does not have a primary key, some methods don't work or will be missing
  - `Avram::Model#reload` raises a compilation error if invoked without a primary key
  - `Model::BaseQuery` no longer includes `Avram::PrimaryKeyQueryable` which has methods like `#find` that require a primary key

No functionality around materialized views is included with this change.

### DIFF
--- a/db/migrations/20201202225520_create_admin_users_view.cr
+++ b/db/migrations/20201202225520_create_admin_users_view.cr
@@ -1,0 +1,14 @@
+class CreateAdminUsersView::V20201202225520 < Avram::Migrator::Migration::V1
+  def migrate
+    execute <<-SQL
+      CREATE VIEW admin_users AS
+        SELECT users.*
+        FROM users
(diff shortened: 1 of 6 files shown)


matthewmcgarvey:
🚫 SchemaEnforcer fails. All columns for views are nilable. We'll have to be able to configure that check

matthewmcgarvey:
This requires https://github.com/luckyframework/avram/pull/556 to be merged first

jwoertink (approved):
I think the main thing will be to see how this works inside Lucky itself. But lets get it out there and give it a spin!

PR #556, opened 2020-12-04
Schema enforcer update

matthewmcgarvey:
Related to https://github.com/luckyframework/avram/pull/555

In order for views to work with SchemaEnforcer, `EnsureMatchingColumns` has to have a configuration to not check required vs. nilable on columns since all columns are nilable in views. This update also changes how we set up validations for SchemaEnforcer so that tables _could_ have different validations than views and there's a way to add/remove/customize validations

### DIFF
--- a/src/avram/schema_enforcer/validation.cr
+++ b/src/avram/schema_enforcer/validation.cr
@@ -1,14 +1,20 @@
 abstract class Avram::SchemaEnforcer::Validation
   private getter model_class : Avram::Model.class
-  private getter database_info : Avram::Database::DatabaseInfo
 
   def initialize(@model_class)
-    @database_info = @model_class.database.database_info
   end
 
   abstract def validate!
 
   private def table_name
     model_class.table_name.to_s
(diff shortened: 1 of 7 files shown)


src/avram/schema_enforcer/validation.cr
```diff
@@ -1,14 +1,20 @@
 abstract class Avram::SchemaEnforcer::Validation
   private getter model_class : Avram::Model.class
-  private getter database_info : Avram::Database::DatabaseInfo
 
   def initialize(@model_class)
-    @database_info = @model_class.database.database_info
```
matthewmcgarvey, line 6: This could no longer live here because `#database_info` causes a database query which must happen after the database has been configured. Moving it into the class level of models meant that it is no longer the case.

jwoertink (approved):
I've tried to review this PR 4 times now, and I keep getting side tracked 2 seconds in 😅  I think it's good 😂  If not, we'll find out pretty quick

Issue #558, opened 2020-12-08
No longer able to use model.to_param in route methods in Lucky


jwoertink:
When we start supporting views in models (very soon), those models won't have a primary key which means no `id` method. To support this in Avram, we had to remove the requirement of an `id` method as well as removing the model `to_param` method (which called .id). Because of this change, you can no longer call a route method like `Users::Show.with(current_user)`. This will fail because there's no `to_param` method. 

We should define the `to_param` method again, but this time make it a compile-time error that says you either need to override the method with your own `to_param`, or use a specific column.

PR #559, opened 2020-12-08
Add back Avram::Model#to_param

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/558

Adds an implementation when primary key is defined, otherwise raises a compilation error when used.

### DIFF
--- a/src/avram/primary_key_methods.cr
+++ b/src/avram/primary_key_methods.cr
@@ -65,4 +65,10 @@ module Avram::PrimaryKeyMethods
     query = yield base_query_class.new
     query.find(id)
   end
+
+  # For integration with Lucky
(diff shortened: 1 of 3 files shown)


src/avram/primary_key_methods.cr
```diff
@@ -65,4 +65,10 @@ module Avram::PrimaryKeyMethods
     query = yield base_query_class.new
     query.find(id)
   end
+
+  # For integration with Lucky
+  # This allows an `Avram::Model` to be passed into a Lucky::Action to create a url/path
```
jwoertink, line 70: 👍  good call.

jwoertink (approved):
sweet!

PR #560, opened 2020-12-08
Define Avram::Model#delete only when primary key defined

matthewmcgarvey:

The delete method only works when there is a primary key defined. This updates the implementation to raise a compile-time error if used when no primary key is defined.

Also a small modification to the other compile time error messages.


jwoertink (approved):
Looks good!

PR #561, opened 2020-12-09
Can preload has many associations with an existing record

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/531

This is one of the last updates to support preloading associations with existing records. It adds support for has many and has many through associations.

There were some overloaded methods added that are not defined for belongs to or has one associations to support has many through's. I will make another PR adding those so that all associations have every method overload.


jwoertink (approved):
Nice! 👍

PR #562, opened 2020-12-09
Add missing overloads for belongs to and has one preloading with existing records

matthewmcgarvey:
Related to https://github.com/luckyframework/avram/issues/531

There were overloads added of preloading functions in https://github.com/luckyframework/avram/pull/561 that were not defined for belongs_to or has_one associations.


jwoertink (approved):
oh nice. 👍

PR #564, opened 2020-12-14
Adding helper method for defaults.

### LINKED ISSUE #547: Setting default queries
Ref: https://github.com/luckyframework/website/issues/497#issue-751202165

We've always had the ability to set default queries by just adding them to your `initialize` method. This is super handy, and pretty intuitive since you always instantiate your query object. If this is really broken, then we need to fix it. Though, if it's broken due to how we construct queries, it may not be an easy fix.

jwoertink:
Fixes #547 

When we merged in https://github.com/luckyframework/avram/pull/411 this broke the ability to easily set defaults on your query objects.

Before that PR you could do this:

```crystal
class SomeQuery < User::BaseQuery
  def initialize
    age.gte(21)
  end
end

# WHERE age >= 21
SomeQuery.new
```

After that PR, you would have to manually set the query object to get the same effect

```crystal
class SomeQuery < User::BaseQuery
  def initialize
    self.query = age.gte(21).query
  end
end
```

With this PR, you can still set the query object manually if you want, but this adds in a new `defaults` method to make it a little more clear on what your intension is. Note that this method is set to `private` because you shouldn't call this from outside of the query class. 

```crystal
class SomeQuery < User::BaseQuery
  def initialize
    defaults &.age.gte(21)
  end
end

# WHERE age >= 21
SomeQuery.new
```

### DIFF
--- a/spec/queryable_spec.cr
+++ b/spec/queryable_spec.cr
@@ -20,9 +20,28 @@ class JSONQuery < Blob::BaseQuery
   end
 end
 
+class QueryWithDefault < User::BaseQuery
+
+  def initialize
+    defaults &.age.gte(21)
+  end
+end
+
 describe Avram::Queryable do
   it "can chain scope methods" do
     ChainedQuery.new.young.named("Paul")
+  end
+
+  it "can set default queries" do
+    query = QueryWithDefault.new.query
+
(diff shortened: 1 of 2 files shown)


matthewmcgarvey (approved):
🎉

PR #566, opened 2020-12-17
Updates to Avram Enums : ==, ===, enum constants

robacarp:
This implements `==` and `===` methods on avram_enum generated structures, and also implements explicit constant accessors to decrease the difference between Enum and avram_enum.

### DIFF
--- a/src/avram/charms/enum_extensions.cr
+++ b/src/avram/charms/enum_extensions.cr
@@ -24,6 +24,10 @@ macro avram_enum(enum_name, &block)
       @enum = Avram{{ enum_name }}.from_value(enum_value.to_i)
     end
 
+    def ==(other : {{ enum_name }}) : Bool
+      self.enum == other.enum
+    end
+
     delegate to_s, to_i, to: @enum
 
     forward_missing_to @enum
--- a/src/avram/charms/enum_extensions.cr
+++ b/src/avram/charms/enum_extensions.cr
@@ -1,12 +1,25 @@
+macro redeclare_avram_enum(name)
+  {% for member in name.resolve.constants %}
+    {{ member }} = {{ name }}::{{ member }}
+  {% end %}
+end
+
 macro avram_enum(enum_name, &block)
-  enum Avram{{ enum_name }}
+  {% avram_enum = ("Avram" + enum_name.names.join("::")).id %}
+  enum {{ avram_enum }}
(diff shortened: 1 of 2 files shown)


jwoertink (commented):
I love this. Question.... Should we also add in the threequals `===`? I think `case` uses that, right?

akadusei:
Using a `struct` for the enum wrapper instead of a `class` should solve the core problem this PR seeks to resolve.

An instance of a `class` is a different object from another instance of the same class. An instance of a `struct` will always be equal to another instance of the same struct.

robacarp:
@akadusei By converting the enum wrapper to a struct I was indeed able to remove the explicit definition of the #== operator. Honestly I don't know all the implications of this kind of change, though from all my reading a struct is better than a class wherever you can. I don't know what the knock-on effects of this will be in developer applications.

I added it here a a separate commit which can be rebased out if undesired.

src/avram/charms/enum_extensions.cr
```diff
@@ -37,10 +37,6 @@ macro avram_enum(enum_name, &block)
       @enum = Avram{{ enum_name }}.from_value(enum_value.to_i)
     end
 
-    def ==(other : {{ enum_name }}) : Bool
-      self.enum == other.enum
-    end
-
     delegate :===, to_s, to_i, to: @enum
```
akadusei, line 40: Do we still need to delegate `===` here?

akadusei:
Looks good. Apps should not have any issues, unless they are reopening the class (now struct) for some reason.

src/avram/charms/enum_extensions.cr
robacarp: Yep, the case equality tests fail without it.

src/avram/charms/enum_extensions.cr
```diff
@@ -1,13 +1,26 @@
...
+  enum {{ avram_enum }}
     {{ block.body }}
+
+    def ===(other : Int32)
+      value == other
+    end
   end
 
-  class {{ enum_name }}
+  struct {{ enum_name }}
```
jwoertink, line 17: nice! 👍

jwoertink (approved):
I haven't used the enum yet, but I think this looks good. Only way to really know is get it in the hands of people!

Issue #568, opened 2020-12-18
Primary key methods that are overridden can raise exceptions


jwoertink:
Take a look at this example: https://play.crystal-lang.org/#/r/a4y0

If you override one of the primary key methods to do your own deal, like in this example, you'll run in to a compile-time error when you wouldn't expect it.

The idea here was basically doing a custom delete where it was expected to basically delete the current model, and delete another model. 

In any case, the fix for this is to add `include Avram::PrimaryKeyMethods` to the `Base` model, and things are fine. I don't think this is something we need to patch because it's sort of an edge case that you'd define your own version of one of these methods when they're already included; however, I think we can work out the error message a little nicer to maybe give the user a heads up that they can include this module if they need.

robacarp:
Is the suggestion then to do something with a before/after hook?

jwoertink:
Well, personally, I'd recommend just not overriding methods like `delete`, but if you need to, then the suggestion is to just include the `Avram::PrimaryKeyMethods` module in your model class that overwrote the delete method.

My guess is the next release we're going to start working on will actually solve the original issue of why it had to be overwritten in the first place, and this will all be moot.

matthewmcgarvey:
Just using this as a place to put my thoughts...

The way it is currently set up, `Avram::Model` has all the methods defined that would normally require a primary key but they raise a compile time error, instead. In user defined models, when `primary_key` is used, then it includes the `Avram::PrimaryKeyMethods` module that actually defines those methods.

It's set up in this way so that `Avram::Model` _always_ has those methods available and classes that take in and call those methods don't have to have special type signatures.

The failure of this setup is that people might want to override and implement special logic around the methods that were moved into the module and then call `super` to get the original functionality. This is totally understandable and was not considered when the changes were made.

## Potential Solution

I'm thinking that we might go down the route that Rails has for creating read-only models where it is a configuration switch in the model. Instead of only including the features when they are needed in the model, we take them away when they are not used. How this looks in Avram, I'm not sure.  There's two pieces that would need to have separate fixes:

- the primary-key being present/missing
- the db table being read-only

PR #573, opened 2020-12-24
Features/delete operations

### LINKED ISSUE #413: Delete should be an operation
Currently, `delete` is a method in `Avram::Model`. It should be an operation. I need a `before` method to check if the user is permitted to delete the object, and if other constraints are satisfied. And after my object is deleted, I need an `after` method to clean up: change object counts I maintain in its parents, make sure nothing references that record's URL, etc. And I don't have access to the same error mechanism used in record creation and updating.

For now, I guess I can hard-code these into my action, but I will have to add a transaction there too. If `delete` was an operation, the transaction,  the callback structure and the error mechanism would be there. There is also the matter of symmetry, currently `delete` is asymmetrical with object creation.

I think this is what he was getting at in https://github.com/luckyframework/lucky/issues/559, filed on Lucky (rather than Avram) two years ago.

I think it also makes sense to access form fields from the delete operation. One can have a "Yes, I know what I'm doing!" field, and a field that says "delete all of the children of this record too", and one can throw the delete operation back to the action asking that the user correct decisions in the forms.

jwoertink:
Fixes #413 
Related: #568 

This is a pretty big PR, so I'll try to break down motivation, and what each bit is doing.

## Motivation

Deleting an object is generally pretty easy. You click a button, sometimes get "Are you sure?" and you're done. But I've noticed it's becoming more common place for sites to add some sort of confirmation. Github makes you type the name of a repo before deleting. Amazon has you type "DELETE ME" or something like that before it deletes some things. To implement these, you'll need more than just a button, and what better way than an Operation?

Next thing is when using soft deletes. With regular deletes, we can rely on the database to cascade. With soft deletes, it's up to us to handle. If a user becomes soft deleted, you may want all of their associated stuff to also be "soft deleted".

Finally, maybe your system uses roles to see who can delete. Generally these might go in your action to decide what a `current_user` has access to delete, but now with a DeleteOperation, you can say it needs a user, and use a `before_delete` callback to ensure the user has the role you require.

## Interface

It should look and feel pretty familiar. I essentially just copied what the SaveOperation does.

```crystal
# src/operations/delete_user.cr
class DeleteUser < User::DeleteOperation
  attribute confirm_delete : String

  before_delete do
    if confirm_delete.value == "nO, dOn'T rEaLly DeLeTe It 🤪"
      confirm_delete.add_error("You need to confirm that you really want this to happen")
    end
  end

  after_delete do |deleted_user|
    GoodByeUserEmail.new(deleted_user.email).deliver_now
  end
end

# src/actions/users/delete.cr
class Users::Delete < BrowserAction
  delete "/users/:user_id" do
    user = UserQuery.find(user_id)

    DeleteUser.destroy(user) do |operation, deleted_user|
      if operation.deleted?
         redirect to: Home::Index
      else
         TimeBomb.start_countdown(10.seconds)
      end
    end
  end
end
```

## Final thoughts

We're at a point with this where's there's quite a bit of duplication. I don't want to do a feature and refactor in the same PR, so a refactor is something we can look at in the future to help clean some of this up. Also, since I did literally copy quite a bit, there may be some code moved over that's not used; if someone sees any, feel free to point it out.


src/avram/needy_initializer_and_delete_methods.cr
```diff
@@ -0,0 +1,190 @@
...
+        {{ attribute_params.id }}
+      )
+
+      {% if with_bang %}
+        operation.delete!
+      {% else %}
+        if operation.delete
+          yield operation, operation.record.not_nil!
+        else
+          #operation.published_delete_failed_event
```
jwoertink, line 132: oops. This is actually in the delete_operation.cr already.

src/avram/delete_operation_template.cr
```diff
@@ -0,0 +1,21 @@
...
+      ::{{ type }}::BaseQuery
+    end
+
+    def delete_operation_class : ::{{ type }}::DeleteOperation.class
+      ::{{ type }}::DeleteOperation
+    end
+
+    class ::{{ type }}::DeleteOperation < Avram::DeleteOperation({{ type }})
+      macro inherited
+        FOREIGN_KEY = "{{ type.stringify.underscore.id }}_id"
```
jwoertink, line 15: I don't think I'm using this. My thought was maybe we need to know the foreign key to delete an associated record?  But the database probably handles it directly 🤔

src/avram/errors.cr
```diff
@@ -37,13 +37,13 @@ module Avram
...
 
-  # Raised when using the create! or update! methods on an operation when it does not have the proper attributes
+  # Raised when using the create!, update!, or destroy! methods on an operation when it does not have the proper attributes
   class InvalidOperationError < AvramError
     getter errors : Hash(Symbol, Array(String))
 
     def initialize(operation)
       message = String.build do |string|
-        string << "Could not save #{operation.class.name}."
+        string << "Could not complete #{operation.class.name}."
```
jwoertink, line 46: Also thought of "Could not perform". I'm open to other wording suggestions to make this more generic.

src/avram/delete_operation.cr
```diff
@@ -0,0 +1,234 @@
...
+
+    private def column_attributes
+      [
+        {% for attribute in attributes %}
+          {{ attribute[:name] }},
+        {% end %}
+      ]
+    end
+
+    def required_attributes
```
jwoertink, line 224: pretty sure this isn't needed.

src/avram/delete_operation.cr
```diff
@@ -0,0 +1,234 @@
...
+  macro inherited
+    @@permitted_param_keys = [] of String
+
+    @record : T
+    @params : Avram::Paramable
+    getter :record, :params
+    property delete_status : DeleteStatus = DeleteStatus::Unperformed
+  end
+
+  def self.delete(*args, **named_args, &block)
```
jwoertink, line 33: Maybe this can be moved in to that `NeedyInitializerAndDeleteMethods` so it can generate both `delete` and `delete!` with or without the block?

src/avram/delete_operation.cr
```diff
@@ -0,0 +1,234 @@
...
+    Deleted
+    DeleteFailed
+    Unperformed
+  end
+
+  macro inherited
+    @@permitted_param_keys = [] of String
+
+    @record : T
+    @params : Avram::Paramable
```
jwoertink, line 28: The SaveOperation doesn't have these in the `inherited`, but I kept getting an error saying `@record wasn't initialized properly` or whatever. That may be because the SaveOperation has a nilable record, but DeleteOperation shouldn't.

matthewmcgarvey:
Right now you can call `model.delete` and you can also call delete on a query. Do you think we should handle that as well? That will allow bulk deleting and avoid actually loading the record if you don't want to but maintaining the operation style

jwoertink:
I had thought about that too. It's sort of weird calling delete on a query object. My original thought was what if we made an override that allowed the operation to take in a query?

```crystal
DeleteUsers.destroy_all(UserQuery.new.inactive) do |operation, user_count|

end
```

You could create either use the same operation in your code, or make a separate one so one handles single delete, and one handles bulk delete. The only thing is that SaveOperation doesn't work in bulk.... but maybe it should? `SaveUsers.update_all(query_obj)` ? 🤔

matthewmcgarvey (commented):
Like you were saying, large PR with a lot of copy/paste is difficult to review

src/avram/callbacks.cr
```diff
@@ -72,6 +72,19 @@ module Avram::Callbacks
...
+  # Same as `before_save`, but with a different name
+  macro before_delete(method_name, if _if = nil, unless _unless = nil)
+    {% unless _if.is_a?(SymbolLiteral) || _if.is_a?(NilLiteral) %}
+      conditional_error_for_inline_callbacks(:before_delete, {{ method_name }}, :if)
+    {% end %}
+    {% unless _unless.is_a?(SymbolLiteral) || _unless.is_a?(NilLiteral) %}
+      conditional_error_for_inline_callbacks(:before_delete, {{ method_name }}, :unless)
+    {% end %}
+    before_delete(if: {{ _if }}, unless: {{ _unless }}) do
+      {{ method_name.id }}
```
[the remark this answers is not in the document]
jwoertink: hmm... 🤔  That's not a bad idea. I guess you technically have access to it through just calling `self.record`. I'd be ok with adding it in though.

src/avram/delete_operation.cr
```diff
@@ -0,0 +1,234 @@
...
+require "./needy_initializer_and_delete_methods"
+
+abstract class Avram::DeleteOperation(T)
+  include Avram::NeedyInitializerAndDeleteMethods
+  include Avram::DefineAttribute
+  include Avram::Validations
+  include Avram::OperationErrors
+  include Avram::ParamKeyOverride
+  include Avram::Callbacks
+  include Avram::InheritColumnAttributes
```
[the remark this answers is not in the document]
jwoertink: My thought was you may want to use those like checking if the user has certain permissions, or maybe checking things like `created_at` to say you can't delete since you just created your account. Dunno, but it could add lots of extra flexibility.

src/avram/delete_operation.cr
jwoertink: Correct. We actually add things like this throughout the codebase to help guide people that may want to call certain methods. Of course you'd just get a "undefined method" error normally, but this just lets us direct the user to use `destroy` instead.

matthewmcgarvey:
This would also fix https://github.com/luckyframework/lucky/issues/559

spec/operations/delete_operation_spec.cr
```diff
@@ -0,0 +1,113 @@
...
+
+      BasicDeleteUser.destroy(user) do |operation, deleted_user|
+        operation.valid?.should be_true
+        operation.delete_status.should eq BasicDeleteUser::DeleteStatus::Deleted
+        deleted_user.not_nil!.name.should eq user.name
+        UserQuery.new.select_count.should eq 0
+      end
+    end
+
+    it "fails to delete the specified record" do
```
matthewmcgarvey, line 31:
```suggestion
    it "does not delete if operation invalid" do
```

spec/operations/delete_operation_spec.cr
```diff
@@ -0,0 +1,113 @@
...
+    end
+  end
+
+  describe "soft deletes" do
+    it "returns a soft deleted object" do
+      item = SoftDeletableItemBox.create
+
+      deleted_item = SoftDeleteItem.destroy!(item)
+
+      deleted_item.soft_deleted?.should be_true
```
matthewmcgarvey, line 68: Also check that it still exists in the database?

spec/operations/delete_operation_spec.cr
```diff
@@ -0,0 +1,113 @@
...
+    end
+  end
+
+  describe "publishes" do
+    it "publishes a successful delete" do
+      Avram::Events::DeleteSuccessEvent.subscribe do |event|
+        if event.operation_class == "BasicDeleteUser"
+          UserQuery.new.select_count.should eq 0
+        end
+      end
```
matthewmcgarvey, line 92:
```suggestion
      events = [] of Avram::Events::DeleteSuccessEvent
      Avram::Events::DeleteSuccessEvent.subscribe do |event|
        events << event
      end
```

And then can we do assertions on that list after we call `destroy!`. I don't like the idea that we wouldn't know if the subscribe block isn't running

spec/operations/delete_operation_spec.cr
```diff
@@ -0,0 +1,113 @@
...
+
+      BasicDeleteUser.destroy!(user)
+    end
+
+    it "publishes a failed delete" do
+      Avram::Events::DeleteFailedEvent.subscribe do |event|
+        event.operation_class.should eq("FailedToDeleteUser")
+        event.error_messages_as_string.should contain("not today")
+        UserQuery.new.select_count.should eq 1
+      end
```
matthewmcgarvey, line 104: same here as above

src/avram/delete_operation_template.cr
```diff
@@ -0,0 +1,18 @@
+class Avram::DeleteOperationTemplate
+  macro setup(type, columns, *args, **named_args)
+
+    # This makes it easy for plugins and extensions to use the base SaveOperation
+    def base_query_class : ::{{ type }}::BaseQuery.class
+      ::{{ type }}::BaseQuery
+    end
```
matthewmcgarvey, line 7:
```suggestion
```

Duplicate of https://github.com/luckyframework/avram/blob/be876ef87ae120650ce27b93f586ba314570663a/src/avram/save_operation_template.cr#L4-L7

src/avram/events/delete_failed_event.cr
```diff
@@ -0,0 +1,17 @@
...
+
+  def initialize(
+    @operation_class : String,
+    @errors : Hash(Symbol, Array(String))
+  )
+  end
+
+  def error_messages_as_string
+    String.build do |msg|
+      errors.map do |key, messages|
```
matthewmcgarvey, line 12:
```suggestion
      errors.each do |key, messages|
```

src/avram/callbacks.cr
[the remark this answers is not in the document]
jwoertink: I keep going back and forth on this... Would it be too weird that it doesn't match up with the `before_run` and `before_save` interfaces? Those make sense to not pass down the variable... Do you think it would be more helpful than confusing?

matthewmcgarvey (approved):
I'm happy enough with the code, as is. We can definitely improve it later. Nice job!

PR #574, opened 2020-12-25
fixing bug with associations

### LINKED ISSUE #571: Has_many association too greedy
I really don't want to spam Lucky repos issues but after migrating to 0.25 I got some some errors (https://github.com/luckyframework/website/issues/541) which I thought were fixed. 

Now I have some weird behaviors which are best explained by this custom Avram spec which fails :

```crystal
    it "joins the two associations correctly" do
      tag = TagBox.create
      post = PostBox.create
      post2 = PostBox.create
      TagBox.create
      TaggingBox.new.tag_id(tag.id).post_id(post.id).create

      post2.tags.size.should eq 0
    end
```

which is a variation of the original :

```crystal
    it "joins the two associations" do
      tag = TagBox.create
      post = PostBox.create
      TagBox.create
      TaggingBox.new.tag_id(tag.id).post_id(post.id).create

      post.tags.should eq [tag]
    end
 ```
 
 post2.tags lists **_all_** tags regardless if a `Tagging` binds them with `post2`.
 
 In my application, `current_user.groups` returns all groups available in the database, which is very concerning security wise. It was working as expected before 0.25.
 
 If this issue is moot because of my misunderstanding (most probable), what is the best way to get all the tags of a post ? Is post.tags a code smell ? Should I use a PostQuery instead ? In other words, what does `post.tags` really do ? post is a Post instance and not a PostQuery, right ? So what does it gets translated into ?

jwoertink:
Fixes #571

Turns out this affected all associations. We never caught this because the tests only ever setup a single record. Since we were calling `type`, the associations never matched because `Tag` would never equal `:belongs_to`.

### DIFF
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -4,12 +4,14 @@ describe Avram::Model do
   describe "has_many" do
     it "gets the related records" do
       post = PostBox.create
+      post2 = PostBox.create
       comment = CommentBox.new.post_id(post.id).create
 
       post = Post::BaseQuery.new.find(post.id)
 
-      post.comments.to_a.should eq [comment]
+      post.comments.should eq [comment]
       comment.post.should eq post
+      post2.comments.size.should eq 0
     end
 
     it "gets the related records for nilable association that exists" do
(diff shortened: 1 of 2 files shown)


matthewmcgarvey (approved):
Good catch 👍

Issue #422, opened 2020-07-11
Migrator obscures stack trace of unhandled exceptions


digitalcora:
Trying to set up a Lucky project for the first time, I ran into a strange error that took quite a while to debug. Part of the difficulty was that it turned out Avram was suppressing the real stack trace of the error.

in https://github.com/luckyframework/avram/blob/63be781/src/avram/migrator/migrator.cr:

```crystal
module Avram::Migrator
  def self.run
    yield
  rescue e : PQ::PQError
    raise e.message.colorize(:red).to_s
  rescue e : Exception   # <=
    raise e.message.to_s # <=
  end
end
```

The marked lines catch all exceptions and re-raise them, with a stack trace pointing here instead of the original source. The purpose of these lines is unclear to me, and in fact once I deleted them from my project's copy of Avram in `lib`, I got a much better stack trace that allowed me to solve the problem.

If there would be no harm in doing so, maybe these lines could be removed?

jwoertink:
Hey @digitalcora Welcome to Lucky! Sorry you ran in to some issues. Hopefully you got it all set now. 

That is a good call. Normally when we catch errors like that, it's because we can usually provide a better error message than the one provided by Crystal's default. In this case since we're just re-raising the same error, I'm not seeing any benefit to having it. My guess is that it would be ok to remove those. What are your thoughts @paulcsmith ?

paulcsmith:
I think the original reason was to highlight the error in red. But if it makes the stacktrace worse then we should remove it :D Thanks for opening the issue @digitalcora!

matthewmcgarvey:
These are also places I believe are obscuring problems during migrations:

https://github.com/luckyframework/avram/blob/1e6f4edd4d49c518c858ee9ecc277a3b963051eb/src/avram/migrator/runner.cr#L192-L193

https://github.com/luckyframework/avram/blob/1e6f4edd4d49c518c858ee9ecc277a3b963051eb/src/avram/migrator/migration.cr#L99-L108

PR #577, opened 2020-12-26
Don't obscure errors during migrations

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/422

While the errors might look "uglier", they will now definitely contain the error message and the stacktrace.

Instead of dropping exceptions for new messages with a new stacktrace, this raises a new error with the original error as the cause so that we can add info without forgetting context. The only problem was that Crystal does not print the "cause" of unhandled exceptions ([source](https://github.com/crystal-lang/crystal/blob/33670931a7ac56fa3e7b737957a9c6413605eb83/src/crystal/system/print_error.cr#L15-L28)). To overcome this, I updated `Avram::Migrator` to print the error so that we can get the cause and then to exit so that it's still recognized as a failure.

### DIFF
--- a/spec/migrator/migration_spec.cr
+++ b/spec/migrator/migration_spec.cr
@@ -66,7 +66,7 @@ end
 
 describe Avram::Migrator::Migration::V1 do
   it "executes statements in a transaction" do
-    expect_raises Exception, %(relation "table_does_not_exist" does not exist) do
+    expect_raises Avram::FailedMigration do
       MigrationThatPartiallyWorks::V999.new.up
     end
 
--- a/src/avram/migrator/migration.cr
+++ b/src/avram/migrator/migration.cr
@@ -97,15 +97,7 @@ abstract class Avram::Migrator::Migration::V1
       end
     end
   rescue e : PQ::PQError
-    raise <<-ERROR
-    There was a problem running this statement:
-
-      #{statements.join("\n")}
-
-    Problem:
-
-      #{e.message}
-    ERROR
+    raise FailedMigration.new(migration: self.class.name, statements: statements, cause: e)
   end
 
   def reset_prepared_statements
--- a/src/avram/migrator/migrator.cr
+++ b/src/avram/migrator/migrator.cr
@@ -1,9 +1,8 @@
 module Avram::Migrator
   def self.run
     yield
-  rescue e : PQ::PQError
-    raise e.message.colorize(:red).to_s
-  rescue e : Exception
-    raise e.message.to_s
+  rescue e
+    puts e.inspect_with_backtrace
+    exit 1
   end
 end
--- a/src/avram/tasks/db/base_task.cr
+++ b/src/avram/tasks/db/base_task.cr
@@ -1,0 +1,9 @@
+abstract class BaseTask < LuckyCli::Task
+  abstract def run_task
+
+  def call
+    Avram::Migrator.run do
+      run_task
+    end
+  end
+end
--- a/tasks.cr
+++ b/tasks.cr
@@ -3,19 +3,5 @@ require "./src/avram"
 require "./config/*"
 require "./db/migrations/*"
 
-class Db::Reset < LuckyCli::Task
-  summary "Drop, creates and migrates the database"
-
-  def call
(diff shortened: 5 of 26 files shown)


src/avram/migrator/migration.cr
```diff
@@ -96,16 +96,6 @@ abstract class Avram::Migrator::Migration::V1
         yield tx
       end
     end
-  rescue e : PQ::PQError
```
jwoertink, line 99: Any idea if the actual stack trace will include the statement that failed? I'm wondering if instead of removing this one, we just figure out how to add the filename to it. I'm worried letting it error on its own wouldn't really provide that nice of an error.

jwoertink (commented):
Just the one comment, but I think the rest is good to go. I'm also noticing the interface for these tasks are a bit inconsistent. Like some run and then do a puts, but others don't. Another PR for another time 😂

src/avram/migrator/migration.cr
matthewmcgarvey: So it wouldn't. One thing I was thinking about is handling all errors that are raised from migrations, wrapping them with more information (name of migration, statement that caused it), and printing them ourselves so that we can include the cause of exceptions and then exiting as a failure. What do you think?

matthewmcgarvey: Calling https://crystal-lang.org/api/0.35.1/Exception.html#inspect_with_backtrace:String-instance-method on exceptions includes the cause, I don't know why Crystal doesn't use it in the unhandled exception code flow.

jwoertink: Yeah, I like that. One thing that sets us apart from others is how we display really nice error messages. Do you want to do that in this PR, or a different one? I'm cool with either way.

matthewmcgarvey: I'll update this PR. I believe this will help https://github.com/luckyframework/avram/pull/578#issuecomment-751812307 as well

matthewmcgarvey: @jwoertink done 👍  this is much better, I think

spec/migrator/migration_spec.cr
```diff
@@ -66,7 +66,7 @@ end
 
 describe Avram::Migrator::Migration::V1 do
   it "executes statements in a transaction" do
-    expect_raises Exception, %(relation "table_does_not_exist" does not exist) do
+    expect_raises Avram::FailedMigration do
```
matthewmcgarvey, line 69: Can no longer assert on the message because it is in the cause, not in the expected error

src/avram/migrator/migrator.cr
```diff
@@ -1,9 +1,8 @@
 module Avram::Migrator
   def self.run
     yield
-  rescue e : PQ::PQError
-    raise e.message.colorize(:red).to_s
-  rescue e : Exception
-    raise e.message.to_s
+  rescue e
+    puts e.inspect_with_backtrace
+    exit 1
```
matthewmcgarvey, line 6: This is to be able to print the error with the cause but it means we are explicitly `exit`-ing in code

src/avram/tasks/db/base_task.cr
```diff
@@ -0,0 +1,9 @@
+abstract class BaseTask < LuckyCli::Task
```
matthewmcgarvey, line 1: This abstract class was pulled out so that we could test and use tasks without accidentally running into the `exit` code.

tasks.cr
```diff
@@ -3,19 +3,5 @@ require "./src/avram"
 require "./config/*"
 require "./db/migrations/*"
 
-class Db::Reset < LuckyCli::Task
```
matthewmcgarvey, line 6: There is already a separate `Db::Reset` task that was overwriting this one. This was old and I noticed it while updating tasks to the new pattern

src/avram/tasks/db/base_task.cr
jwoertink: oh nice! Good call.

jwoertink (approved):
Nice! Yeah, I feel a lot better about this direction 👍

PR #578, opened 2020-12-27
Generate UUID primary keys in the database

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/122

When a user creates a migration that uses a UUID primary key, the migration will automatically set the default value to be `gen_random_uuid()` which comes from the `pgcrypto` extension. If they do not have that extension enabled, the migration will fail saying that `gen_random_uuid()` is an unknown function. They will need to add `enable_extension "pgcrypto"` in their migration before creating the table to get it working.

This does not remove or change how Avram currently sets the id value itself before inserting the record into the database https://github.com/luckyframework/avram/blob/1e6f4edd4d49c518c858ee9ecc277a3b963051eb/src/avram/save_operation_template.cr#L15-L22

We will definitely need to consider how we want to make that change but it doesn't need to be considered right now since there's no harm in doing both.

### DIFF
--- a/db/migrations/20180625202051_create_line_item.cr
+++ b/db/migrations/20180625202051_create_line_item.cr
@@ -1,5 +1,7 @@
 class CreateLineItem::V20180625202051 < Avram::Migrator::Migration::V1
   def migrate
+    enable_extension "pgcrypto"
+
     create :line_items do
       primary_key id : UUID
       add_timestamps
--- a/db/migrations/20180628193054_create_product.cr
+++ b/db/migrations/20180628193054_create_product.cr
@@ -4,9 +4,6 @@ class CreateProduct::V20180628193054 < Avram::Migrator::Migration::V1
       primary_key id : UUID
       add_timestamps
     end
-
-    enable_extension "uuid-ossp"
-    execute("ALTER TABLE products ALTER COLUMN id SET DEFAULT uuid_generate_v4();")
   end
 
   def rollback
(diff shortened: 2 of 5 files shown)


jwoertink (approved):
Interesting. I've never used the pgcrypto before, only ever used the `uuid-ossp` one. Taking a look at the docs, it looks like this is supported back to PG 9.X at least, and the uuid-ossp docs actually say

> Note: If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.

So I think this is great 👍

wontruefree:
I have to do this all the time in Rails and Lucky.  Thanks for this!

matthewmcgarvey:
@jwoertink I'm not sure if I made it clear but this would be a breaking change since migrations won't work that have UUID primary keys until they enable the extension. Are you ok with this going in?

jwoertink:
Oh, like existing migrations where people might rollback or whatever? If you happen to know a way we can catch that, or show a nice error like "You need to enable the pgcrypto extension to use UUID. Here's how: ....". If not, then I think it's fine to go in.

matthewmcgarvey:
After looking into it, there's no simple way to add a helpful error to tell users to enable the extensions. For right now, I'm fine with the way it works. If it becomes a problem we can add it. It would be a rescue somewhere that looks for a message like `function gen_random_uuid() does not exist` and then wrap it. It would have to be a general rescue

matthewmcgarvey:
Users with existing tables without the default will need to create a migration that does this for every table in the database that needs it

```crystal
execute("ALTER TABLE users ALTER COLUMN id SET DEFAULT gen_random_uuid();")
```

PR #579, opened 2020-12-29
Restrict validate_size_of to string attributes

### LINKED ISSUE #570: validate_size_of implementation is confused
The comments here confuse whether validate_size_of is supposed to validate the numeric value of something, or the size of something. This produces incorrect documentation at https://luckyframework.github.io/avram/Avram/Validations.html#validate_size_of(attribute:Avram::Attribute,min=nil,max=nil,allow_nil:Bool=false)-instance-method
This is also reflected in the incorrect documentation in the guides on the web site, see https://github.com/luckyframework/website/issues/489

```
  # Validate the size of the attribute is within a `min` and/or `max`
  #
  # ```
  # validate_size_of age, min: 18, max: 100
  # validate_size_of account_balance, min: 500
  # ```
  def validate_size_of(
    attribute : Avram::Attribute,
    min = nil,
    max = nil,
    allow_nil : Bool = false
  )
    if !min.nil? && !max.nil? && min > max
      raise ImpossibleValidation.new(
        attribute: attribute.name,
        message: "size greater than #{min} but less than #{max}")
    end

    unless allow_nil && attribute.value.nil?
      size = attribute.value.to_s.size
```

matthewmcgarvey:
Fixes #570 

I have seen `validate_size_of` mistakenly used to validate a numeric value since the name doesn't give any indication that the implementation expects strings. As the connected issue points out, even the documentation seems confused.

This change updates the type signature to allow only attributes that contains strings. While the type signature says it only takes in `Avram::Attribute(String?)`, I found it also handles `Avram::Attribute(String)`. I didn't expect it to be that smart 😄

### DIFF
--- a/spec/validations_spec.cr
+++ b/spec/validations_spec.cr
@@ -13,7 +13,7 @@ class UniquenessWithCustomMessageSaveOperation < User::SaveOperation
   end
 end
 
-private def attribute(value)
+private def attribute(value : T) : Avram::Attribute(T) forall T
   Avram::Attribute.new(value: value, param: nil, param_key: "fake", name: :fake)
 end
 
--- a/src/avram/validations.cr
+++ b/src/avram/validations.cr
@@ -110,7 +110,7 @@ module Avram::Validations
   # validate_size_of api_key, is: 32
   # ```
   def validate_size_of(
-    attribute : Avram::Attribute,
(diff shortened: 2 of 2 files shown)


jwoertink (approved):
Yup 👍  I've updated the docs on the website already that reflect this change.

Issue #537, opened 2020-11-17
Add validation for numeric values


matthewmcgarvey:
The lack of a numeric validation and the presence of `validate_size_of` causes confusion on how to validate numbers.

At the very least it should have a `min` and `max` field.

Ex: `validate_numeric inventory_count, min: 0, allow_nil: false`

Related: https://github.com/luckyframework/lucky/issues/1308

PR #580, opened 2020-12-29
Add validation for numbers

matthewmcgarvey:
Completes #537 

```crystal
validate_numeric age, greater_than: 18, less_than: 65
```

Any type of number can be used for the attribute or the conditions since they can all be compared.


jwoertink (approved):
👍

PR #581, opened 2020-12-29
Load correct has_one association with lazy loading

matthewmcgarvey:
The code was still depending on the query being mutated instead of immutable so it was always loading the first association in the database. This only affected code that was using lazy loading (by enabling it or using the bang methods).


robacarp (approved):

jwoertink (approved):
Good catch

PR #583, opened 2020-12-31
Update error message when attribute called with type union

### LINKED ISSUE #466: Confusing error when trying to use a nilable attribute

matthewmcgarvey:
Fixes #466 

CleanShot 2020-12-30 at 22 52 51@2x

### DIFF
--- a/src/avram/define_attribute.cr
+++ b/src/avram/define_attribute.cr
@@ -36,7 +36,22 @@ module Avram::DefineAttribute
 
   macro attribute(type_declaration)
     {% if type_declaration.type.is_a?(Union) %}
-      {% raise "attribute must use just one type" %}
+      {% if type_declaration.value.is_a?(Nop) %}
+        {% default_value = "" %}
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Yeah, this is WAY better!

PR #316, opened 2020-03-08
[WIP] Add 'query' option to 'has_many'

paulcsmith:
Closes #85

PR #584, opened 2021-01-01
Had to get 1 last PR in for 2020...

### LINKED ISSUE #85: Allow specifying conditionals in associations
https://gitter.im/luckyframework/Lobby?at=5cd5dd6f79d70050994c8059

```crystal
class Posts
  table :posts do
    # No conditions
    has_many comments : Comment
    # Only return public comments related to this post
    has_many public_comments : Comment, filter: CommentQuery.new.public(true)
  end
end
```

jwoertink:
Fixes #85
Closes #316 

Because of all of the changes to associations since #316 was opened, I didn't want to fight conflicts. This essentially just copies what @paulcsmith already wrote though.

### DIFF
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -68,18 +68,6 @@ describe Avram::Model do
       post2.tags.size.should eq 0
     end
 
-    it "joins associations with the query option" do
-      tag = TagBox.create
-      funny_tag = TagBox.create &.name(Tag::FUNNY_TAG_NAME)
-      post = PostBox.create
-
-      TaggingBox.new.tag_id(tag.id).post_id(post.id).create
-      TaggingBox.new.tag_id(funny_tag.id).post_id(post.id).create
-
-      post.tags.should eq [tag, funny_tag]
-      post.funny_tags.should eq [funny_tag]
-    end
-
     it "counts has_many through belongs_to associations" do
       tag = TagBox.create
       post = PostBox.create
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -27,6 +27,19 @@ describe Avram::Model do
       manager.customers.should eq [customer]
     end
 
+    it "gets the related records for a query association" do
+      post = PostBox.create
+      post2 = PostBox.create
+      comment = CommentBox.create &.post_id(post.id)
+      nice_comment = CommentBox.create &.nice.post_id(post.id)
+
+      post.comments.should eq [comment, nice_comment]
+      post.nice_comments.should eq [nice_comment]
+
+      post2.comments.size.should eq 0
+      post2.nice_comments.size.should eq 0
+    end
+
     it "returns nil for nilable association that doesn't exist" do
       employee = EmployeeBox.create
       employee.manager.should eq nil
@@ -53,6 +66,18 @@ describe Avram::Model do
 
       post.tags.should eq [tag]
       post2.tags.size.should eq 0
+    end
+
+    it "joins associations with the query option" do
+      tag = TagBox.create
+      funny_tag = TagBox.create &.name(Tag::FUNNY_TAG_NAME)
+      post = PostBox.create
+
+      TaggingBox.new.tag_id(tag.id).post_id(post.id).create
+      TaggingBox.new.tag_id(funny_tag.id).post_id(post.id).create
+
+      post.tags.should eq [tag, funny_tag]
+      post.funny_tags.should eq [funny_tag]
     end
 
     it "counts has_many through belongs_to associations" do
--- a/spec/support/models/post.cr
+++ b/spec/support/models/post.cr
@@ -8,8 +8,11 @@ class Post < BaseModel
     column title : String
     column published_at : Time?
     has_many comments : Comment
+    has_many nice_comments : Comment,
+      query: Comment::BaseQuery.new.body(Comment::NICE_COMMENT_BODY)
     has_many taggings : Tagging
     has_many tags : Tag, through: [:taggings, :tag]
+    has_many funny_tags : Tag, through: [:taggings, :tag], query: Tag::BaseQuery.new.name(Tag::FUNNY_TAG_NAME)
   end
 end
 
--- a/spec/support/models/post.cr
+++ b/spec/support/models/post.cr
@@ -12,7 +12,6 @@ class Post < BaseModel
       query: Comment::BaseQuery.new.body(Comment::NICE_COMMENT_BODY)
     has_many taggings : Tagging
     has_many tags : Tag, through: [:taggings, :tag]
-    has_many funny_tags : Tag, through: [:taggings, :tag], query: Tag::BaseQuery.new.name(Tag::FUNNY_TAG_NAME)
   end
 end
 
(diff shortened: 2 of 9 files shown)


spec/associations_spec.cr
```diff
@@ -55,6 +68,18 @@ describe Avram::Model do
...
+    it "joins associations with the query option" do
+      tag = TagBox.create
+      funny_tag = TagBox.create &.name(Tag::FUNNY_TAG_NAME)
+      post = PostBox.create
+
+      TaggingBox.new.tag_id(tag.id).post_id(post.id).create
+      TaggingBox.new.tag_id(funny_tag.id).post_id(post.id).create
+
+      post.tags.should eq [tag, funny_tag]
+      post.funny_tags.should eq [funny_tag]
```
jwoertink, line 80: This spec fails `Nil assertion failed (NilAssertionError)` which wasn't very helpful running this locally... Maybe @matthewmcgarvey has some idea?

src/avram/associations/has_many.cr
```diff
@@ -180,16 +181,16 @@ module Avram::Associations::HasMany
     end
 
     private def lazy_load_{{ assoc_name }} : Array({{ model }})
+      {% query = query || "#{model}::BaseQuery.new".id %}
       {% if through %}
-        through_results = {{ through.first.id }}_query.preload_{{ through[1].id }}.results
+        through_results = {{ through.first.id }}_query.preload_{{ through[1].id }}({{ query }}).results
```
jwoertink, line 186: In my spec, this would expand to `taggings_query.preload_tags(Tag::BaseQuery.new.name(...).results)`. But I'm not sure if we can always just pass in the BaseQuery like that and have it work. Maybe this line has to change depending on if a `query` is passed in?

spec/support/models/post.cr
```diff
@@ -8,8 +8,11 @@ class Post < BaseModel
     column title : String
     column published_at : Time?
     has_many comments : Comment
+    has_many nice_comments : Comment,
+      query: Comment::BaseQuery.new.body(Comment::NICE_COMMENT_BODY)
     has_many taggings : Tagging
     has_many tags : Tag, through: [:taggings, :tag]
+    has_many funny_tags : Tag, through: [:taggings, :tag], query: Tag::BaseQuery.new.name(Tag::FUNNY_TAG_NAME)
```
jwoertink, line 15: Maybe we don't allow the `query` on `through` to start? We could always add that in later 🤔

src/avram/associations/has_many.cr
```diff
@@ -37,15 +37,16 @@ module Avram::Associations::HasMany
...
-    define_has_many_base_query({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{ through }})
+    define_has_many_lazy_loading({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{ through }}, {{ query }})
+    define_has_many_base_query({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{ through }}, {{ query }})
   end
 
-  private macro define_has_many_base_query(assoc_name, model, foreign_key, through)
+  private macro define_has_many_base_query(assoc_name, model, foreign_key, through, query)
     class BaseQuery
       def self.preload_{{ assoc_name }}(record)
         preload_{{ assoc_name }}(record: record, preload_query: {{ model }}::BaseQuery.new)
```
matthewmcgarvey, line 52:
```suggestion
        preload_{{ assoc_name }}(record: record, preload_query: {{ query }})
```

right? Since query could be nil, you might update the `has_many` macro to default it to `{{ model }}::BaseQuery.new` if missing

There's other places below that news up a query as well

src/avram/associations/has_many.cr
matthewmcgarvey: see my comment above to default it if it's nil

spec/support/models/post.cr
matthewmcgarvey, line 15: I don't think has many through will work as expected if the "through" has a query provided...

matthewmcgarvey (commented):
You picked quite an issue to make your last PR 😆 . There are going to be so many edge cases we'll have to check with this

spec/support/models/post.cr
jwoertink: That's what I was thinking too. I'd be ok with saying that the `through` can't have a `query`. Or at least to start.

jwoertink:
Thinking about some of the edge cases on this, I'm not sure what should happen here:

```crystal
# 🤔 
Post::BaseQuery.new.preload_nice_comments(Comment::BaseQuery.new.not_nice)
```

Does it un-preload the nice comments and re-preload the not nice comments? I'm torn on this change now. Maybe we just document how to handle preloading these sort of queries? Or do we just not allow overriding the preload on these?

matthewmcgarvey:
What about adding an overload to `preload_xxx` that takes a block? That way we could point people to use that unless they want to completely provide the query.

matthewmcgarvey (changes requested):
I believe this line needs to be updated as well https://github.com/luckyframework/avram/blob/bd5b7eef9368f72f2f109338dfa4380b16561f3c/src/avram/model.cr#L190

Without that line being updated, using the bang method to load a has many through where the "through" has a query specified would return invalid results

jwoertink:
Ok, a few things going on here. I wrote a spec to test the bang method queries with the code as is, and it all passed. Which means this change `{{ assoc[:type] }}::BaseQuery.new.{{ assoc[:foreign_key].id }}(id)` doesn't seem to be necessary. The confusing thing is that it *should* be.... `{{ assoc[:query] }}.{{ assoc[:foreign_key].id }}(id)`. I'm not sure why it passes with out it.

The next thing is that this brings up a few issues with all of the new preload overload methods. I mentioned one in my comment above, but really what would you expect there? Personally, I would expect Lucky to just completely tank at compile-time. But let's show an example of all of them.

```crystal
PostQuery.new.preload_nice_comments(not_nice_comment)

PostQuery.new.preload_nice_comments(not_nice_comment) do |nice_query|
  nice_query.not_nice
end

PostQuery.new.preload_nice_comments(not_nice_comment, CommentQuery.new.not_nice)

PostQuery.new.preload_nice_comments([not_nice_comment, nice_comment])

PostQuery.new.preload_nice_comments([not_nice_comment, nice_comment]) do |nice_query|
  nice_query.not_nice
end

PostQuery.new.preload_nice_comments([not_nice_comment, nice_comment], CommentQuery.new.not_nice)

PostQuery.new.preload_nice_comments(CommentQuery.new.not_nice)

PostQuery.new.preload_nice_comments do |nice_query|
  nice_query.not_nice
end
```

These are all preload methods that get generated currently. On the one hand, this gives you a way to override an existing query with your own query. On the other hand, it's weird that you're trying to preload a very specific set of records only to tell it to not do the one thing you wanted it to do.... 

@paulcsmith I'd love to get your input on this if you get a free moment.

jwoertink:
I keep coming back to this and thinking about it, but I'm still not convinced on it being completely necessary. The original concept for having it was so that you could have a custom association that could be preloaded. But I feel like preloading `[a,b]` and then returning `[c]` would be a weird case that shouldn't work.

I bet we could figure out a better way to do it, but it's not a high priority for before 1.0. For now I'm going to close this out.

Issue #408, opened 2020-06-26
Support for (virtual) array attributes


oneiros:
I tried to setup a `Avram::Operation` with an array attribute:

```crystal
class CreateRoles < Avram::Operation
  attribute role_names : Array(String)
end
```

Sadly this will not compile, because crystal does not currently support nested constants in generics. In #151 @jwoertink added array support for column values and found a nice workaround.

So as a first step, I tried to use this workaround and change

https://github.com/luckyframework/avram/blob/af999118e8353f2874c1d70aa6f6f5ca548d7be8/src/avram/define_attribute.cr#L95

to

```crystal
parse_result = {{ type }}.adapter.parse({{ name }}_param)
```

Then my app did compile successfully. I have not checked if it actually works yet, because I thought this would be a nice little fix I could prepare a PR for.

Sadly, I failed because when I tried to add specs, they blew up:

```
There was a problem expanding macro 'setter'

Code in src/avram/attribute.cr:4:3

 4 | setter :value
     ^
Called macro defined in macro 'macro_139913478760976'

 483 | macro setter(*names)

Which expanded to:

 > 1 |       
 > 2 |         
 > 3 |           def value=(@value)
                            ^----
Error: instance variable '@value' of Avram::Attribute(Array(String) | Nil) must be (Array(String) | Nil), not (String | Nil)
```

I _think_ the reason is that the specs use `Avram::Params` which only supports a flat `String => String` mapping. If my theory is correct, my app compiles, because it uses `Lucky::Params` which does support nested structures.

If someone knows an easy fix, I would be happy to continue the work on a PR.

But more importanty I think this could be valuable input for luckyframework/lucky#1209

wout:
I've been in the same place while testing file params. The way I worked around this was to create a subclass for `Avram::Params` just for testing purposes:

https://github.com/luckyframework/avram/pull/404/files#diff-dcf82d31916d537fe7e9380227ffc216

But that's obviously not an ideal solution. I think `Avram::Params` needs some work to bring it more in line with `Lucky::Params`.

matthewmcgarvey:
I'd like to refactor Avram to be able to remove `Avram::Paramable` and `Avram::Params`. They are rough stand-ins for `Lucky::Params`. This won't be usable until https://github.com/luckyframework/lucky/issues/1380 is finished but I'd like to cut the dependency and allow `Avram::Attribute` to be any type as long as it can extract the type from whatever object is passed in (like `Lucky::Params`)

[...]

PR #586, opened 2021-01-04
Update how generics work for Avram::Attribute

matthewmcgarvey:
I'm starting to look into https://github.com/luckyframework/avram/issues/408 but noticed that the generics for `Avram::Attribute` were a bit weird. We document that values for attributes are always nilable but that wasn't necessarily true. It's always nilable because where ever we were "new-ing" up an attribute we were doing so in a macro and adding in a `?` to whatever type was supplied but the attribute's type could have been anything (including not nilable). We can avoid doing that in the macros by making the attribute's value nilable within the class. So `Avram::Attribute(String?)` becomes `Avram::Attribute(String)` but `Avram::Attribute(String)#value` is still `String?`

This would be a breaking change for anyone referencing the type. For instance, some validations had to be updated so anyone that defined their own and specified the type will be broken.

### DIFF
--- a/src/avram/attribute.cr
+++ b/src/avram/attribute.cr
@@ -1,16 +1,12 @@
 class Avram::Attribute(T)
   alias ErrorMessage = String | Proc(String, String, String) | Avram::CallableErrorMessage
-  getter :name
-  setter :value
-  getter :param_key
+  getter name : Symbol
+  setter value : T?
+  getter param_key : String
   @errors = [] of String
+  @param : Avram::Uploadable | String?
 
-  def initialize(
-    @name : Symbol,
-    @param : Avram::Uploadable | String?,
-    @value : T,
-    @param_key : String
-  )
+  def initialize(@name, @param, @value : T?, @param_key)
     @original_value = @value
   end
 
@@ -18,7 +14,7 @@ class Avram::Attribute(T)
 
   def permitted
     @_permitted ||= begin
-      Avram::PermittedAttribute.new(name: @name, param: @param, value: @value, param_key: @param_key).tap do |attribute|
+      Avram::PermittedAttribute(T).new(name: @name, param: @param, value: @value, param_key: @param_key).tap do |attribute|
         errors.each do |error|
           attribute.add_error error
         end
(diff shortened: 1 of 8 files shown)


src/avram/attribute.cr
```diff
@@ -1,77 +1,69 @@
 class Avram::Attribute(T)
   alias ErrorMessage = String | Proc(String, String, String) | Avram::CallableErrorMessage
-  getter :name
-  setter :value
-  getter :param_key
+  getter name : Symbol
+  setter value : T?
+  getter param_key : String
```
matthewmcgarvey, line 5: I change symbol to type declaration out of preference, hopefully that's ok 😄

src/avram/attribute.cr
```diff
@@ -1,77 +1,69 @@
...
-  def add_error(message : ErrorMessage = "is invalid")
-    perform_add_error(message)
-  end
-
-  private def perform_add_error(message : String = "is invalid")
+  def add_error(message : String = "is invalid")
     @errors << message
   end
 
-  private def perform_add_error(message : (Proc | Avram::CallableErrorMessage))
```
matthewmcgarvey, line 41: Moving this overload into the `add_error` method simplifies the code and error stack traces

src/avram/attribute.cr
```diff
@@ -1,77 +1,69 @@
...
+  def original_value : T?
     ensure_no_blank(@original_value)
   end
 
-  private def ensure_no_blank(value : T)
+  private def ensure_no_blank(value : T?) : T?
     if value.is_a?(Avram::Uploadable | String) && value.blank?
       nil
     else
       value
```
jwoertink, line 58: I wonder if we should be adding a `not_nil!` here so the compiler knows this type is locked in? 🤔

jwoertink (approved):
I think this makes sense. We lock in the type the attribute wants, but say the value could be nil. 👍

src/avram/attribute.cr
matthewmcgarvey: The value can always be nil, this code is just the logic that also treats blank strings as nil.

PR #587, opened 2021-01-04
Use .adapter instead of using Lucky namespace for parsing

matthewmcgarvey:
Related to #408 

This brings several advantages

- We can potentially remove the Lucky namespace from Avram code
- Custom types only have to implement the method instead of the namespace
- There were bugs with generic classes where `Array(String)` was made into `Array(String)::Lucky` which doesn't work but the method does

This is a breaking change as anyone using custom types will need to add this adapter method if they didn't already.

This does not fix the related issue because the `Array.adapter` method delegates to the generic class which means there's no support for turning, for example, a string param into an array

https://github.com/luckyframework/avram/blob/03d1a7bcc981a75b103c2f54b81a430096ea0c84/src/avram/charms/array_extensions.cr#L2-L6

### DIFF
--- a/spec/support/custom_email.cr
+++ b/spec/support/custom_email.cr
@@ -1,4 +1,8 @@
 class CustomEmail
+  def self.adapter
+    Lucky
+  end
+
   def initialize(@email : String)
   end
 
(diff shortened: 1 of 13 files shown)


jwoertink (approved):
I love how much this cleans up the code. I think this was the original intension anyway. We wanted the `adapter` to allow 3rd party shards to hook in where needed eventually. This makes that process easier, I think.

PR #591, opened 2021-01-04
Add a criteria method to the type adapter

matthewmcgarvey:
No connected issue

After https://github.com/luckyframework/avram/pull/587, there was a still one place with a hard-coded reference to the type extension namespaces https://github.com/luckyframework/avram/blob/5f51bebce324a2b3a0aa3b0457b3799d567a3150/src/avram/base_query_template.cr#L14-L19

This adds a `criteria` method that takes in the query and the column name and returns an instance of an `Avram::Criteria` for the given type.

One of the other benefits it gives us is that we don't need special handling for generic types since we are calling methods on the class instead of using it to access a namespace.

This would be a breaking change since the `criteria` method is completely new and custom types would need to implement it.

The rules for adding custom types would be:

- The type must have an `adapter` class-level method that returns a class that includes the `Avram::Type` module
- That class must have a `criteria` class-level method that returns an instance of `Avram::Criteria` for the type

I considered moving the `criteria` method onto the type we extend but I didn't want to pollute the core library types with another function.

### DIFF
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -11,10 +11,9 @@ class Avram::BaseQueryTemplate
         include Avram::PrimaryKeyQueryable({{ type }})
       {% end %}
 
-      macro generate_criteria_method(query_class, name, type)
+      macro generate_criteria_method(name, type)
         def \{{ name }}
-          column_name = "#{table_name}.\{{ name }}"
-          \{{ type }}::Lucky::Criteria(\{{ query_class }}, \{{ type }}).new(self, column_name)
+          \{{ type }}.adapter.criteria(self, "#{table_name}.\{{ name }}")
         end
       end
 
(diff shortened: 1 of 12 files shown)


src/avram/base_query_template.cr
```diff
@@ -48,20 +47,11 @@ class Avram::BaseQueryTemplate
           {{ column[:name] }}.eq(value)
         end
 
-        {% if column[:type].is_a?(Generic) %}
-          # Checking Array type
-          generate_criteria_method(BaseQuery, {{ column[:name] }}, {{ column[:type].type_vars.first }})
+        generate_criteria_method({{ column[:name] }}, {{ column[:type] }})
```
jwoertink, line 50: I'm a little confused on why we don't need to check for `Array(String)` and it will know to use `String.adapter` 🤔

jwoertink (commented):
This is probably fine, just that one part confusing me a little

src/avram/base_query_template.cr
matthewmcgarvey: Given a type of `Array(String)` it will call `Array(String).adapter` and that method calls `.adapter` on the generic type of the array which in this case is `String.adapter`.

https://github.com/luckyframework/avram/blob/efe1d8afaf337809c2accf51300daac48dba1cdc/src/avram/charms/array_extensions.cr#L1-L7

jwoertink: ahh! boom! I missed that part.

jwoertink (approved):

PR #592, opened 2021-01-05
Added simple task to enter db console

### LINKED ISSUE #516: Add db.console task
I would like a `lucky db.console` task that just runs `psql postgres:///my_db`. I've had to hop in to psql a few times in the last few days, and that task would be much quicker and easier since Avram knows how to build out the URL already.

jwoertink:
Fixes #516

This task just calls `psql` passing in your configured DB URL. I have no idea how to test this 😂 

Screen Shot 2021-01-05 at 11 12 53 AM

### DIFF
--- a/src/avram/tasks/db/console.cr
+++ b/src/avram/tasks/db/console.cr
@@ -1,0 +1,42 @@
+class Db::Console < LuckyCli::Task
+  summary "Access PostgreSQL console"
+
+  def help_message
+    <<-TEXT
+    #{summary}
+
+    Enters the postgres REPL. Check config/database.cr
+    for database configuration.
+
+    Examples:
+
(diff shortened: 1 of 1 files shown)


jwoertink:
I want to make this task precompiled, but I need to test that locally first to see if it works.

jwoertink:
🤨  ok.... what?

Screen Shot 2021-01-05 at 11 32 35 AM


but how?

jwoertink:
Ok, it turns out that if you pre-compile the db tasks, then change the db settings, the pre-compiled tasks will use the old settings, and not the new changes. Lame 😒 ...For now we can't precompile this, but it would be amazing if we could figure out how to precompile all of those.

PR #596, opened 2021-01-08
Handle updates for nested `.has_one` save operations

akadusei:
Resolves https://github.com/luckyframework/avram/issues/7

This PR seeks to support updating for `.has_one`s. The previous implementation worked for creates, not for updates. So if you called `SomeOperation.update` for an operation that called `.has_one`, you would get an error.

### DIFF
--- a/src/avram/nested_save_operation.cr
+++ b/src/avram/nested_save_operation.cr
@@ -1,22 +1,45 @@
 module Avram::NestedSaveOperation
   macro has_one(type_declaration)
     {% name = type_declaration.var %}
-    {% type = type_declaration.type %}
-    @_{{ name }} : {{ type }} | Nil
+    {% type = type_declaration.type.resolve %}
 
-    def {{ name }}
-      @_{{ name }} ||= {{ type }}.new(params)
-    end
+    {% model_type = type.ancestors.find do |t|
+         t.stringify.starts_with?("Avram::SaveOperation(")
+       end.type_vars.first %}
 
-    after_save save_nested_{{ name }}
+    {% assoc = T.constant(:ASSOCIATIONS).find do |assoc|
+         assoc[:relationship_type] == :has_one &&
+           assoc[:type].resolve.name == model_type.name
+       end %}
 
-    def save_nested_{{ name }}(record)
-      {{ name }}.{{ @type.constant(:FOREIGN_KEY).id }}.value = record.id
+    {% unless assoc %}
+      {% raise "#{T} must have a has_one association with #{model_type}" %}
+    {% end %}
 
-      if !{{ name }}.save
+    after_save save_{{ name }}
+
+    def save_{{ name }}(saved_record)
+      unless {{ name }}.save
+        add_error(:{{ name }}, "failed")
         mark_nested_save_operations_as_failed
         database.rollback
       end
+    end
+
+    def {{ name }}
+      @{{ name }} ||= if new_record?
+        {{ type }}.new(params)
+      else
+        {{ type }}.new(record.not_nil!.{{ assoc[:assoc_name].id }}!, params)
+      end
+
+      nested = @{{ name }}.not_nil!
+
+      record.try do |record|
+        nested.{{ @type.constant(:FOREIGN_KEY).id }}.value = record.id
+      end
+
+      nested
     end
 
     def nested_save_operations
@@ -28,8 +51,8 @@ module Avram::NestedSaveOperation
   end
 
   def mark_nested_save_operations_as_failed
-    nested_save_operations.each do |f|
-      f.as(Avram::MarkAsFailed).mark_as_failed
+    nested_save_operations.each do |operation|
+      operation.as(Avram::MarkAsFailed).mark_as_failed
     end
   end
 
--- a/src/avram/nested_save_operation.cr
+++ b/src/avram/nested_save_operation.cr
@@ -1,79 +1,52 @@
 module Avram::NestedSaveOperation
-  macro included
-    macro inherited
-      define_has_one
-    end
-  end
+  macro has_one(type_declaration)
+    {% name = type_declaration.var %}
+    {% type = type_declaration.type.resolve %}
 
-  macro define_has_one
-    macro has_one(type_declaration)
-      \{% name = type_declaration.var %}
-      \{% type = type_declaration.type.resolve %}
+    {% model_type = type.ancestors.find do |t|
+         t.stringify.starts_with?("Avram::SaveOperation(")
+       end.type_vars.first %}
 
-      \{% model_type = type.ancestors.find do |t|
-          t.stringify.starts_with?("Avram::SaveOperation(")
-        end.type_vars.first %}
+    {% assoc = T.constant(:ASSOCIATIONS).find do |assoc|
+         assoc[:relationship_type] == :has_one &&
+           assoc[:type].resolve.name == model_type.name
+       end %}
 
-      \{% assoc = T.constant(:ASSOCIATIONS).find do |assoc|
-          assoc[:relationship_type] == :has_one &&
-            assoc[:type].resolve.name == model_type.name
-        end %}
+    {% unless assoc %}
+      {% raise "#{T} must have a has_one association with #{model_type}" %}
+    {% end %}
 
-      \{% if ! assoc %}
-        \{% raise "#{T} must have a has_one association with #{model_type}" %}
-      \{% end %}
+    after_save save_{{ name }}
 
-      \{% for need in type.constant(:OPERATION_NEEDS) %}
-        needs \{{ name }}_\{{ need }}
-      \{% end %}
-
-      after_save save_\{{ name }}
-
-      def save_\{{ name }}(saved_record)
-        unless \{{ name }}.save
-          add_error(:\{{ name }}, "failed")
-          mark_nested_save_operations_as_failed
-          database.rollback
-        end
-      end
-
-      def \{{ name }}
-        @\{{ name }} ||= if new_record?
-          \{% begin %}\{{ type }}.new(
-            params,
-            \{% for need in type.constant(:OPERATION_NEEDS) %}
-              \{{ need.var }}: \{{ name }}_\{{ need.var }},
-            \{% end %}
-          )\{% end %}
-        else
-          \{% begin %}\{{ type }}.new(
-            record.not_nil!.\{{ assoc[:assoc_name].id }}!,
-            params,
-            \{% for need in type.constant(:OPERATION_NEEDS) %}
-              \{{ need.var }}: \{{ name }}_\{{ need.var }},
-            \{% end %}
-          )\{% end %}
-        end
-
-        nested = @\{{ name }}.not_nil!
-
-        record.try do |record|
-          nested.\{{ @type.constant(:FOREIGN_KEY).id }}.value = record.id
-        end
-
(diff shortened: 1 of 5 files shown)


matthewmcgarvey (commented):
Sorry for the slow reply, I've been thinking about this though. While the operation method to call for these nested operations is "has_one" I'm not sure if it is right to assume that it is necessarily connected to a model's `SaveOperation` and set foreign keys on it. I view that as a concern of the nested operation to manage that. 

This would also allow you to avoid digging into the type of the operation to pull out the model. Though, admittedly, I'm not really clear on what this change is doing

(I just noticed that the old code was setting the foreign key as well, I'd be open to that changing)

akadusei:
> Though, admittedly, I'm not really clear on what this change is doing

This PR seeks to support updating for `.has_one`s. The previous implementation worked for creates, not for updates. So if you called `SomeOperation.update` for an operation that calls `.has_one`, you would get an error.

> Sorry for the slow reply, I've been thinking about this though. While the operation method to call for these nested operations is "has_one" I'm not sure if it is right to assume that it is necessarily connected to a model's SaveOperation and set foreign keys on it. I view that as a concern of the nested operation to manage that.

That's OK. My understanding is that `.has_one` is for automatically saving a model that has an actual `has_one` association with another model, such that saving one necessitates the saving of the other. (At least, that's how I have always used it).

I'm not  sure if this can be expanded to a general use though; It could be worth the try.

jwoertink:
> My understanding is that .has_one is for automatically saving a model that has an actual has_one association

That's correct. The `Avram::Operation` won't have a `has_one` method. There's a few things this needs to do to update. Like it should probably raise a compile-time error if the associated SaveOperation is unrelated. But things like that can come later. I had planned on working on it, but work has been crazy lately.

I haven't had a chance to really look over this, but if the idea is to make it work for both creates and updates, then I think that's a good step.

akadusei:
> Like it should probably raise a compile-time error if the associated SaveOperation is unrelated.

A very good point. I'll try to work on that. If it proves cumbersome, I'll just leave for a later PR.

src/avram/nested_save_operation.cr
```diff
@@ -1,35 +1,61 @@
...
+
     def nested_save_operations
       {% if @type.methods.map(&.name).includes?(:nested_save_operations.id) %}
         previous_def +
       {% end %}
       [{{ name }}]
     end
+
+    private def create_{{ name }}
+      @{{ name }} ||= {{ type }}.new(params)
```
matthewmcgarvey, line 45: Let's move this memoization back up to the `def {{ name }}` and out of the create and update methods. You can memoize with multiple actions on the right side by doing
```crystal
@{{ name }} ||= begin
  # ...
end
```

jwoertink:
I haven't forgot about this, still processing some stuff around it. I just implemented some operations using `has_one` only for update, and I don't get any errors like how you mention in the description. I do, however, get a different issue. If your operation is successful, but the nested save operation fails during an update, there's nothing that bubbles the errors back up.

```crystal
   has_one account : SaveAccount
end

SaveUser.update(user, params) do |op, saved_user|
  # if the `SaveAccount` fails, these are still true
  # op.valid? == true
  # saved_user != nil
end
```

To get around that,  you'd have to also look at `op.account.valid?` and `op.account.errors`. Maybe that's fine for 1 association, but if you had let's say 4 of them, then it starts to get super messy. 

I'll need to come back to review once I have a free moment to think on this, but my initial thought is that we may not need a whole lot of changes to make all this work better. Thanks!

src/avram/nested_save_operation.cr
```diff
@@ -1,35 +1,61 @@
...
+    {% type = type_declaration.type.resolve %}
+    {% model_type = type.superclass.superclass.type_vars.first.resolve %}
 
-    def {{ name }}
-      @_{{ name }} ||= {{ type }}.new(params)
-    end
+    {% assoc = T.resolve.constant(:ASSOCIATIONS).find do |assoc|
+         assoc[:relationship_type] == :has_one &&
+           assoc[:type].resolve.name == model_type.name
+       end %}
```
matthewmcgarvey, line 10: I'm worried about this section. I'm fairly certain if you call `.resolve` on a class from a file that hasn't been required yet it will fail. That in itself is not terrible it's just something that we've/I've tried to avoid.

That being said, calling `.superclass.superclass` is understandable but I think we want to avoid it just in case someone subclasses a subsclass of an operation.  What I **think** you can do is call `type.ancestors[-3]` instead

```crystal
class Foo
end

class Bar < Foo
end

{{ Bar.ancestors }} # [Foo, Reference, Object]
{{ Bar.ancestors[-3] }} # Foo
```
Not entirely sure but I think that's what you want.

I just wish this could be solved without digging so far into all these classes 😬

akadusei:
> \# saved_user != nil

This is as expected for `.update`, which yields a user whether or not operation succeeded. `.create` on the other hand, may yield a user or `nil`.

> \# op.valid? == true

Yup, missed that. I should add that to the specs. Would something like this suffice?:

```crystal
def save_{{ name }}(saved_record)
  unless {{ name }}.save
    add_error(:nested_{{ name }}, "failed") # <===
    mark_nested_save_operations_as_failed
    database.rollback
  end
end
```

src/avram/nested_save_operation.cr
akadusei: Sure. Will work on it.

src/avram/nested_save_operation.cr
akadusei: > What I think you can do is call `type.ancestors[-3]` instead

I like this better. I'll give it a shot.

> I'm fairly certain if you call `.resolve` on a class from a file that hasn't been required yet it will fail.

I'm not sure what else is expected. I'll try doing away with `.resolve` and compare strings instead. We'll see.

akadusei: Doesn't work.

This:

```crystal
{% puts type.ancestors %}
```
...produces:
```bash
[EmailAddress::SaveOperation, Avram::SaveOperation(EmailAddress), Avram::InheritColumnAttributes, Avram::MarkAsFailed, Avram::NestedSaveOperation, Avram::DatabaseValidations(EmailAddress), Avram::Callbacks, Avram::NeedyInitializerAndSaveMethods, Avram::ParamKeyOverride, Avram::OperationErrors, Avram::Validations, Avram::DefineAttribute, Reference, Object, Spec::ObjectExtensions, Colorize::ObjectExtensions]
```

It seems all included modules become part of the type hierarchy in a way that makes it unreliable.

akadusei: I got it. This works:

```crystal
{% model_type = type.ancestors.find do |t|
  t.stringify.starts_with?("Avram::SaveOperation(")
end.type_vars.first.resolve %}
```

matthewmcgarvey (approved):
I think I'm ok with this getting merged if @jwoertink also signs off.

Good job on this @akadusei 👍

akadusei:
Looking at #276, should we add error to all nested operations if parent operation failed? I'm thinking this:

```crystal
  def mark_nested_save_operations_as_failed
    nested_save_operations.each do |operation|
      operation.add_error(:parent, "failed") # <<<===
      operation.as(Avram::MarkAsFailed).mark_as_failed
    end
  end
```

akadusei:
Wait! We did not factor in operation needs for nested operations. If a nested operation has needs, we get a compile error. This is getting out of hand. :rofl: 

I'll see what I can come up with.

jwoertink:
> This is getting out of hand. 🤣

yup! There's quite a bit to unpack. I think I have some free time tomorrow to take a look at this. I have some ideas on how to do this.

src/avram/nested_save_operation.cr
```diff
@@ -1,52 +1,79 @@
...
+        \{% raise "#{T} must have a has_one association with #{model_type}" %}
+      \{% end %}
 
-    def {{ name }}
-      @{{ name }} ||= if new_record?
-        {{ type }}.new(params)
-      else
-        {{ type }}.new(record.not_nil!.{{ assoc[:assoc_name].id }}!, params)
+      \{% for need in type.constant(:OPERATION_NEEDS) %}
+        needs \{{ name }}_\{{ need }}
```
akadusei, line 27: Not sure about this, but needed to forward the needs to the nested operation somehow.

akadusei:
> I think I have some free time tomorrow to take a look at this. I have some ideas on how to do this.

Great! I just pushed a commit, but still interested in your ideas.

src/avram/nested_save_operation.cr
jwoertink: I think we should back out these changes. It'll be something broken, but we can open up an issue and tackle it for a future PR. My thought is you'd pass it in like `SaveBusiness.create(save_tax_id: {...})`, but there's some other complicated things that need to be figured out first.

src/avram/nested_save_operation.cr
```diff
@@ -1,35 +1,85 @@
...
+              \{{ need.var }}: \{{ name }}_\{{ need.var }},
+            \{% end %}
+          )\{% end %}
+        end
+
+        nested = @\{{ name }}.not_nil!
+
+        record.try do |record|
+          nested.\{{ @type.constant(:FOREIGN_KEY).id }}.value = record.id
+        end
```
jwoertink, line 62: This should probably happen in the `after_save` call. You may call the association during some before_saves and it keeps trying to make this assignment on an accessor method.

src/avram/nested_save_operation.cr
```diff
@@ -1,35 +1,85 @@
...
+    macro inherited
+      define_has_one
     end
+  end
+
+  macro define_has_one
+    macro has_one(type_declaration)
+      \{% name = type_declaration.var %}
+      \{% type = type_declaration.type.resolve %}
+
```
jwoertink, line 12: The `@_{{ name }} : {{ type }} | Nil` can be added back in here.

src/avram/nested_save_operation.cr
```diff
@@ -1,35 +1,85 @@
...
+      \{% end %}
 
-    after_save save_nested_{{ name }}
+      after_save save_\{{ name }}
 
-    def save_nested_{{ name }}(record)
-      {{ name }}.{{ @type.constant(:FOREIGN_KEY).id }}.value = record.id
+      def save_\{{ name }}(saved_record)
+        unless \{{ name }}.save
+          add_error(:\{{ name }}, "failed")
```
jwoertink, line 34: This is fine to start. In a separate PR we will want to actually bubble up all of the errors. We can just open an issue for now.

jwoertink (changes requested):
Thinking about this more, there's a ton of things that need to happen, but I think this PR is already really big. I'd like to start by taking out the addition of accounting for `needs`, then file an issue about adding that in later. There's a few other things like `attribute` and until we remove it, `file_attribute` that also have to be accounted for. 

Overall, I think this is doing some good things that will help, but I left a few requests to change. Once those are updated, I'll give it one last look over, and we can start thinking about the next iteration on making this better. Thanks for tackling this!

jwoertink (approved):
Awesome! Yeah, this update feels a lot better. I know we still have a ton of stuff missing, but doing this in small steps definitely makes it easier to digest. Thanks for handling this and sorry for taking so long. I'll get this merged in, and we can start adding in the issues for what else needs to be done.

PR #600, opened 2021-01-09
Add `#revert` for save operations

akadusei:
Reverts a saved record to its initial state -- a newly-created record is deleted, and an updated existing record is reversed to its original state.

Among other things, this is needed to complete https://github.com/luckyframework/avram/pull/596. When an operation contains multiple nested operations (via `.has_one` macro), this should help us revert already-saved nested operations if one nested operation fails.

Currently, when a nested operation fails, the parent operation marks all nested operations as failed. But any nested operation that successfully saved before the said failure could have its changes intact in the database in particular instances.

After this change, #596 is as simple as:

```crystal
  # This method would be renamed
  def mark_nested_save_operations_as_failed
    nested_save_operations.each do |operation|
      operation.revert # <= Magic happens here
      operation.as(Avram::MarkAsFailed).mark_as_failed
    end
  end
```

I've tested this locally, and it works. All specs pass.

### DIFF
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -20,6 +20,7 @@ abstract class Avram::SaveOperation(T)
   include Avram::NestedSaveOperation
   include Avram::MarkAsFailed
   include Avram::InheritColumnAttributes
+  include Avram::RevertSaveOperation
 
   enum SaveStatus
     Saved
@@ -346,51 +347,6 @@ abstract class Avram::SaveOperation(T)
   # for an update, independent of the stage we are at in the operation.
   def new_record? : Bool
     {{ T.resolve.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
-  end
-
-  # TODO:
-  #   - Factor in operation needs
-  #   - Use the new delete operation for deleting the record
-  def revert : self
-    if persisted?
-      operation = self.class.new(record.not_nil!)
-    else
-      operation = self.class.new
-    end
-
-    unless saved?
-      operation.add_error(:revert, "Cannot revert an unsaved record")
-      return operation
-    end
-
-    if new_record?
-      revert_create(operation)
-    else
-      revert_update(operation)
-    end
-
-    operation
-  end
-
-  private def revert_create(operation)
-    if record.not_nil!.delete.rows_affected < 1
-      operation.add_error(:revert, "Could not delete record")
-    end
-  end
-
-  private def revert_update(operation)
-    {% for attribute in @type.constant(:ATTRIBUTES) %}
-      operation.{{ attribute.var }}.value = {{ attribute.var }}.original_value
-    {% end %}
-
-    {% for column in @type.constant(:COLUMN_ATTRIBUTES) %}
-      operation.{{ column[:name].id }}.value =
-        {{ column[:name].id }}.original_value
-    {% end %}
-
-    unless operation.save
-      operation.add_error(:revert, "Could not save record")
-    end
   end
 
   private def insert_or_update
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -339,13 +339,22 @@ abstract class Avram::SaveOperation(T)
     !!record_id
   end
 
+  # `#persisted?` always returns `true` in `after_*` hooks, whether
+  # a new record was created, or an existing one was updated.
+  #
+  # This method should always return `true` for a create or `false`
+  # for an update, independent of the stage we are at in the operation.
+  def new_record? : Bool
+    {{ T.resolve.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
+  end
+
   def revert : self?
     return unless saved?
 
     saved_record = record.not_nil!
     operation = self.class.new(saved_record)
 
-    if id.value.nil?
+    if new_record?
       operation if saved_record.delete.rows_affected > 0
     else
       {% for attribute in @type.constant(:ATTRIBUTES) %}
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -348,25 +348,48 @@ abstract class Avram::SaveOperation(T)
     {{ T.resolve.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
   end
 
-  def revert : self?
-    return unless saved?
+  # TODO:
+  #   - Factor in operation needs
+  #   - Use the new delete operation for deleting the record
+  def revert : self
+    if persisted?
+      operation = self.class.new(record.not_nil!)
+    else
+      operation = self.class.new
+    end
 
-    saved_record = record.not_nil!
-    operation = self.class.new(saved_record)
+    unless saved?
+      operation.add_error(:revert, "Cannot revert an unsaved record")
+      return operation
+    end
 
     if new_record?
-      operation if saved_record.delete.rows_affected > 0
+      revert_create(operation)
     else
-      {% for attribute in @type.constant(:ATTRIBUTES) %}
-        operation.{{ attribute.var }}.value = {{ attribute.var }}.original_value
-      {% end %}
+      revert_update(operation)
+    end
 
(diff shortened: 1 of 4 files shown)


src/avram/save_operation.cr
```diff
@@ -339,6 +339,37 @@ abstract class Avram::SaveOperation(T)
...
+  # `#persisted?` always returns `true` in `after_*` hooks, whether
+  # a new record was created, or an existing one was updated.
+  #
+  # This method should always return `true` for a create or `false`
+  # for an update, independent of the stage we are at in the operation.
+  def new_record? : Bool
+    {{ T.resolve.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
+  end
+
+  def revert : self?
```
wontruefree, line 351: should another value besides `nil` be returned?

src/avram/save_operation.cr
```diff
@@ -339,6 +339,37 @@ abstract class Avram::SaveOperation(T)
     !!record_id
   end
 
+  # `#persisted?` always returns `true` in `after_*` hooks, whether
+  # a new record was created, or an existing one was updated.
+  #
+  # This method should always return `true` for a create or `false`
+  # for an update, independent of the stage we are at in the operation.
+  def new_record? : Bool
+    {{ T.resolve.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
```
jwoertink, line 348: This essentially relies on the fact that we're not currently updating the attributes on an operation after updating a record. We should add a test for this method just in case we decide to change how attributes are handled.

src/avram/save_operation.cr
jwoertink: hmm... Yeah, I'm not sure about the `self?` return here 🤔  It's really returning either a new operation instance, or nothing as opposed to the current operation instance or nil. Maybe you want to know why a revert failed. We could have this either return a new operation if it succeeded, or return the old operation if it failed, and add to the `errors` object using `add_error(:revert, "Record must be saved before reverting")` or `add_error(:revert, "Could not delete record")`

src/avram/save_operation.cr
```diff
@@ -339,6 +339,37 @@ abstract class Avram::SaveOperation(T)
...
+  end
+
+  def revert : self?
+    return unless saved?
+
+    saved_record = record.not_nil!
+    operation = self.class.new(saved_record)
+
+    if new_record?
+      operation if saved_record.delete.rows_affected > 0
```
jwoertink, line 358: What happens if the `saved_record` is SoftDeletable? Also, all record deleting should start moving in to the new `DeleteOperation`.

src/avram/save_operation.cr
```diff
@@ -339,6 +339,37 @@ abstract class Avram::SaveOperation(T)
...
+  # for an update, independent of the stage we are at in the operation.
+  def new_record? : Bool
+    {{ T.resolve.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
+  end
+
+  def revert : self?
+    return unless saved?
+
+    saved_record = record.not_nil!
+    operation = self.class.new(saved_record)
```
jwoertink, line 355: If the SaveOperation has specific `needs`, then the new operation won't get them. When it calls `save` below, it'll run all of the callbacks which may end up failing.

jwoertink (changes requested):
I like the concept, and I can see how this may be useful, especially for the other PR that was started. My main hesitation with this current implementation is that there are a lot of edge cases to consider. I've notated a few of the cases in the comments, but I'd also like to give @matthewmcgarvey a chance to review before we decide on the best way to move forward.

This is a great start though! Thanks for submitting 👍

src/avram/save_operation.cr
wontruefree: I like the idea of adding an `error` to the retruned record and returning the record.  it means you have to check the record for failures but it mean you can add more context to the reverted record.

akadusei: > Yeah, I'm not sure about the self?

I wasn't sure about it either. I just needed a way to represent a failed revert. `#add_error` seems a good fit, although I am not sure mutating the current operation is a good idea. if a revert failed, you may end up with `operation.saved?` being `true` while `operation.valid?` would be `false` for the same operation. I don't know if that would be OK?

> It's really returning either a new operation instance, or nothing as opposed to the current operation instance or nil

We already have the current operation, on which you are calling `#revert`, so returning the same operation may be redundant.

> I like the idea of adding an error to the retruned record and returning the record

Sounds good. It means you can call `operation.revert.valid?` to check if revert succeeded, while maintaining the state of the current operation.

src/avram/save_operation.cr
akadusei: Sure, makes a lotta sense.

src/avram/save_operation.cr
akadusei: Good catch! Totally missed that part. Will see what I can come up with.

src/avram/save_operation.cr
wontruefree: I dont know if this work can be done in a follow up PR but just as an idea we could mimic the pattern Operations have.

This is what a save operation is like.
```crystal
SaveUser.create(params) do |operation, user|
  if user # the user was saved
    html Users::ShowPage, user: user
  else
    html Users::NewPage, save_user: operation
  end
end
```

Some api for reverting
```crystal
MyModel.revert do |operation, user|
  if user # the user was reverted
    successful_reverted
  else
    unsuccessful_reverted
  end
end
```

This might be overly complex for revert.

src/avram/save_operation.cr
akadusei: > What happens if the `saved_record` is SoftDeletable?

It is hard-deleted, since we're reverting the creation of that record. I'm not very familiar with soft deletes, though.

> Also, all record deleting should start moving in to the new `DeleteOperation`.

Not familiar with `DeleteOperation` either. I'll look into it, but I may need help.

src/avram/save_operation.cr
akadusei: It seems impossible to call revert on models, since a model has no knowledge of it's previous state. An operation, on the other hand, does; it holds previous values in the form of `#original_value`s. Or maybe I'm missing the point?

matthewmcgarvey:
@akadusei Can you update the description to explain how this would be used? My assumption is that you intend to call `#revert` on an operation if a nested operation fails? It was my assumption that the whole thing is wrapped in a transaction so exiting the transaction undoes the changes, right?

@jwoertink probably need your insight here

akadusei:
I've made some changes:

- Added specs for `#new_record?`
- `#revert` is returning the new operation, with errors forwarded to that operation.
- I add a `#revert!` method that raises `Avram::InvalidOperationError` if failed.
- Added support for operation needs.

I'm still not sure how the record can be deleted with the new `DeleteOperation`.

akadusei:
> @akadusei Can you update the description to explain how this would be used?

Done. Kindly let me know if the description still needs improvement.

akadusei:
> I thought that the whole thing is wrapped in a transaction so exiting the transaction undoes the changes, right?

Based on my tests, it would seem so. Except for when the parent operation has no attributes changed. In that particular instance, any updated nested record remains intact.

akadusei:
Continue the review, but please do not merge yet. I'll explain why in a new issue, soon.

wontruefree:
I am really liking where this is going.  Can we have a spec for the example in the description?

akadusei:
> I am really liking where this is going. Can we have a spec for the example in the description?

That is already covered in #596, where the example was taken from.

jwoertink:
@akadusei Now that #612 is merged in, do you think this method is still useful? My guess is we no longer need it. What are your thoughts?

akadusei:
> @akadusei Now that #612 is merged in, do you think this method is still useful? My guess is we no longer need it. What are your thoughts?

I can't think of a use case now. I'll close this. We can always get back to it if we should find it necessary in future.

PR #603, opened 2021-01-12
Support datetime-local time parsing

### LINKED ISSUE #602: Lucky datetime input not persisting time element
**Describe the bug**
When adding a `datetime` input to a form and setting the **time** component (either via the interface or manually modifying the `value` attribute of the HTML element, the date element seems to persist through to the DB while the time is ignored.

**To Reproduce**

1. Add a datetime input to any Lucky form
2. Set the datetime **time** value
3. Note that the time does not persist

**Expected behavior**

Users will expect that both the date and time of a `datetime` input are persisted to the DB

**Screenshots/code**

I have a form field that looks like this:

```crystal
    mount Shared::Field, operation.published_at, &.datetime_input(attrs: [:required])
```

That `published_at` field is of type `Time`

To see what was going on, I added a `before_save` hook and printed out the `published_at` `Avram::PermittedAttribute`:

```
web          | #<Avram::PermittedAttribute(Time | Nil):0x108666a00
web          |  @_permitted=nil,
web          |  @errors=[],
web          |  @name=:published_at,
web          |  @original_value=2021-01-16 00:00:00.0 UTC,
web          |  @param="2021-01-16T04:27",
web          |  @param_key="video",
web          |  @value=2021-01-16 00:00:00.0 UTC>
```

Note that while the `@param` value correctly registered the time, the `@value` was set without the time.

**Versions (please complete the following information):**
 - Lucky version (check in shard.lock): 0.25.0
 - Crystal version (`crystal --version`): 0.35.1
 - OS: Mac OSX Big Sur

stephendolan:
This PR intends to close #602.

You can read more about DateTime Local inputs at this link:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

From what I can gather, browsers basically use [RFC 3339](https://tools.ietf.org/html/rfc3339) without the timezone. Because of this, our `Time::Format::RFC_3339` parser wasn't actually extracting the time element from a `datetime-local` input type in Lucky.

### PR Components

- [TimeExtensions#parse] Add a custom `Time::Format` that will parse these `datetime-local` inputs correctly, though it _will_ always use `Time::Location.local` for the timezone.
- [Specs] Moved the hardcoded strings to use time formatting where possible so that we can test more date/time ranges and check specific time values more easily. 
- [Specs] Add some checks on specific parsed time components so that we can be more confident in the future that we're not missing components of the time.
- [Specs] Removed a test scenario for `space_separated_yaml`. Best I can tell, that was added [here](https://github.com/luckyframework/avram/pull/146) without any additional details, but it has never actually worked since we didn't have a formatter that matched down to the time element. Without a clear user story, I didn't feel confident adding a custom parser that would allow the new test validations to pass for that format.

### DIFF
--- a/spec/type_extensions/time_spec.cr
+++ b/spec/type_extensions/time_spec.cr
@@ -5,15 +5,26 @@ describe "Time column type" do
     it "casts various formats successfully" do
       time = Time.local
       times = {
-        iso8601:              time.to_s("%FT%X%z"),
-        rfc2822:              time.to_rfc2822,
-        rfc3339:              time.to_rfc3339,
-        space_separated_yaml: "2001-12-14 21:59:43.10 -5",
-        http_date:            "Sun, 14 Feb 2016 21:00:00 GMT",
+        iso8601:             time.to_s("%FT%X%z"),
+        rfc2822:             time.to_rfc2822,
+        rfc3339:             time.to_rfc3339,
+        datetime_html_input: time.to_s("%Y-%m-%dT%H:%M:%S"),
+        http_date:           time.to_s("%a, %d %b %Y %H:%M:%S GMT"),
       }
       times.each do |_format, item|
         result = Time.adapter.parse(item)
         result.should be_a(Avram::Type::SuccessfulCast(Time))
+
+        unless result.is_a? Avram::Type::SuccessfulCast(Time)
+          next
+        end
+
+        result.value.year.should eq(time.year)
+        result.value.month.should eq(time.month)
+        result.value.day.should eq(time.day)
+        result.value.hour.should eq(time.hour)
+        result.value.minute.should eq(time.minute)
+        result.value.second.should eq(time.second)
       end
     end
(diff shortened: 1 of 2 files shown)


matthewmcgarvey (approved):
Nice work on this 🎉🎉🎉

jwoertink:
I have to dig in more later, but this PR breaks specs for me locally:

```
Failures:

  1) Time column type .parse casts various formats successfully
     Failure/Error: "#{result.value.hour} #{_format}".should eq("#{time.hour} #{_format}")

       Expected: "8 iso8601"
            got: "16 iso8601"

     # spec/type_extensions/time_spec.cr:25
```

I added some debugging to the spec to see which time format is breaking. I don't want to open an issue just yet since I'm not sure if this is only me, but I'm running these locally on my mac (Catalina 10.15.7), and not in Docker. Also, I'm in PST timezone.

PR #608, opened 2021-01-13
Add support for case insensitive text columns (citext)

### LINKED ISSUE #550: Support citext column type
Sometimes a case insensitive text column type might be useful : https://github.com/luckyframework/authentic/pull/51

[citext](https://www.postgresql.org/docs/current/citext.html) is exactly that.

I was thinking about a `CIString` or [Citext](https://github.com/anykeyh/clear/blob/b98e23311eb3ef9086958e25e8e87f6789cc765d/src/clear/extensions/citext/citext.cr#L3) crystal type to map a citext column, but maybe an option (`case_sensitive: false`) could be better.

matthewmcgarvey:
Fixes #550 

## Usage

In a migration you'd pass the the `case_sensitive: false` flag to a string column to make its type `citext`. The [citext extension](https://www.postgresql.org/docs/current/citext.html) must be enabled.

```crystal
enable_extension "citext"

create table_for(User) do
  add name : String
  add email : String, case_sensitive: false
end
```

This would create the `users` table with a case sensitive `name` column and a case insensitive `email` column.

### DIFF
--- a/db/migrations/20180612221318_create_businesses.cr
+++ b/db/migrations/20180612221318_create_businesses.cr
@@ -13,10 +13,12 @@ class CreateBusinesses::V20180612221318 < Avram::Migrator::Migration::V1
       add_belongs_to business : Business, on_delete: :cascade
     end
 
+    enable_extension "citext"
+
     create :email_addresses do
       primary_key id : Int64
       add_timestamps
-      add address : String
+      add address : String, case_sensitive: false
       add_belongs_to business : Business?, on_delete: :cascade
     end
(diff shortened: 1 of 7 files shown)


jwoertink (approved):
Wow, this is pretty awesome. I had never even heard of this before. I kinda want to switch my columns over to it. I assume this works on `alter` table as well, right?

matthewmcgarvey:
Yep, pretty sure we splat any extra options onto the type so it will get passed along. 🎉

PR #609, opened 2021-01-14
Add support for Array(UUID)

### LINKED ISSUE #589: no overload matches 'UUID::Lucky.to_db' with type Array(UUID)
Looks like we missed this overload.

It should go after this https://github.com/luckyframework/avram/blob/5f51bebce324a2b3a0aa3b0457b3799d567a3150/src/avram/charms/uuid_extensions.cr#L24-L26

and possibly just copy the String one https://github.com/luckyframework/avram/blob/5f51bebce324a2b3a0aa3b0457b3799d567a3150/src/avram/charms/string_extensions.cr#L22-L24

jwoertink:
Fixes #589


matthewmcgarvey (approved):
Very nice 👍

PR #612, opened 2021-01-14
Run after_save and after_commit even if no changes were made

### LINKED ISSUE #604: Make the trip to the database always in save operations
Currently, *Avram* does not make the trip to the database if no columns changed. I assume this is for performance?

While working on #596, I encountered a situation where some of the nested save operations are never rolled back. As @matthewmcgarvey rightly pointed out, "... the whole thing is wrapped in a transaction so exiting the transaction undoes the changes, right?"

This is true, except for when the parent operation has no columns changed. In that particular instance, any updated nested record remains intact. When no columns changed, nested operations run, outside of any transaction, here:

https://github.com/luckyframework/avram/blob/84e431c6dfde0078e29dc0ad8f603a2082b8c917/src/avram/save_operation.cr#L316-L320

The essence of #600 is to revert any nested saves that could not be rolled back automatically. From my tests done locally, it works.

However, I believe we are trying to solve a problem that should not exist at all. Just let the save happen whether or not columns change. I'm talking about changing `#save` to something like this:

```crystal
  def save : Bool
    before_save

    if valid?
      transaction_committed = database.transaction do
        insert_or_update
        saved_record = record.not_nil!
        after_save(saved_record)
        true
      end

      if transaction_committed
        saved_record = record.not_nil!
        after_commit(saved_record)
        Avram::Events::SaveSuccessEvent.publish(
          operation_class: self.class.name,
          attributes: generic_attributes
        )
        true
      else
        mark_as_failed
        false
      end
    else
      mark_as_failed
      false
    end
  end
```

This would solve a lot of problems, including:

- `updated_at` column would always be updated. If I called `SavePost.update`, I would expect `updated_at` to be updated, even if no other column changed. Currently, it doesn't.
- `after_save` would always be guaranteed to run for valid operations.
- We might not need `#revert` (#600). At least, nested `.has_one` operations would work without it.
- We would not need `after_completed`

The current implementation creates API inconsistencies between creates and updates, and introduces unnecessary cognitive overhead for updates:

- `after_save` always runs for create operations. It is not guaranteed to run in update operations.
- `after_commit` always runs for create operations if transaction committed successfully. This renders `after_completed` redundant in create operations.
- In update operations, where you need the nested operations to run even if no column changed on the parent, `after_completed` seems like a good fit. Except you may want to roll back if a nested operation fails -- roll backs happen only in `after_save`. In such situations, you may call your nested operations in `after_save` and account for when `after_save` may not run in `after_completed`:

  ```crystal
  after_save my_nested_operation

  after_completed do |saved_record|
    my_nested_operation(saved_record) if changes.empty?
  end
  ```

From an API consumers perspective, there is too much internal implementation details I have to worry about, especially regarding updates.

Remove the need to avoid the database trip, and all these issues go away.

There's still a lot I do not know about *Avram*, so maybe my proposal has issues I have not thought about? I would be happy to know.

akadusei:
Closes #604

This PR ensures all nested operations called in `after_save` run, at all times, and in a transaction. It is, therefore, automatically rolled back if something goes wrong. Other potential benefits are discussed in the linked issue.

With this, we no longer need #600 for completing #596.

### DIFF
--- a/src/avram/callbacks.cr
+++ b/src/avram/callbacks.cr
@@ -525,8 +525,7 @@ module Avram::Callbacks
     {% raise <<-ERROR
       after_completed has been removed
 
-      after_save and after_commit now runs even if no changes were made
-      to an existing record.
+      The after_save and after_commit callbacks are called even if no changes were made
 
       Try this...
 
@@ -552,8 +551,7 @@ module Avram::Callbacks
     {% raise <<-ERROR
       after_completed has been removed
 
-      after_save and after_commit now runs even if no changes were made
-      to an existing record.
+      The after_save and after_commit callbacks are called even if no changes were made
 
       Try this...
 
--- a/src/avram/callbacks.cr
+++ b/src/avram/callbacks.cr
@@ -522,9 +522,17 @@ module Avram::Callbacks
   # ```
   #
   macro after_completed(method_name)
-    after_completed do |object|
-      {{ method_name.id }}(object)
-    end
+    {% raise <<-ERROR
+      after_completed has been removed
+
+      after_save and after_commit now runs even if no changes were made
+      to an existing record.
+
+      Try this...
+
+        ▸ after_commit #{method_name.id}
+      ERROR
+    %}
   end
 
   # Run the given block after save and after successful transaction commit
@@ -541,16 +549,17 @@ module Avram::Callbacks
   # end
   # ```
   macro after_completed(&block)
-    def after_completed(%object : T)
-      {% if @type.methods.map(&.name).includes?(:after_completed.id) %}
-        previous_def
-      {% else %}
-        super
-      {% end %}
+    {% raise <<-ERROR
+      after_completed has been removed
 
-      {{ block.args.first }} = %object
-      {{ block.body }}
-    end
+      after_save and after_commit now runs even if no changes were made
+      to an existing record.
+
+      Try this...
+
+        ▸ after_commit #{block.id}
+      ERROR
+    %}
   end
 
   # :nodoc:
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -291,19 +291,17 @@ abstract class Avram::SaveOperation(T)
 
   def save : Bool
     before_save
-    if valid? && (!persisted? || changes.any?)
+
+    if valid?
       transaction_committed = database.transaction do
-        insert_or_update
-        saved_record = record.not_nil!
-        after_save(saved_record)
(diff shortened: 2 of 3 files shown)


matthewmcgarvey (approved):
I think this is a really good change 👍

src/avram/callbacks.cr
```diff
@@ -522,9 +522,17 @@ module Avram::Callbacks
...
   #
   macro after_completed(method_name)
-    after_completed do |object|
-      {{ method_name.id }}(object)
-    end
+    {% raise <<-ERROR
+      after_completed has been removed
+
+      after_save and after_commit now runs even if no changes were made
+      to an existing record.
```
jwoertink, line 529:
```suggestion
      The after_save and after_commit callbacks are called even if no changes were made
```

src/avram/callbacks.cr
```diff
@@ -541,16 +549,17 @@ module Avram::Callbacks
...
-        super
-      {% end %}
+    {% raise <<-ERROR
+      after_completed has been removed
 
-      {{ block.args.first }} = %object
-      {{ block.body }}
-    end
+      after_save and after_commit now runs even if no changes were made
+      to an existing record.
```
jwoertink, line 556:
```suggestion
      The after_save and after_commit callbacks are called even if no changes were made
```

src/avram/save_operation.cr
```diff
@@ -291,19 +291,17 @@ abstract class Avram::SaveOperation(T)
...
   def save : Bool
     before_save
-    if valid? && (!persisted? || changes.any?)
+
+    if valid?
       transaction_committed = database.transaction do
-        insert_or_update
-        saved_record = record.not_nil!
-        after_save(saved_record)
+        insert_or_update if changes.any? || !persisted?
```
jwoertink, line 297: 👍

src/avram/save_operation.cr
```diff
@@ -357,7 +351,19 @@ abstract class Avram::SaveOperation(T)
...
+
+      after_save and after_commit now runs even if no changes were made
+      to an existing record.
+
+      Try this...
+
+        ▸ use after_commit instead
+      ERROR
+    %}
+  end
```
jwoertink, line 366: This is actually no longer needed. We can just delete it.

jwoertink (changes requested):
I think this is the best option. Great work on this! Just a few small changes, and we can get this merged in!

jwoertink (approved):
Awesome! Thank you 🎉

PR #614, opened 2021-01-17
Renaming Box to Factory

### LINKED ISSUE #498: Change Box -> Factory
See https://github.com/luckyframework/lucky/discussions/1282

We will change `Box` to `Factory`

jwoertink:
Fixes #498

This change renames `Avram::Box` to `Avram::Factory`. The main reasons are that the term "Factory" seems to be more common use for this type of object, as well as Crystal already has a built-in [Box](https://crystal-lang.org/api/0.35.1/Box.html) class.


src/avram/factory.cr
```diff
@@ -1,11 +1,17 @@
...
+
+abstract class Avram::Factory
   getter operation
 
   SEQUENCES = {} of String => Int32
 
   macro inherited
     {% unless @type.abstract? %}
-      {% operation = @type.name.gsub(/Box/, "::SaveOperation").id %}
+      {% operation = @type.name.gsub(/Factory$/, "::SaveOperation").id %}
```
jwoertink, line 14: Added the `$` here to ensure it's replacing the ending `Factory` word in case someone has a model called something like `FactorySetting`

src/avram/factory.cr
```diff
@@ -31,10 +37,6 @@ abstract class Avram::Box
     end
   end
 
-  def self.save
-    {% raise "'Box.save' has been renamed to 'Box.create' to match 'SaveOperation.create'" %}
-  end
```
jwoertink, line 36: This has been removed for quite a few versions now, so just getting rid of it instead of fixing the error message.

matthewmcgarvey (approved):

Issue #129, opened 2019-07-02
Add support for composite primary keys


paulcsmith:
Composite (PRIMARY KEY (id_1, id_2))

Extracted from #96

Whaxion:
Any news on this ?

jwoertink:
Hey @Whaxion! There hasn't been any movement on this yet. We know it's something we'd eventually like to add, but sort of a low priority for the core team at the moment while we try to clean up some other issues. We'd welcome any PR though if you're willing to give this a shot. @matthewmcgarvey recently did some work on primary key stuff, and may be able to assist or provide guidance.

Whaxion:
Hello, I've started working on an implementation of composite primary keys, if you want to see my progression, you can check here https://github.com/Whaxion/avram/tree/composite-primary-keys :smile:

matthewmcgarvey:
@Whaxion before you get too far on that, I'd really like for you to discuss your plans in here first. Looking at your branch, I see some things that we might not want to do in Avram.

The things I'd really like to know before diving into this

- how do other ORMs handle this? (Rails, Elixir/Ecto, statically typed orms like Diesel)
- What kind of API changes will this entail? (I assume `UserQuery.find` will take multiple args but shouldn't they be named to match the oclumns?)
- How can we break this down into small pieces?

jwoertink:
> Really excited to have you helping out with this! 

I'd like to second that @Whaxion! Super stoked to see someone taking this on. We know it's a pretty huge task.

Whaxion:
@matthewmcgarvey To be honest, I haven't thought of looking at how other ORMs handle this, my bad. I'll do it just after posting this really long message.
About the API changes, I'm trying to avoid breaking changes and currently, I haven't made any breaking changes.

So, now I'll explain what I've already made.
Currently, the migrator primary_key in create_table support composite primary keys.
It works like this
```crystal
primary_key id1 : Int64, id2 : UUID
```
Because of the way I've done it, a non composite primary_key still works exactly like before (I almost haven't touched the old code)
So `primary_key id : Int64` still works exactly like before (the spec is full green)
For the `Model#primary_key` it works the same so `primary_key id1 : Int64, id2 : UUID` and old code still works exactly like before.
[...]

PR #616, opened 2021-01-21
Composite primary key migration

Whaxion:
This PR is about the migration part of the composite primary key implementation.
It adds `composite_primary_key` to the create table migrator.
It works like this
```crystal
def migrate
  create table_for(UserPost) do
    add_belongs_to user : User, on_delete: :cascade
    add_belongs_to post : Post, on_delete: :cascade
    primary_key :user_id, :post_id
  end
end
```

There is still some work to do with add_belongs_to if it's a composite PK but for that it needs some work on the model part.
Issue #129

### DIFF
--- a/src/avram/migrator/create_table_statement.cr
+++ b/src/avram/migrator/create_table_statement.cr
@@ -75,6 +75,9 @@ class Avram::Migrator::CreateTableStatement
   end
 
   macro composite_primary_key(*columns)
+    {% if columns.size < 2 %}
+    {% raise "composite_primary_key expected at least two primary keys, instead got #{columns.size}" %}
+    {% end %}
     constraints << "  PRIMARY KEY ({{columns.join(", ").id}})"
   end
 
--- a/src/avram/migrator/create_table_statement.cr
+++ b/src/avram/migrator/create_table_statement.cr
@@ -6,6 +6,7 @@ class Avram::Migrator::CreateTableStatement
   include Avram::Migrator::MissingOnDeleteWithBelongsToError
 
   private getter rows = [] of String
+  private getter constraints = [] of String
 
   def initialize(@table_name : Symbol)
   end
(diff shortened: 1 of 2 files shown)


src/avram/migrator/create_table_statement.cr
```diff
@@ -71,6 +74,13 @@ class Avram::Migrator::CreateTableStatement
       .build
   end
 
+  macro composite_primary_key(*columns)
+    {% if columns.size < 2 %}
+    {% raise "composite_primary_key expected at least two primary keys, instead got #{columns.size}" %}
+    {% end %}
+    constraints << "  PRIMARY KEY ({{columns.join(", ").id}})"
```
matthewmcgarvey, line 81: Can you explain why you added this `constraints` array instead of adding it to the rows array?

Whaxion: It's because constraints must be at the end of create table declaration, so if it was in row, the user should call `composite_primary_key` at the end of the migration.
So user couldn't do a migration like this:
```crystal
def migrate
  create table_for(UserPost) do
    add_belongs_to user : User, on_delete: :cascade
    add_belongs_to post : Post, on_delete: :cascade
    composite_primary_key :user_id, :post_id

    add example : String
  end
end
```
But should do this:
```crystal
def migrate
  create table_for(UserPost) do
    add_belongs_to user : User, on_delete: :cascade
    add_belongs_to post : Post, on_delete: :cascade

    add example : String

    composite_primary_key :user_id, :post_id
  end
end
```
And I find that's sad that we can't put the `composite_primary_key` where we want so I added a constraints array.
But if that's not okay I can change that.

matthewmcgarvey: @Whaxion I didn't know that it had to be at the end. Good thinking. Could you add a spec showing that a composite key is added at the end of the sql statement even if it is declared first? I'd like to make sure we remember to check for that behavior even if we change things in the future.

matthewmcgarvey (approved):
Nice job @Whaxion 🎉

PR #621, opened 2021-01-25
Do not run operation if invalid

### LINKED ISSUE #619: Operation should not `#run` if not valid
We are calling `#run` before checking if operation is valid:

https://github.com/luckyframework/avram/blob/e6623e6e6351560389d4cce1017b30816d1e12b5/src/avram/operation.cr#L53-L63

I think what we want is something like this:

```crystal
    def self.run(params : Avram::Paramable, *args, **named_args)
      operation = self.new(params, *args, **named_args)
      value = nil

      operation.before_run

      if operation.valid?
        value = operation.run # <<== Call `#run` here instead
        operation.after_run(value)
      end

      yield operation, value
    end
```

This way `#run` is never called for an invalid operation. I'm having a record deleted in an operation that is invalid (using *basic* operations, **not** the new delete operations), which is weird.

akadusei:
Resolves #619

### DIFF
--- a/spec/operations/operation_spec.cr
+++ b/spec/operations/operation_spec.cr
@@ -25,6 +25,18 @@ private class FailingTestOperation < Avram::Operation
   end
 end
 
+private class PassingTestOperation < Avram::Operation
+  attribute name : String
+
+  before_run do
+    validate_required name
+  end
+
+  def run
+    "run_method_called"
+  end
+end
+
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Originally I had a reason why I did it this way, but now I don't remember why 🤔 . I think this change makes sense 👍 Thanks!

PR #625, opened 2021-01-30
Remove special handling of Float64 columns

matthewmcgarvey:
I didn't really dig into the history of why it was written the way it was, but the `ColumnType` alias for `Float64` was specifically made for that and was unused.


jwoertink (approved):
I think we had the mapping setup before the ColumnType and just never went back to update. This is a nice cleanup though!

PR #626, opened 2021-01-31
Remove bool return type restriction on Avram::Database#transaction

### LINKED ISSUE #416: Avram::Database.transaction block return behavior undocumented and may be incorrect
`Avram::Database.transaction` wants the transaction block to always return `true`. This is not documented, and is probably a left-over from when someone was designing how the user would roll back the transaction, perhaps returning `false` was considered for initiating a rollback. Obviously we have Avram::Database.rollback to do that today, so the transaction block should have its return value ignored.

matthewmcgarvey:
Fixes #416

This allows for code to use transactions like:

```crystal
AppDatabase.transaction do
  StepOneOperation.run!
  StepTwoOperation.run!
end
```

This would only have worked previously if `StepTwoOperation.run!` returned a boolean. Otherwise `true` would have to be added as the last line because we required the block to return a Bool.

I believe the original type restriction was put in place to match crystal-db's but it was not intention https://github.com/crystal-lang/crystal-db/issues/145

### DIFF
--- a/src/avram/database.cr
+++ b/src/avram/database.cr
@@ -156,7 +156,7 @@ abstract class Avram::Database
   end
 
   # :nodoc:
-  def transaction : Bool
+  def transaction
     if current_transaction
       yield
     else
--- a/src/avram/database.cr
+++ b/src/avram/database.cr
@@ -156,9 +156,10 @@ abstract class Avram::Database
   end
 
   # :nodoc:
-  def transaction
+  def transaction : Bool
     if current_transaction
       yield
+      true
     else
       wrap_in_transaction do
         yield


src/avram/database.cr
```diff
@@ -156,7 +156,7 @@ abstract class Avram::Database
   end
 
   # :nodoc:
-  def transaction : Bool
+  def transaction
```
jwoertink, line 159: Not too sure about this change. In theory, if it's ok, then we should be able to remove [this line](https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L299) and still have things work. I pulled down this PR, and got 3 failures on SaveOperations.

src/avram/database.cr
```diff
@@ -156,7 +156,7 @@ abstract class Avram::Database
...
     if current_transaction
       yield
```
jwoertink, line 161: Now, as a sort of "hack", we can leave the return type as `Bool`, but then add a `true` right after this line. The `wrap_in_transaction` method will already return the bool for that branch. This lets people put whatever in the block, and we still have that return type. The reason I say "hack" though is, i'm not a huge fan of just have a random static value just chillin there, but It would be a step in the right direction, and no different than what we're doing now 🤔

matthewmcgarvey:
Forgot about this PR but I have updated it to ignore the return value of the yield so that the block isn't forced to return a boolean.

jwoertink (approved):

src/avram/database.cr
BrucePerens: Wouldn't it be better if this method was declared `def transaction : Nil` ? Anything that attempted to use the value would get a complaint.

Issue #636, opened 2021-02-12
Double alias no longer works


matthewmcgarvey:
This seems to not working anymore with lucky 0.26

```
alias Double = Float64
class Agency < BaseModel
    column latitude : Double?
    column longitude : Double?
end
```

The error:
```
PG::ResultSet#read returned a Float64. A (PG::Numeric | Nil) was expected.
```

I see here https://github.com/luckyframework/avram/commit/8bf9a1fa5f30058f6676c25ad43c5d69129c12bc#diff-7388ecacebfb91d79a5f8b87e42556bdbc805add6ebe0c5005ad7b180ba8efddR7 an update of the ColumnType

_Originally posted by @brunto in https://github.com/luckyframework/avram/issues/200#issuecomment-778076297_

brunto:
I think simply update the `ColumnType` should work: `alias ColumnType = Float64 | ::PG::Numeric`

matthewmcgarvey:
Completed by https://github.com/luckyframework/avram/pull/637

PR #637, opened 2021-02-17
Add Float64 to ColumnType

brunto:
This is a solution for the issue: https://github.com/luckyframework/avram/issues/636

### DIFF
--- a/db/migrations/20210217224833_add_latitude_and_longitude_to_businesses.cr
+++ b/db/migrations/20210217224833_add_latitude_and_longitude_to_businesses.cr
@@ -1,0 +1,16 @@
+class AddLatitudeAndLongitudeToBusinesses::V20210217224833 < Avram::Migrator::Migration::V1
+  def migrate
+    execute <<-SQL
+      ALTER TABLE businesses
+      ADD COLUMN latitude DOUBLE PRECISION,
+      ADD COLUMN longitude DOUBLE PRECISION;
+    SQL
+  end
+
+  def rollback
+    alter :businesses do
+      remove :latitude
+      remove :longitude
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


matthewmcgarvey:
@brunto we're not going to remember why this type union was added. Could you add a spec that would fail before the change and this fixes?

jwoertink (approved):
Nice! Thanks for adding these specs. It's good to have them in.

brunto:
It's not finished I have one more test to add.￼

brunto:
Ok now I'm done.

jwoertink:
@brunto now that you added that DOUBLE, I'm not sure if that type actually gets checked. I think it's just at runtime when you query on it, so you'd have to now add a spec that checks you can query on a DOUBLE column.

jwoertink (changes requested):
I think the rest of it is fine. I'm not sure the best place to put the spec, but you can probably just add it [here](https://github.com/luckyframework/avram/blob/master/spec/queryable_spec.cr) and say that we can query float columns defined as `DOUBLE`.

brunto:
@jwoertink I just saw the "bucket" table. Maybe it's better to put the "Double" columns inside Bucket instead of Business?

jwoertink:
@brunto the buckets was used for Array stuff. I guess if you wanted to test a postgres array of double, then you can add an additional one there too, but otherwise what you have in business is fine.

jwoertink:
Ok, I pulled this in to one of my projects, and it seems to work. @brunto if you get a sec, can you rebase to make sure everything is good?

brunto:
I rebased the code but I have one more spec to push.

brunto:
@jwoertink Ok all is good now.

jwoertink (approved):
woo hoo! Thanks for tackling this. 🎉

PR #652, opened 2021-03-18
New method to add parenthesis to queries

### LINKED ISSUE #488: New wrap_query method
With the new https://github.com/luckyframework/avram/pull/442 OR queries added in, it's possible you could create some complex queries that will return the wrong data. In some cases you may need to group your conditions by wrapping them in parenthesis. We need to add a new method that will allow us to group these conditions and wrap them in parenthesis.

```crystal
# WHERE (name = 'Bill' OR name = 'Bob')
UserQuery.new.where(&.name("Bill").or(&.name("Bob")))
```

jwoertink:
Fixes #488

When it comes to more complex queries, you may need to specify precedence for your conditions. This is done by wrapping your condition in parenthesis.

This PR adds a new `where` overload for your query objects that takes a block, and will wrap whatever query is defined inside the block.

```crystal
# This Query
UserQuery.new
  .name("Bill")
  .or { |q|
    q.where(&.age(19).nickname("Duke"))
  }
```

```sql
-- Gives us this SQL
SELECT *
FROM users
WHERE name = 'Bill' OR ( age = 19 AND nickname = 'Duke' )
```

This opens Avram up to a lot more complex (but type-safe) queries.

### DIFF
--- a/spec/queryable_spec.cr
+++ b/spec/queryable_spec.cr
@@ -432,6 +432,20 @@ describe Avram::Queryable do
       orig_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users"
       new_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE ( users.nickname = $1 ) LIMIT 3"
     end
+
+    it "doesn't add parenthesis when query to wrap is provided" do
+      query = UserQuery.new.name("Susan").where do |q|
+        some_condition = false
+        if some_condition
+          q.name("john")
+        else
+          q
+        end
+      end.age(25).query
+
+      query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE users.name = $1 AND users.age = $2"
+      query.args.should eq ["Susan", "25"]
+    end
   end
 
   describe "#limit" do
--- a/spec/queryable_spec.cr
+++ b/spec/queryable_spec.cr
@@ -401,6 +401,30 @@ describe Avram::Queryable do
     end
   end
 
+  describe "#where with block" do
+    it "wraps a simple where clause with parenthesis" do
+      query = UserQuery.new.where(&.age(30)).query
+
+      query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE ( users.age = $1 )"
+      query.args.should eq ["30"]
+    end
+
+    it "wraps complex queries" do
+      query = UserQuery.new.where { |user_q|
+        user_q.where { |q|
+          q.age(25).or(&.age(26))
+        }.where { |q|
+          q.name("Billy").or(&.name("Tommy"))
+        }
+      }.or { |q|
+        q.nickname("Strange")
+      }.query
+
+      query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE ( ( users.age = $1 OR users.age = $2 ) AND ( users.name = $3 OR users.name = $4 ) ) OR users.nickname = $5"
+      query.args.should eq ["25", "26", "Billy", "Tommy", "Strange"]
+    end
+  end
+
   describe "#limit" do
     it "adds a limit clause" do
       queryable = UserQuery.new.limit(2)
--- a/spec/queryable_spec.cr
+++ b/spec/queryable_spec.cr
@@ -423,6 +423,15 @@ describe Avram::Queryable do
       query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE ( ( users.age = $1 OR users.age = $2 ) AND ( users.name = $3 OR users.name = $4 ) ) OR users.nickname = $5"
       query.args.should eq ["25", "26", "Billy", "Tommy", "Strange"]
     end
+
+    it "clones properly when assigned to a variable and updated later" do
+      orig_query = UserQuery.new
+
+      new_query = orig_query.where(&.nickname("BusyCat")).limit(3)
+
+      orig_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users"
+      new_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE ( users.nickname = $1 ) LIMIT 3"
+    end
   end
 
   describe "#limit" do
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -309,6 +309,11 @@ class Avram::QueryBuilder
     @wheres.last.conjunction = Avram::Where::Conjunction::None unless @wheres.empty?
   end
 
+  # Removes the last `Avram::Where` to be added
+  def remove_last_where
+    @wheres.pop
+  end
+
   @_wheres_sql : String?
 
   private def wheres_sql
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -308,6 +308,12 @@ class Avram::QueryBuilder
     block.call(self)
   end
 
+  # Clears the last conjunction
+  # e.g. users.age = $1 AND -> users.age = $1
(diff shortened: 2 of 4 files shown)


src/avram/query_builder.cr
```diff
@@ -308,6 +308,12 @@ class Avram::QueryBuilder
     block.call(self)
   end
 
+  # Clears the last conjunction
+  # e.g. users.age = $1 AND -> users.age = $1
+  def clear_conjunction
+    @wheres.last.conjunction = Avram::Where::Conjunction::None unless @wheres.empty?
```
jwoertink, line 314: This was needed because the last where might `"("` and it was generating `WHERE ( AND users.age...`

src/avram/query_builder.cr
```diff
@@ -322,6 +328,9 @@ class Avram::QueryBuilder
         [clause, sql_clause.conjunction.to_s]
       end
 
+      # Remove blank conjunctions
+      statements.reject!(&.blank?)
```
jwoertink, line 332: By adding in the `Avram::Where::Conjunction::None`, the `statements` might look like `["(", "", "users.age = $1", "AND", ")", "AND"]`

jwoertink:
@stephendolan [you mentioned](https://github.com/luckyframework/avram/issues/488#issuecomment-757543951) needing to use this, can you see if this would work for your usecase?

stephendolan:
Heck yeah! I'll pop in and try it this weekend, @jwoertink ! Thanks for putting this together.

src/avram/queryable.cr
```diff
@@ -122,6 +122,14 @@ module Avram::Queryable(T)
     clone.tap &.query.where(sql_clause)
   end
 
+  def where : self
+    query.where(Avram::Where::PrecedenceStart.new)
```
matthewmcgarvey, line 126: I _think_ you're going to need a clone call before this. If you assign the query to the variable before calling this where without the clone it will have the opening parenthesis but nothing else on it after.

matthewmcgarvey (changes requested):
I think I'm happy with this 👍 Just the one issue

src/avram/queryable.cr
jwoertink: oh, interesting. Ok, I'll add a spec for that. Thanks

jwoertink: Good catch. What it does is adds a blank `WHERE`. `SELECT * FROM users WHERE ;`... But after adding the clone, now I have a new issue where it won't add the `(` to the query. `SELECT * FROM users WHERE age = $1 );`

```crystal
clone.tap &.query.where(Avram::Where::PrecedenceStart.new)
result = yield self
result.query.clear_conjunction
result.query.where(Avram::Where::PrecedenceEnd.new)
result
```

If I throw the whole thing inside of the `clone.tap`, then it adds the parens, but omits the actual where clauses.

jwoertink: Ahh... Ok, that works. But wow.... All this cloning has to be expensive, right? 🤔

matthewmcgarvey:
Before I approve, @jwoertink what happens in this scenario:

```crystal
UserQuery.new.name("Susan").where do |q|
  if false
    q.name("john")
   else
     q
  end
end.age(25).to_sql
```

I'm wondering how the clear last conjunction works if the where doesn't add any conditions.

jwoertink:
> Before I approve, @jwoertink what happens in this scenario:
> 
> ```crystal
> UserQuery.new.name("Susan").where do |q|
>   if false
>     q.name("john")
>    else
>      q
>   end
> end.age(25).to_sql
> ```
> 
> I'm wondering how the clear last conjunction works if the where doesn't add any conditions.

lol 😂  another good catch... Maybe I need to rethink how this is done...

You'll end up with this:

```
"SELECT * FROM users WHERE users.name = $1 AND ( ) AND users.age = $2
```

jwoertink:
Should it raise an exception if you use this `where` method, and there's no query being altered? Like, doing `.where(&.itself)` should throw an exception? Or do I account for that, and just remove the parens from the stack completely?

matthewmcgarvey:
If that is valid SQL, I'm ok with it for now. I have a rough idea of how we might refactor the where storage in the future.

jwoertink:
It is not valid SQL...

```
avram_dev=# SELECT * FROM users WHERE users.name = 'Susan' AND ( ) AND users.age = 25;
ERROR:  syntax error at or near ")"
LINE 1: ...ECT * FROM users WHERE users.name = 'Susan' AND ( ) AND user...
                                                             ^
```

jwoertink:
Though, it probably is an edge case... I guess we could open up a separate issue and come back to refactor later?

matthewmcgarvey:
I'd prefer not to push a feature with a know bug. Could you check if the last item in the array is an open parens and remove it if so?

spec/queryable_spec.cr
```diff
@@ -401,6 +401,53 @@ describe Avram::Queryable do
...
+      new_query = orig_query.where(&.nickname("BusyCat")).limit(3)
+
+      orig_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users"
+      new_query.query.statement.should eq "SELECT #{User::COLUMN_SQL} FROM users WHERE ( users.nickname = $1 ) LIMIT 3"
+    end
+
+    it "doesn't add parenthesis when query to wrap is provided" do
+      query = UserQuery.new.name("Susan").where do |q|
+        some_condition = false
+        if some_condition
```
jwoertink, line 439: Because ameba throws an error if you use a bool literal as a condition

jwoertink:
Yeah, I was able to just check the last added Where, and see it was the PrecedenceStart. Then I can just remove it, and skip the rest.

src/avram/query_builder.cr
```diff
@@ -303,6 +303,17 @@ class Avram::QueryBuilder
...
 
+  # Clears the last conjunction
+  # e.g. users.age = $1 AND -> users.age = $1
+  def clear_conjunction
+    @wheres.last.conjunction = Avram::Where::Conjunction::None unless @wheres.empty?
+  end
+
+  # Removes the last `Avram::Where` to be added
+  def remove_last_where
+    @wheres.pop
```
jwoertink, line 314: Should this return `self`? 🤔

matthewmcgarvey (approved):
Lots of tiny issues, but you got it there, nice job! 🎉

src/avram/query_builder.cr
jwoertink: @matthewmcgarvey what are your thoughts on this? I assume no one will call this but us... but should this return `self`? or do you think it even matters?

matthewmcgarvey: I was looking at query builder yesterday and couldn't see any pattern to when we clone queryable vs query builder. I wouldn't worry about it

jwoertink: Cool. Thanks for the reviews!

stephendolan:
Finally got around to playing with this in my production app, and it worked beautifully.

This is really handy to throw into a scope so that no matter where you use it, you can guarantee that your results are grouped. Here's an example where I want to search videos by a search term across lots of attributes:

```crystal
class VideoQuery < Video::BaseQuery
  def matching_search_term(search_term : String)
    where do |query|
      query.title.ilike(search_term)
        .or(&.description.ilike(search_term))
        .or(&.left_join_topics.where_topics(TopicQuery.new.label.ilike(search_term)))
    end
  end
end
```

paulcsmith:
This is pretty dope!! Can you also pass a query object to where, or does it need to be a block? e.g. can you do `.where(PostQuery.published)` instead of `where(&.published(true))`?

jwoertink:
You can do `where_posts(PostQuery.published)`, but not for the parenthesis wrapping part. That still needs to be done with a block.

Issue #622, opened 2021-01-28
Add a change_belongs_to Migrator method


jwoertink:
We currently have `rename_belongs_to` and `remove_belongs_to`, but say you originally set your `on_delete: :do_nothing`, but wanted to update that to `on_delete: :cascade`. It would be great to have a `change_belongs_to` macro to handle doing that.

PR #654, opened 2021-03-19
Add new method for droping a foreign key constraint in migrations.

jwoertink:
In order to complete #622, we have to drop the constraint first, then recreate it with the new `on_delete` statement.

This PR adds in the ability to drop a foreign key constraint which would allow you to change your `belongs_to` if you needed.

```crystal
# In a previous migration, you had
def migrate
  create table_for(Post) do
    add_belongs_to user : User, on_delete: :nullify
  end
end

# Now you want to change it to cascade
def migrate
  drop_foreign_key :posts, :users
  create_foreign_key :posts, :users, on_delete: :cascade
end
```

This is more of an escape hatch without you having to write all of the SQL by hand. The next step would be a separate PR that implements a `change_belongs_to` type macro.

### DIFF
--- a/src/avram/migrator/drop_foreign_key_statement.cr
+++ b/src/avram/migrator/drop_foreign_key_statement.cr
@@ -1,0 +1,22 @@
+# Generates an ALTER TABLE statement for dropping a foreign key constraint on a table.
+#
+# ### Usage
+#
+# ```
+# DropForeignKeyStatement.new(from: :comments, references: :users, column: :author_id).build
+# # => "ALTER TABLE comments DROP CONSTRAINT comments_author_id_fk;"
+# ```
+class Avram::Migrator::DropForeignKeyStatement
+  def initialize(@from : Symbol, @references : Symbol, @column : Symbol? = nil)
+  end
+
+  def build
+    foreign_key = @column || Wordsmith::Inflector.singularize(@references.to_s) + "_id"
+    String.build do |index|
+      index << "ALTER TABLE"
+      index << " #{@from}"
+      index << " DROP CONSTRAINT #{@from}_#{foreign_key}_fk"
+      index << ";"
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


src/avram/migrator/drop_foreign_key_statement.cr
```diff
@@ -0,0 +1,22 @@
...
+class Avram::Migrator::DropForeignKeyStatement
+  def initialize(@from : Symbol, @references : Symbol, @column : Symbol? = nil)
+  end
+
+  def build
+    foreign_key = @column || Wordsmith::Inflector.singularize(@references.to_s) + "_id"
+    String.build do |index|
+      index << "ALTER TABLE"
+      index << " #{@from}"
+      index << " DROP CONSTRAINT #{@from}_#{foreign_key}_fk"
```
matthewmcgarvey, line 18: hmm, would it be better to have the optional value be the foreign key name? So, the foreign key would either be created from the reference or passed in explicitly? 🤔

jwoertink: Is this what you mean? https://github.com/luckyframework/avram/pull/654/files#diff-ba436c31cb078c2767a29a28ec723092b8441934ea17e02cb97831099417ed60R10 There's that third arg that lets you specify the column it's on. Or do you mean just pass in the raw fk name itself? `name: "what_ever_i_want_fk"`

matthewmcgarvey: Yea, I'm thinking that we provide an extra config option for this method but it doesn't actually support all the reasons for not being able to use the default flow

jwoertink: I was thinking about this over the weekend. The CreateForeignKey class doesn't give you the option to do a raw name https://github.com/luckyframework/avram/blob/master/src/avram/migrator/create_foreign_key_statement.cr#L10. I wonder if we should say at that level, you just need to run the `execute` yourself. These classes are more like the midway between the "lucky way" (using helper methods), and the escape hatch (writing yourself). 

If we think these should open up the options all the way, then we will want to make sure they all follow those same rules. What are your thoughts?

matthewmcgarvey: I'm fine with this thinking 👍

matthewmcgarvey (approved):

PR #655, opened 2021-03-23
Cleaning up for Crystal 1.0 support

jwoertink:
Updating the crystal version constraint to allow for through 1.0. Also updating the CI to match the rest with the matrix setup.


src/avram/object_extensions.cr
```diff
@@ -1,4 +1,16 @@
 class Object
+  def blank?
```
wontruefree, line 2: was this removed from the standard lib?

jwoertink: It never existed in standard lib on Object from what I can tell; only on String. We had to use a shard to get it on all objects. Since the implementation was small, I figured it was easier to just bring it over.

PR #657, opened 2021-03-31
Fix table inferrencing for constant-case model names

### LINKED ISSUE #656: Fix inferring table names for constant case models
A model called `HTML` should have the table name inferred to be `htmls` but it is inferred to be `html_ls` because we pluralize before we underscore.

Line: https://github.com/luckyframework/avram/blob/1dc80d424992a412864d10c63d959e03539ad129/src/run_macros/infer_table_name.cr#L3

Reference: https://github.com/luckyframework/wordsmith/issues/5#issuecomment-810387228

Rails calls underscore before pluralizing:
https://github.com/rails/rails/blob/6675f6b785d9ea7f6650344fb98555fc5acb0a3d/activesupport/lib/active_support/inflector/methods.rb#L194-L196

matthewmcgarvey:
Fixes #656

### DIFF
--- a/spec/migrator/table_for_spec.cr
+++ b/spec/migrator/table_for_spec.cr
@@ -1,0 +1,17 @@
+require "../spec_helper"
+
+describe Avram::TableFor do
+  describe ".table_for" do
+    it "turns model name into pluralized symbol" do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
nice 👍

PR #660, opened 2021-04-03
Allow table name to be Symbol or String, remove run from macros

matthewmcgarvey:
I had heard that using `run` in macros slowed down compilation times. Turns out, it's true!

I ran benchmarks on master and this branch (`hyperfine --warmup 3 'crystal build --release spec/model_spec.cr'`)
master: 5.016 seconds (average)
with these changes: 4.156 seconds (average)

**That's a 17% improvement in release compile time!**

## Changes

Our only use for `run` was to infer the table name in migrations and in the models. We turned the model name into a symbol that was the table name. We only ever needed this info at runtime so it made sense to bring the code that was in the file being `run`. The only problem is that we can't return a symbol anymore because the table name isn't determined until runtime and it's not possible to construct symbols at runtime. I added a `TableName` alias that is just `String | Symbol` and updated _everywhere_ that takes in a table name to use it. I also found that all the places we were using `Model::TABLE_NAME` could be replaced by the class method `Model.table_name` so I removed the constant as well.

### DIFF
--- a/spec/migrator/table_for_spec.cr
+++ b/spec/migrator/table_for_spec.cr
@@ -2,16 +2,16 @@ require "../spec_helper"
 
 describe Avram::TableFor do
   describe ".table_for" do
-    it "turns model name into pluralized symbol" do
-      Avram::TableFor.table_for(User).should eq :users
+    it "turns model name into pluralized table name" do
+      Avram::TableFor.table_for(User).should eq "users"
     end
 
     it "removes double colons" do
-      Avram::TableFor.table_for(Test::Model).should eq :test_models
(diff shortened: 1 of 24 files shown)


jwoertink (approved):
Yeah, I love this idea especially since speeding up the compile time can also help increase dev turn-around time.

I don't think we really lose any compile-time checks here because the schema enforcer should (in theory) be the one checking if you typed `table "usres"` instead of `table "users"`, right?

matthewmcgarvey:
Yep, no loss of safety, schema enforcer was already converting the symbol to a string https://www.github.com/luckyframework/avram/tree/master/src/avram/schema_enforcer/validation.cr

PR #669, opened 2021-04-23
Add Postgres versions and experimental builds to CI test matrix

stephendolan:
We currently only officially test Avram against Postgres 10, which seems less than ideal since the current version is actually 13.

This tests against 10, 11, 12, and 13.
It also adds an experimental test against the Crystal nightly build.


jwoertink (approved):
This is nice. I'm shocked that it can test all of those versions in 5 minutes.

Issue #666, opened 2021-04-15
Custom error messages for all validations


jwoertink:
Most of the validations allow you to specify your own custom error message

https://github.com/luckyframework/avram/blob/master/src/avram/validations.cr 

except validate_size_of https://github.com/luckyframework/avram/blob/master/src/avram/validations.cr#L130 and validate_numeric https://github.com/luckyframework/avram/blob/master/src/avram/validations.cr#L161

PR #670, opened 2021-04-25
Added custom error message for validate_numeric and validate_size_of …

vishalsodani:
…#666

### DIFF
--- a/src/avram/validations.cr
+++ b/src/avram/validations.cr
@@ -127,10 +127,12 @@ module Avram::Validations
   # validate_size_of feedback, min: 18, max: 100
   # validate_size_of password, min: 12
   # ```
+  # ameba:disable Metrics/CyclomaticComplexity
   def validate_size_of(
     attribute : Avram::Attribute(String),
     min = nil,
     max = nil,
+    message = nil,
     allow_nil : Bool = false
   )
     if !min.nil? && !max.nil? && min > max
@@ -143,11 +145,11 @@ module Avram::Validations
       size = attribute.value.to_s.size
 
       if !min.nil? && size < min
-        attribute.add_error "is too short"
+        attribute.add_error message || "is too short"
       end
 
       if !max.nil? && size > max
-        attribute.add_error "is too long"
+        attribute.add_error message || "is too long"
       end
     end
   end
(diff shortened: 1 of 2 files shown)


src/avram/validations.cr
```diff
@@ -143,11 +144,11 @@ module Avram::Validations
       size = attribute.value.to_s.size
 
       if !min.nil? && size < min
-        attribute.add_error "is too short"
+        attribute.add_error message
```
matthewmcgarvey, line 147: Can we try to keep the previous messages for both of these validations?

```suggestion
        attribute.add_error message || "is too short"
```

The message param would default to nil

matthewmcgarvey (changes requested):

src/avram/validations.cr
matthewmcgarvey, line 147: These still need to have the same changes as you did below, then it will be good 👍

matthewmcgarvey (changes requested):

matthewmcgarvey (approved):
Nice job! 🎉🎉🎉

vishalsodani:
> Nice job! tadatadatada

Thanks!

matthewmcgarvey:
@vishalsodani woops, didn't realize that the CI wasn't running. Could you go ahead and tell ameba to ignore those issues? I might try to figure out how to fix them later but it's overkill for this change

Here's docs on how to do it https://github.com/crystal-ameba/ameba#inline-disabling

vishalsodani:
> @vishalsodani woops, didn't realize that the CI wasn't running. Could you go ahead and tell ameba to ignore those issues? I might try to figure out how to fix them later but it's overkill for this change
> 
> Here's docs on how to do it https://github.com/crystal-ameba/ameba#inline-disabling

I added the ignore comments.

matthewmcgarvey:
Looks like the disabling didn't get put in the right place. You should be able to check it locally with `bin/ameba` (it's generated with `shards install`)

vishalsodani:
> Looks like the disabling didn't get put in the right place. You should be able to check it locally with `bin/ameba` (it's generated with `shards install`)

Thanks for the hint. Fixed.

Issue #620, opened 2021-01-25
Not all DB tasks have a "quiet" option 


jwoertink:
Some of the tasks like [db.create](https://github.com/luckyframework/avram/blob/master/src/avram/tasks/db/create.cr#L5) have an option to [silence them](https://github.com/luckyframework/lucky_cli/blob/master/src/web_app_skeleton/spec/setup/setup_database.cr#L1-L2). This is used in specs so when you run `crystal spec` you don't see the migration running every time. The `db.create` and `db.migrate` tasks have this option, but [db.reset](https://github.com/luckyframework/avram/blob/master/src/avram/tasks/db/reset.cr) and [db.drop](https://github.com/luckyframework/avram/blob/master/src/avram/tasks/db/drop.cr) do not.

nodeMD:
Hello 😃 
I've added a PR to address that issue: https://github.com/luckyframework/avram/pull/675
Please take a look and review when you will have a moment

PR #675, opened 2021-05-13
Added "quiet" option to db.reset and db.drop

nodeMD:
PR for Issue: https://github.com/luckyframework/avram/issues/620

### DIFF
--- a/src/avram/tasks/db/drop.cr
+++ b/src/avram/tasks/db/drop.cr
@@ -7,7 +7,7 @@ class Db::Drop < BaseTask
   end
 
   def run_task
-    Avram::Migrator::Runner.drop_db
+    Avram::Migrator::Runner.drop_db(@quiet)
     puts "Done dropping #{Avram::Migrator::Runner.db_name.colorize(:green)}"
   end
 
--- a/src/avram/tasks/db/drop.cr
+++ b/src/avram/tasks/db/drop.cr
@@ -8,7 +8,6 @@ class Db::Drop < BaseTask
 
   def run_task
     Avram::Migrator::Runner.drop_db(@quiet)
-    puts "Done dropping #{Avram::Migrator::Runner.db_name.colorize(:green)}"
   end
 
   def help_message
--- a/src/avram/tasks/db/drop.cr
+++ b/src/avram/tasks/db/drop.cr
@@ -3,6 +3,9 @@ require "colorize"
 class Db::Drop < BaseTask
   summary "Drop the database"
 
+  def initialize(@quiet : Bool = false)
+  end
+
   def run_task
     Avram::Migrator::Runner.drop_db
     puts "Done dropping #{Avram::Migrator::Runner.db_name.colorize(:green)}"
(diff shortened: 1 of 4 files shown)


src/avram/tasks/db/reset.cr
```diff
@@ -3,6 +3,9 @@ require "colorize"
 class Db::Reset < BaseTask
   summary "Drop, recreate, and run migrations."
 
+  def initialize(@quiet : Bool = false)
```
jwoertink, line 6: This instance variable will need to be passed in to the `new()` methods in the `run_task` method below so those will pick up the option.

src/avram/tasks/db/drop.cr
```diff
@@ -3,6 +3,9 @@ require "colorize"
 class Db::Drop < BaseTask
   summary "Drop the database"
 
+  def initialize(@quiet : Bool = false)
```
jwoertink, line 6: Similar to the [create_db](https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L67) method, this variable will need to be passed in to the `drop_db` method in the `run_task` method below.

jwoertink (changes requested):
Awesome! Thanks so much for this PR. This is a great start, but now those variables will just need to be used. I left a few comments, but let me know if you have questions on those.

nodeMD:
Thank you for the review! 🙂
I have passed the variable to the run_task methods.

src/avram/tasks/db/drop.cr
```diff
@@ -3,8 +3,11 @@ require "colorize"
...
+  end
+
   def run_task
-    Avram::Migrator::Runner.drop_db
+    Avram::Migrator::Runner.drop_db(@quiet)
```
jwoertink, line 10: I don't think this will work just yet. That method currently doesn't take any arguments. https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L57 You'll have to update the method just like the create_db https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L67 and then it has to actually be implemented.

jwoertink:
Thanks for the update @nodeMD. I think it's really close. Just needs 1 more bit and it should be good.

src/avram/tasks/db/drop.cr
nodeMD: aaah yes, that makes a lot of sense 😁 Thank you for the tip!

src/avram/tasks/db/drop.cr
```diff
@@ -3,8 +3,11 @@ require "colorize"
...
     puts "Done dropping #{Avram::Migrator::Runner.db_name.colorize(:green)}"
```
jwoertink, line 11: Sorry, I think this is the last thing 😅  This `puts` line should now go in to the drop_db method https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L57 That way the `quiet?` option can be used to silence it like how the `create_db` does https://github.com/luckyframework/avram/blob/master/src/avram/migrator/runner.cr#L69-L71

nodeMD: ah yes 👍 Looks like I didn't got it 😅 without that it wouldn't really be quiet 😁 Thanks for all the comments 💪 It gives me much better understanding how it is working 😃

jwoertink (approved):
Awesome! Thank you so much for taking this on. It's nice when all of the code is a lot more consistent.

PR #676, opened 2021-05-18
Fix edge cases in migrator when using fill_existing_with with a boolean value

jinn999:
This covers a very edge case when performing migrations that imply adding a not null boolean column filling previous records with the false value.
In practice when performing this kind of modification

```
alter table_for(SomeTable) do
 add a_column : Boolean, fill_existing_with: false
```
the migration fails as the fill_existing_with is not considered.
The patch fixes this particular case.

Note 1:
There is currently a way to bypass this by switching the former code to:
```
alter table_for(SomeTable) do
 add a_column : Boolean, fill_existing_with: "false"
```
(add double quotes)

Note 2:
the resulting statement is now:
```
UPDATE users SET admin = 'false';
```
Is it correct? I didn't touch the logic in the BoolColumn, but I would assume that in theory it should be without single quotes. However, it seems that the same applies to FloatColumn as well...

Note 3:
Why is the default value for fill_existing_with in the add_macro nil and not :nothing?

### DIFF
--- a/spec/migrator/alter_table_statement_spec.cr
+++ b/spec/migrator/alter_table_statement_spec.cr
@@ -104,7 +104,6 @@ describe Avram::Migrator::AlterTableStatement do
       built.statements[1].should eq "UPDATE users SET admin = 'false';"
       built.statements[2].should eq "ALTER TABLE users ALTER COLUMN admin SET NOT NULL;"
     end
-
   end
 
   describe "associations" do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Good catch! Thanks for the fix. This is definitely the right direction over using `"false"`.

As for why it's defaulted to `nil` instead of `:nothing`... I'm not sure 😅  It would seem like we could just set the default to to `:nothing`, but I would want to dig in a bit more to double check that wouldn't mess up something else.

PR #677, opened 2021-05-26
Adding ability to listen for notifications from postgres

jwoertink:
This allows us to listen for `pg_notify()` calls for any number of channels. 

```sql
SELECT pg_notify('workers', json_build_object('id', 1, 'name', 'test')::text)
```

```crystal
AppDatabase.listen("workers") do |notify|
  data = JSON.parse(notify.payload)
  puts data["id"] #=> 1
  puts data["name"] #=> "test"
end
```

### DIFF
--- a/src/avram/database.cr
+++ b/src/avram/database.cr
@@ -39,6 +39,22 @@ abstract class Avram::Database
     new.delete
   end
 
+  # Listens for `pg_notify()` calls on each channel in `channels`
+  # Yields a `PQ::Notification` object with `channel`, `payload`, and `pid`.
+  #
+  # ```
+  # # pg_notify("callback", "123")
+  # AppDatabase.listen("callback", "jobs") do |notification|
+  #   notification.channel # => "callback"
+  #   notification.payload # => "123"
+  # end
+  # ```
+  def self.listen(*channels : String, &block : PQ::Notification ->) : Nil
+    new.listen(*channels) do |notification|
+      block.call(notification)
+    end
+  end
+
   @@database_info : DatabaseInfo?
 
   def self.database_info : DatabaseInfo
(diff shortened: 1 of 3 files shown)


wontruefree:
I have been wanting to use this feature forever but have not found a good feature yet.  I am excited this is being added to Avram/Lucky

jwoertink:
Yeah, I have a few upcoming uses for Lucky, so it's time to add it in!

src/avram/database.cr
```diff
@@ -39,6 +39,22 @@ abstract class Avram::Database
...
+  # # pg_notify("callback", "123")
+  # AppDatabase.listen("callback", "jobs") do |notification|
+  #   notification.channel # => "callback"
+  #   notification.payload # => "123"
+  # end
+  # ```
+  def self.listen(*channels : String, &block : PQ::Notification ->) : Nil
+    new.listen(*channels) do |notification|
+      block.call(notification)
+    end
```
matthewmcgarvey, line 55:
```suggestion
    new.listen(*channels, &block)
```

I think that works?

jwoertink: It does, but I actually did this way intentionally thinking that maybe it works better for documentational purposes when people are looking at the code. I wanted to use `yield`, but Crystal blows up because the block is actually a proc 😒  This was me just trying to be clever 😅  but I'll revert it.

matthewmcgarvey: @jwoertink just because you have `&block` as a parameter you can still call yield.

So it could be:

```crystal
new.listen(*channels) do |notification|
  yield notification
end
```

Same goes for all the other places if that's what you want

jwoertink: I tried that, but I get this error:

```
In src/avram/database.cr:54:7

 54 | yield notification
      ^
Error: can't use `yield` inside a proc literal or captured block

Make sure to read the whole docs section about blocks and procs,
including "Capturing blocks" and "Block forwarding":

https://crystal-lang.org/reference/syntax_and_semantics/blocks_and_procs.html
```

It's all good though.

matthewmcgarvey (approved):

PR #683, opened 2021-06-07
Minor updates to Avram::Queryable

matthewmcgarvey:


src/avram/queryable.cr
```diff
@@ -1,8 +1,11 @@
 module Avram::Queryable(T)
   include Enumerable(T)
 
-  @query : Avram::QueryBuilder?
-  setter query
+  property query : Avram::QueryBuilder do
+    Avram::QueryBuilder
+      .new(table: table_name)
+      .select(schema_class.column_names)
+  end
```
jwoertink, line 8: Does this make it memoized? I've never used this syntax on properties before.

jwoertink (approved):

src/avram/queryable.cr
matthewmcgarvey: Yes, it will set the value if it is called and the value is currently nil. The upside of this is that you don't have to define the instance variable as nilable https://crystal-lang.org/api/1.0.0/Object.html#property(*names,&block)-macro

jwoertink: That's awesome! We should start looking at taking advantage of that.

Issue #682, opened 2021-06-07
Allow creating QueryBuilder with "from" as result of subselect


matthewmcgarvey:
## Overview

Right now the "from" is set by the passed in table name in the initializer https://github.com/luckyframework/avram/blob/72ac1d684fb4624defe65495e4854818f1041306/src/avram/query_builder.cr#L18-L19

That's not the only way to specify a from in sql, though.

I would like to update the QueryBuilder to keep track of a `from` instead of `table` and remove the setting of table in the initialize.

I believe all that would need to change for existing usage to work is to call `.from` instead of passing the table name into the initialize in the Queryable here https://github.com/luckyframework/avram/blob/72ac1d684fb4624defe65495e4854818f1041306/src/avram/queryable.cr#L48

I would like to add an overloaded `from` that takes in a QueryBuilder and an alias which would translate to the SQL `from (query_builder statement) as ALIAS`

## Why?

I would like to see Avram have robust layers. You don't need a full model and queryable for this particular sql query? Fine, just use the QueryBuilder for a bit of a nicer experience. And that would be without sacrificing any usability of the higher level constructs.

Ultimately, I would like to be able to experiment with different higher level patterns and need this lower layer to be more flexible.

matthewmcgarvey:
A blocker for this change is that update and insert don't use "from" so a table_name makes sense. For my goals, I think there would have to be a way to break up the different types of sql statements. I'll keep looking into it

PR #687, opened 2021-06-16
Changing how the select_count method counts returned rows.

### LINKED ISSUE #686: Changing how COUNT() works in queries
Today I learned that you can take a query, and wrap it with an outer select count, and 1. it's about the same speed, 2. it allows you to ignore all those weird group by column aggregate errors you get with postgres.

In my case, I have this query that does a `SELECT DISTINCT ON (col) col` which you can't do a count on (with how things work now). You can do `SELECT COUNT(DISTINCT col)`, but it seems this is actually slower. 

Here's an example:

```sql
SELECT COUNT(*) FROM (SELECT DISTINCT ON (users.username) users.id, users.created_at, users.updated_at, users.username FROM users WHERE users.soft_deleted_at IS NULL) AS count;
```

If we changed how the current `select_count` method works, to do this, then it opens up to being about to count on just about any query you run.

Here's some stats on a table I use in production with 5+ million records:

```
production_db=> select count(*) from users;
  count
---------
 5028502
(1 row)

Time: 403.789 ms
production_db=> select count(*) from (select * from users) as count;
  count
---------
 5028502
(1 row)

Time: 403.690 ms
```

I'm proposing we change https://github.com/luckyframework/avram/blob/6c121361311bd190243d0bd663b0976674a6263d/src/avram/query_builder.cr#L193-L195 to just be something more like

```crystal
def select_count
  "SELECT COUNT(*) FROM (#{to_prepared_sql}) AS count"
end
```

Can anyone things of an edge case where this might return the wrong value, or where this would end up being a ton slower?

jwoertink:
Fixes #686

This new way now allows for counting against more complex queries which include using `distinct_on`. The main reason for this is using pagination when you have these complex queries.

Before this PR, you couldn't do this:

```crystal
get "/messages" do
  pages, messages = paginate(MessageQuery.new.distinct_on(&.status))
  html IndexPage, messages: messages, pages: pages
end
```

This would throw an error that `status` needs to be in a "GROUP BY" or used in an aggregate function. But of course, if you added `status` to a group, it would also want `id`, and you'd still have to change the count from `COUNT(*)` to `COUNT(status)` or something else... 

Also as @wyhaines pointed out in [686](https://github.com/luckyframework/avram/issues/686#issuecomment-862664764), there's several cases where this method is actually faster than the previous. All of the queries I tried seemed to perform about the same speed if not faster.

### DIFF
--- a/spec/queryable_spec.cr
+++ b/spec/queryable_spec.cr
@@ -765,14 +765,13 @@ describe Avram::Queryable do
       query.select_count.should eq 1
     end
 
-    it "raises when used with offset or limit" do
-      expect_raises(Avram::UnsupportedQueryError) do
-        UserQuery.new.limit(1).select_count
-      end
+    it "works with distinct_on" do
+      UserFactory.new.age(30).create
+      UserFactory.new.age(30).create
 
-      expect_raises(Avram::UnsupportedQueryError) do
-        UserQuery.new.offset(1).select_count
-      end
+      query = UserQuery.new.distinct_on(&.age)
+
+      query.select_count.should eq 1
     end
 
     it "returns 0 if postgres returns no results" do
--- a/src/avram/queryable.cr
+++ b/src/avram/queryable.cr
@@ -196,7 +196,10 @@ module Avram::Queryable(T)
   end
 
   def select_count : Int64
-    exec_scalar(&.select_count).as(Int64)
+    table = "(#{query.statement}) AS temp"
+    new_query = Avram::QueryBuilder.new(table).select_count
+    result = database.scalar new_query.statement, args: query.args, queryable: schema_class.name
+    result.as(Int64)
   rescue e : DB::NoResultsError
     0_i64
   end


src/avram/queryable.cr
```diff
@@ -196,7 +196,10 @@ module Avram::Queryable(T)
   end
 
   def select_count : Int64
-    exec_scalar(&.select_count).as(Int64)
+    table = "(#{query.statement}) AS temp"
+    new_query = Avram::QueryBuilder.new(table).select_count
```
jwoertink, line 200: This could be cleared up if we implemented https://github.com/luckyframework/avram/issues/682

spec/queryable_spec.cr
```diff
@@ -765,14 +765,13 @@ describe Avram::Queryable do
       query.select_count.should eq 1
     end
 
-    it "raises when used with offset or limit" do
-      expect_raises(Avram::UnsupportedQueryError) do
-        UserQuery.new.limit(1).select_count
-      end
```
jwoertink, line 771: You can now use a limit and offset in the select_count queries. The reason you couldn't before was because `SELECT COUNT()` would always ever return 1 row. So it didn't make sense to have a limit or offset. Now with this sub-select way, you can use limit to your advantage.

Say you have a table with 7 records in it. If you do a limit 10, and then run a count, you'll get 7. Do the same thing once there's 12 records in the table, and then you'll only get 10. It's sort of like an upper bounds check.

matthewmcgarvey (approved):

robacarp (approved):
I like this. I run into this problem in Postgres a lot.

PR #688, opened 2021-06-18
[WIP] Experimental params

jwoertink:
This is meant to accompany https://github.com/luckyframework/lucky/pull/1527 where the goal is to not change up how params are handled, but instead just extend what can be handled.

In order to do this, I had to change how `Avram::Params` works. Instead of being `Hash(String, String)`, it needs to be `Hash(String, Array(String))`. This class is really just used for internal to Avram, and ensuring that operations have some sort of "paramable" object when they're instantiated with no args... This really shouldn't affect anyone's apps (aside from maybe mine because we got lazy and used Avram::Params in our apps... 😐 ). 

The tricky part here is that the param stuff is so wide spread. It's like using C, but the h file is here, and the c file is in a whole different repo 😂  On top of that, we have `UploadParams`, `FakeParams`, and `NestedFakeParams` all for the purposes of testing.

*if* we decide to go down this route, then we will just need to deal with the fact that this is going to be messy and complicated until we really sit down and work out a whole new interface (probably post 1.0). 

### Goals

* [ ] - Keep from any major breaking changes to the interface
* [ ] - Support arrays in params
* [ ] - Support file arrays in params

### DIFF
--- a/spec/operations/define_attribute_spec.cr
+++ b/spec/operations/define_attribute_spec.cr
@@ -6,6 +6,7 @@ private class OperationWithAttributes < Avram::Operation
   attribute count : Int32
   attribute checked : Bool = false
   attribute thing : String = "taco"
+  attribute tags : Array(String) = [] of String
   file_attribute :thumb
 
   def update_count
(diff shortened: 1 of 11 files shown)


jwoertink:
This is no longer needed.

PR #693, opened 2021-06-30
Fix DateTime Form Input Parsing

stephendolan:
Closes https://github.com/luckyframework/lucky/issues/1530

I originally fixed this here: https://github.com/luckyframework/avram/pull/603

At least, LuckyCasts DateTime inputs were working at some point with this PR. However, since then, it appears as though that's stopped working:
https://github.com/luckyframework/lucky/issues/1530

I attempted to refactor this test file to make it a bit cleaner and easier to extend with other tests in the future. I'm not sure if I should put the struct/enum in another `support` file or something, but it made sense to me to keep it local to the test since this is the only place it was needed.

I've tested this branch with LuckyCasts, and DateTimes are saving correctly with it in place.

### DIFF
--- a/spec/type_extensions/time_spec.cr
+++ b/spec/type_extensions/time_spec.cr
@@ -1,30 +1,58 @@
 require "../spec_helper"
 
+enum TimeComponent
+  Year
+  Month
+  Day
+  Hour
+  Minute
+  Second
+end
+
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Looks good to me!

Issue #197, opened 2019-08-01
jsonb special query operators


jwoertink:
We can now map to jsonb fields using `JSON::Any` in our migrations and columns, but currently querying on those fields is limited to normal WHERE syntax, and some hefty crystal:

```crystal
# WHERE users.preferences = {"theme": "dark_mode"}
UserQuery.new.preferences(JSON::Any.new({"theme" => JSON::Any.new("dark_mode")}))
```

Postgres supports several different operators for doing special queries on JSON https://www.postgresql.org/docs/current/functions-json.html

We need some type safe ways to do things like "Give me all users that have dark_mode as their theme preference" and such.

jwoertink:
To along with this, I just ran in to an issue with rails. If you want to check that a jsonb column has a specific key, postgres uses `column ? 'key'`. Since the `?` is the same as the bind args, you can't pass that key as a separate argument. You'd have to use string interpolation if you needed that to be dynamic, but obviously that opens the door to some issues. 

```ruby
User.where("tags ? 'active'").count #=> 100
User.where("tags ? '?'", "active").count #=> Wrong number of bind args exception
```

For Avram, it would be super helpful to have a method that does this for us, but in a type safe and secure way.

```crystal
# This would be nice. Or something along these lines.
UserQuery.new.tags.has_key("active").select_count
```

wontruefree:
I like the idea of using method missing to create query methods.

```crystal
UserQuery.new.tags.active_key("active").select_count
```

This would be a little bit of magic but I think it would fit into the API well.

kevinsjoberg:
I'd vote against using `method_missing`. While convenient, I really like the simpleness/explicitness of Lucky in general. There's usually not much "magic" going on. Each concept usually just rely on language constructs of Crystal. 🙂

jwoertink:
Agreed. I like the suggestion @wontruefree, but there was some talk floating around of method missing going away at some point. Plus, we want to avoid random bits of "magic". I think we may be able to even get a bit more type-safe with this if https://github.com/luckyframework/avram/pull/695 gets approved. In the case of the example, `active` would be a property on the object which may allow us to pull that in from the macro level (maybe with an annotation?).

Just half-baked, but something like:

```crystal
struct UserTagsJSON
  include JSON::Serializable
  property active : Bool
end

# We know tags would be a serialized object
# If we pass a block, then send the serialized instance down
# and build a query from that...
UserQuery.new.tags(&.active(true)).select_count
```

[...]

PR #695, opened 2021-07-05
Experimental/json serializable columns

### LINKED ISSUE #691: New serializer columns
Working with `JSON::Any` columns is kind of a pain. I have no idea how this would work, or what things we would run in to, but I had this idea that would look like this:

```crystal
class User < BaseModel
  table do
    serialized preferences : PreferenceSerializer
  end
end

class PreferenceSerializer
  include JSON::Serializable

  property theme : String
  property font_size : String
end

user = UserQuery.first
user.preferences.theme #=> "dark"
user.preferences.font_size #=> "XL"
```

My initial thought was under the hood, it would look something like:

```crystal
macro serialized(type_dec)
  column {{ type_dec.var }}_data : JSON::Any = JSON::Any.new({} of String => JSON::Any)

  def {{ type_dec.var }} : {{ type_dec.type }}
    {{ type_dec.type }}.from_json({{ type_dec.var }}_data.to_json)
  end
end
```

There's a lot of things to consider like how does this look in forms, how does it affect queries, or operations, etc... There's also the fact that it's a pretty big performance hit to convert a JSON::Any object in to a serialized object from JSON. It would be neat if Crystal had that built-in, but until that happens, we'd have to probably just make that as the downside to an easier to work with object. 

I just wanted to jot down some ideas here, and keep track of it. If anyone has some thoughts, let me know.

jwoertink:
Fixes #691 

This is an experimental draft to serializing json columns in to objects. In order to do this, it requires a patch to how the PG shard decodes JSON columns. @matthewmcgarvey submitted https://github.com/will/crystal-pg/pull/232 as a means to fix that hack.

To use them, your migrations would implement the `JSON::Any` as they do now, so nothing changes there. This means existing columns don't need any sort of migration. In your model, you can continue to use `JSON::Any` as is without changing anything. If you want to use the serialized column, you'll change the type from `JSON::Any` to your serializable object type, and then add the column option `serialize: true`.

```crystal
struct UserPreferences
  include JSON::Serializable
  property theme : String = "dark"
  property email_notify : Bool = false
end

class User < BaseModel
  table do
    column preferences : UserPreferences, serialize: true
    column media : JSON::Any
  end
end

user = UserQuery.find(4)
user.preferences.theme == "dark"
user.media["image"]?.try(&.as_s)
```

There's still a few questions left to answer:

* How does the operation handle a serialized column coming in from params?
* What if your column is nullable, but the serialized object has no nilable properties?
* Can we support more complex structures?
* Should there be some sort of runtime escape hatch to get the raw value from the DB of a serialized object?
* Can we provide type-safe compile-time error handling when things aren't setup correctly for this?


src/avram.cr
```diff
@@ -2,6 +2,10 @@ require "dexter"
 require "wordsmith"
 require "habitat"
 require "pulsar"
+require "db"
+require "pg"
+require "uuid"
```
jwoertink, line 7: These had to be moved up so our patch to "pg" comes after the definition

src/avram/model.cr
```diff
@@ -227,7 +237,9 @@ abstract class Avram::Model
     {% else %}
       {% value = nil %}
     {% end %}
-    {% COLUMNS << {name: type_declaration.var, type: data_type, nilable: nilable.id, autogenerated: autogenerated, value: value} %}
+    {% serialized = is_serialized || data_type == "JSON::Any".id %}
```
matthewmcgarvey, line 240: You might experiment, but I think if you brought over the additional changes I made in the `crystal-pg` PR I think you might be able to remove the `JSON::Any` check since it would just work as before

jwoertink: Yup! I just used your branch directly, and it looks like it all still works.

src/avram/charms/json_extensions.cr
```diff
@@ -6,8 +6,13 @@ module JSON::Serializable
   end
 
   module Lucky(T)
+    alias ColumnType = JSON::Any
```
matthewmcgarvey, line 9: What made this necessary?

jwoertink: Oh, that might not be... I was having issues with the Criteria, so I just started adding a ton of stuff until it worked. I'll see if I can remove it.

wontruefree:
I am excited for this feature.  I have been using json more and more these days.

wontruefree:
I think there might also be some query features we can add.

jwoertink:
@wontruefree Have any ideas for an API for query stuff using this? I know one that I'd like to do is some sort of "has_key?" type method. The json query stuff in postgres is pretty gnarly. Also related: https://github.com/luckyframework/avram/issues/197 We could add this in and do a separate PR unless it's something minimal...

paulcsmith (approved):
Lgtm! This is a solid setup to play around with and see how it works

Issue #610, opened 2021-01-14
avram_enum needs better interoperability with Int32, or needs to be stored in DB as an enum.


BrucePerens:
The fact that `avram_enum` is stored in the database as an `Int32` rather than its own type makes it necessary to compare `Enum` with `Int32` a lot. And that breaks, loudly, or worse: silently.

A query containing `.type.nilable_eq(Taxonomy::Type::Language)`, where `type` is an `Int32?` and `Taxonomy::Type::Language` is an `avram_enum` won't compile. Nor will use of `avram_enum` in an interpolated string without `.to_s`, that gets seen as a `Path`.

These compares fail silently when they should not:
```
chldren = o.children.to_a
    categories = o.children.select { |o| o.type == Taxonomy::Type::Category.to_i32 }
    manufacturers = o.children.select { |o| o.type == Taxonomy::Type::Manufacturer.to_i32 }
    products = o.children.select { |o| o.type == Taxonomy::Type::Product.to_i32 }
```

Issue #590, opened 2021-01-04
Unable to set a default value on an enum column


jwoertink:
If you have a model using the `avram_enum`, but you want to set a default value, it fails with a `no overload matches` error.

```crystal
avram_enum State do
  NotStarted
  Started
end

table do
  column state : State = State::NotStarted
end
```

```
> 49 |       def parse(value : AvramState)
> 50 |         SuccessfulCast(State).new(value)
                                      ^--
Error: no overload matches 'Avram::Type::SuccessfulCast(Thing::State).new' with type Thing::AvramState

Overloads are:
 - Avram::Type::SuccessfulCast(T).new(value : T)
```

But then if you try to do something like:

```crystal
table do
  column state : State = State.new(:not_started)
end
```

you'll get

```
> 265 |
> 266 |             parse_result = Thing::State::Lucky.parse(State.new(:not_started))
                                                                 ^----
Error: no overload matches 'Thing::State::Lucky.parse' with type Thing::State

Overloads are:
 - Thing::State::Lucky#parse(value : AvramState)
 - Thing::State::Lucky#parse(value : String)
 - Thing::State::Lucky#parse(value : Int32)
 - Avram::Type#parse(value : Nil)
 - Avram::Type#parse(values : Array(String))
```

Issue #697, opened 2021-07-07
Consider a new interface for Enums


jwoertink:
The ability to use enums in your models is pretty nice, but the current interface did require a small bit of hacking around some of the type restrictions. This has led to using enums being a bit difficult.

Here's the current interface:

```crystal
class GameEvent < BaseModel
  avram_enum Type do
    Connected
    Started
    Ended
    Disconnected
  end
  table do
    column type : GameEvent::Type
  end
end

SaveGameEvent.create!(type: GameEvent::Type.new(:connected))
GameEventQuery.new.type(GameEvent::Type.new(:connected).value)
```

Here's a few of the issues I've found with this setup:

1. The migration requires you to know to create your column type as the Integer. `add type : Int32, default: 0`
2. The model column definition requires the model namespace due to macro stuff. `column type : Type` would fail (probably with a not so pretty error)
3. Saving the enum value requires instantiating this magical class, but querying requires the raw int value
4. If you have a lot of enum members (say 26 like I have in my app), then running SQL and seeing 17 means nothing. (This is more of a preference deal, it may just be a "you're using enums, deal with it" type situation)
5. Getting the enum member is a bit error prone. If your enum member is multi word, you define it like `ControllerActivated`, but to get the value, you either need to use `"ControllerActivated"` or `:controller_activated`. If you use `:controlleractivated`, then you get a pretty bad error that doesn't tell you why. The same goes for if you did `:contoller_activated` (misspelled controller). 

Along with 5 https://cdn.discordapp.com/attachments/743896265057632259/862356608227344427/unknown.png You'll see here it says you can't pass a Symbol. That's because Crystal autocasts Symbol to enum when it matches, so in that case `:super_admin` is treated more like an enum as where `:superadmin` is treated like Symbol. If you're new to Crystal than this is super confusing. 

As for how we solve this, I have no clue... Postgres does support enums directly https://www.postgresql.org/docs/10/datatype-enum.html If we decided to go that route, it would solve 4, but that also means that anytime you wanted to add a new enum member, you'd always have to generate a migration. Adding a new one to your model without the migration would cause issues. Maybe that could be caught by the SchemaEnforcer? 

Another option could be to maybe have an alternate column macro... So it could look like this:

```crystal
class GameEvent < BaseModel
  enum Type
    Connected
    Started
    Ended
    Disconnected
  end
  table do
    enum_column type : Type = Type::Connected
  end
end

SaveGameEvent.create!(type: GameEvent::Type::Connected)
[...]

PR #698, opened 2021-07-09
Use simple enums for Avram columns

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/590
Fixes https://github.com/luckyframework/avram/issues/610

Related to https://github.com/luckyframework/avram/issues/697

I was able to drop the need for having the `avram_enum` macro and allow using regular enums. I'm somewhat worried that by dropping the macro that creates a struct, we will be losing functionality but I can't find anything.

It was added in https://github.com/luckyframework/avram/pull/339 but I don't see any specifics as to why they did what they did.

I did find one problem with `Avram::Criteria` and enums that affects `avram_enums` as well where calling `select_max` would return an `Int32` instead of the enum value.

I think the end goal is to allow storage of enums in the database as something other than `Int32` but I think this PR is necessary because it eliminates a lot of assumptions of complexity before we make more changes.

## Warning

This would be a breaking change. Everyone that uses `avram_enum` in their project would have to change to regular enums though it shouldn't effect anything else because I'd hope no one is using the underlying enum that the macro made. If they did, they will have to update that as well.

### DIFF
--- a/spec/type_extensions/enum_spec.cr
+++ b/spec/type_extensions/enum_spec.cr
@@ -2,12 +2,12 @@ require "../spec_helper"
 
 include LazyLoadHelpers
 
-describe "Enum" do
+describe Enum do
   it "check enum" do
     issue = IssueFactory.create
 
-    issue.status.enum.should eq(Issue::AvramStatus::Opened)
-    issue.role.enum.should eq(Issue::AvramRole::Issue)
+    issue.status.should eq(Issue::Status::Opened)
+    issue.role.should eq(Issue::Role::Issue)
   end
 
   it "update enum" do
@@ -15,8 +15,8 @@ describe "Enum" do
 
     updated_issue = Issue::SaveOperation.update!(issue, status: Issue::Status.new(:closed))
 
-    updated_issue.status.enum.should eq(Issue::AvramStatus::Closed)
-    updated_issue.role.enum.should eq(Issue::AvramRole::Issue)
+    updated_issue.status.should eq(Issue::Status::Closed)
+    updated_issue.role.should eq(Issue::Role::Issue)
   end
 
   it "access enum methods" do
@@ -31,55 +31,5 @@ describe "Enum" do
 
     issue.status.to_s.should eq("Opened")
     issue.status.to_i.should eq(0)
-  end
-
-  it "provides a working ==" do
-    Issue::Status.new(:closed).should eq(Issue::Status.new(:closed))
-    Issue::Status.new(:opened).should_not eq(Issue::Status.new(:closed))
-  end
-
-  it "provides enum-like constant values" do
-    Issue::Status::Closed.should eq(Issue::Status.new(:closed).enum)
-    Issue::Status.new(:closed).enum.should eq(Issue::Status::Closed)
-  end
-
-  it "parses the enum value name" do
-    Issue::Status.new("Closed").should eq(Issue::Status.new(:closed))
-    Issue::Status.new("Opened").should eq(Issue::Status.new(:opened))
-  end
-
-  it "parses the enum value integer when a string" do
-    Issue::Status.new("0").should eq(Issue::Status.new(:opened))
-    Issue::Status.new("1").should eq(Issue::Status.new(:closed))
-  end
-
-  it "implements case equality" do
-    case Issue::Status.new(:closed).value
-    when Issue::Status.new(:closed)
-      true
-    else
-      false
-    end.should be_true
-
-    case Issue::Status::Closed.value
-    when Issue::Status.new(:closed)
-      true
-    else
-      false
-    end.should be_true
-
-    case Issue::Status.new(:opened).value
-    when Issue::Status::Opened
-      true
-    else
-      false
-    end.should be_true
-
-    case Issue::Status::Opened.value
-    when Issue::Status::Opened
-      true
-    else
-      false
-    end.should be_true
   end
 end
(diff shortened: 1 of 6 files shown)


spec/type_extensions/enum_spec.cr
```diff
@@ -2,21 +2,21 @@ require "../spec_helper"
...
-    issue.status.enum.should eq(Issue::AvramStatus::Opened)
-    issue.role.enum.should eq(Issue::AvramRole::Issue)
+    issue.status.should eq(Issue::Status::Opened)
+    issue.role.should eq(Issue::Role::Issue)
   end
 
   it "update enum" do
     issue = IssueFactory.create
 
     updated_issue = Issue::SaveOperation.update!(issue, status: Issue::Status.new(:closed))
```
jwoertink, line 16: Interesting. So we can still instantiate the enum like this? Can we also do `status: Issues::Status::Closed`?

jwoertink (commented):
I love this. I forget why we ended up with the `avram_enum`, but my guess is it may have been an issue with the Type stuff. We've refactored types a bit over time to make them more flexible. 

Do we already have a spec that shows how to query with enums? Also I guess passing a String value in to a SaveOperation. I figured you might have a select dropdown with a value like "closed", but you need to save that as `Issue::Status::Closed`.

spec/type_extensions/enum_spec.cr
matthewmcgarvey: Yep, there's a bunch of ways to instantiate an enum for whatever reason

src/avram/type.cr
```diff
@@ -21,6 +21,10 @@ module Avram::Type
     end
   end
 
+  def parse(value)
+    FailedCast.new
+  end
```
matthewmcgarvey, line 26: This was added because the update to `Avram::Criteria` uses a method that can return a bunch of types. I think a downside of this is that the `parse` methods are how you limit the types that can be passed to query methods which means that any time can be passed into query methods now and they will raise a runtime error

It might be better to add an overload of `parse!` to handle this so that we can keep the type safety at the lower level. It kind of makes sense that the bang method is less safe as well

spec/type_extensions/enum_spec.cr
matthewmcgarvey: So one way of querying that might have worked before but won't work this change is querying with symbols. For example `IssueQuery.new.status(:closed)` will not work. Even though you can do `Issue::Status.new(:closed)`, you can't actually do that through a method where a symbol is passed in.

Here's an example of something I was messing around with in Crystal playground to show the problem

```crystal
enum Color
  Red
  Blue
end

x = Color.new(:red)

def foo(x : Symbol)
  Color.new(x)
end

y = foo(:red)
```

`x` works but `y` causes a compilation error. This seems like it's a bug, not a feature

jwoertink: In this case for `y`, `:red` isn't a `Symbol`, it's a `Color`. So if you do this:

```crystal
enum Color
  Red
  Blue
end

x = Color.new(:red)

def foo(x : Color)
  Color.new(x)
end

y = foo(:red)
```

It'll work. Crystal does the autocasting under the hood to say "You're asking for an Enum, and I see what looks like a Symbol, so I'll cast it.". If you specify Symbol though, crystal says "You're asking for a Symbol, but I need an Enum, so this doesn't work".

src/avram/charms/enum_extensions.cr
```diff
@@ -1,84 +1,53 @@
...
-      def to_db(value : Int32)
-        value.to_s
-      end
+    def criteria(query : V, column) forall V
+      Criteria(V, T).new(query, column)
+    end
 
-      def to_db(value : {{ enum_name }})
-        value.value.to_s
+    class Criteria(T, V) < Int32::Lucky::Criteria(T, V)
```
jwoertink, line 40: Does this mean we restrict Enums to `Int32`? If someone wanted to use `Int16` for example, would this fail?

```crystal
enum Color : Int16
  Red
  Green
end
```

jwoertink: Or, actually.. maybe crystal would upcast because Int16 fits inside of Int32 🤔  I guess then what happens with `Int64`? Or do we just not support other ones?

matthewmcgarvey: To be honest, I had no idea you could change the type like that. This is what we were doing before so I added it back.

Thinking about it, I'm also setting the `Enum::Lucky::ColumnType` to be `Int32` so it's not ok to change the type like that so I think we're fine

spec/type_extensions/enum_spec.cr
matthewmcgarvey: Unfortunately I can't get it to work
CleanShot 2021-07-10 at 13 41 52@2x

matthewmcgarvey: My guess is because it has to go through so many layers of methods

jwoertink: Ah. That's fine I guess. This is still a huge improvement, and we can probably find another way later to make it work that way too.

jwoertink (approved):
I've been trying to think of any case where going this route would be bad, but I think we keep the same level of type-safety (if not better), and it reduces confusion when using them. I say lets roll with it, and see how it goes.

confact:
@matthewmcgarvey good job on this. This has been a thing I have felt bad I introduce myself to Lucky, as it was so hacky. Happy it is gone; my only ugly spot is gone :) 

I think it was very hacky as I couldn't get the module inside Enum to work probably for some reason, with casts and so on. So good you got it working.

PR #700, opened 2021-07-09
Resolve type before using it in migrations

### LINKED ISSUE #689: Confusing error when using incorrect type in migration
When adding a column, I accidentally specified the type as `Boolean` instead of `Bool` and received an error like this:
`Unhandled exception: Negative argument (ArgumentError)`

I would expect a compilation error about the incorrect type.

Example:
```crystal
class AddUsers::V20210625020710 < Avram::Migrator::Migration::V1
  def migrate
    create :users do
      primary_key id : Int64
      add_timestamps
      add admin : Boolean
    end
  end

  def rollback
    drop :users
  end
end

```

matthewmcgarvey:
Fixes #689 

Imagine accidentally doing `add admin : Boolean` where `Boolean` should have actually been `Bool` but you forgot.

Before this change the error you got from that would look like:
CleanShot 2021-07-09 at 16 03 57@2x

With this change, it now looks like:
CleanShot 2021-07-09 at 16 03 31@2x

### DIFF
--- a/src/avram/migrator/alter_table_statement.cr
+++ b/src/avram/migrator/alter_table_statement.cr
@@ -113,16 +113,16 @@ class Avram::Migrator::AlterTableStatement
   end
 
   macro add(type_declaration, index = false, using = :btree, unique = false, default = nil, fill_existing_with = nil, **type_options)
-    {% type = type_declaration.type %}
+    {% type = type_declaration.type.resolve %}
     {% nilable = false %}
     {% array = false %}
     {% should_fill_existing = (!(fill_existing_with == nil)) && (fill_existing_with != :nothing) %}
 
-    {% if type.is_a?(Union) %}
-      {% type = type.types.first %}
+    {% if type.nilable? %}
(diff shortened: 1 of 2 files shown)


src/avram/migrator/alter_table_statement.cr
```diff
@@ -113,16 +113,16 @@ class Avram::Migrator::AlterTableStatement
...
     {% should_fill_existing = (!(fill_existing_with == nil)) && (fill_existing_with != :nothing) %}
 
-    {% if type.is_a?(Union) %}
-      {% type = type.types.first %}
+    {% if type.nilable? %}
+      {% type = type.union_types.reject(&.==(Nil)).first %}
       {% nilable = true %}
     {% end %}
-    {% if type.is_a?(Generic) %}
+    {% if type < Array %}
```
jwoertink, line 125: Oh! Sweet.. That's how you do it 🥳

jwoertink (approved):
Yup! This is sweet. I could never figure out how to best detect Arrays. I love that `type < Array`. That's perfect.

PR #701, opened 2021-07-10
Add uniqueness validation providing query

### LINKED ISSUE #368: Change how `validate_uniqueness_of` query works

matthewmcgarvey:
Fixes #368 
Fixes https://github.com/luckyframework/lucky/issues/1482

Allows for passing a query to the uniqueness validation without specifically passing in a criteria.

If all you wanted to do was limit the uniqueness validation by a different column, before you had to do:

```crystal
validate_uniqueness_of email, query: UserQuery.new.tenant(1).email
```

Now you can do:

```crystal
validate_uniqueness_of email, query: UserQuery.new.tenant(1)
```

It's a small change but it cuts down on confusion and allows users to be clear on what they are actually trying to do.

## Additionally

Other changes were made (I can extract them if necessary)

1. The docs were wrong and have been fixed (on the wrong overloads)
2. The criteria overload was missing the limit to ignore the current record if it was an update
3. Generics were added to the criteria and new overloads to add additional type safety

### DIFF
--- a/src/avram/database_validations.cr
+++ b/src/avram/database_validations.cr
@@ -23,10 +23,10 @@ module Avram::DatabaseValidations(T)
   # database you will not know if there is a collision until all other validations
   # pass and Avram tries to save the record.
   private def validate_uniqueness_of(
-    attribute : Avram::Attribute(V),
-    query : Avram::Criteria(T::BaseQuery, V),
+    attribute : Avram::Attribute,
+    query : Avram::Criteria(T::BaseQuery, _),
     message : String = "is already taken"
-  ) forall V
+  )
     attribute.value.try do |value|
       if limit_query(query.eq(value)).first?
         attribute.add_error message
--- a/src/avram/database_validations.cr
+++ b/src/avram/database_validations.cr
@@ -5,8 +5,11 @@ module Avram::DatabaseValidations(T)
   #
   # > This will only work with attributes that correspond to a database column.
   #
+  # Passing a criteria is useful when wanting to change how it compares the value in the database.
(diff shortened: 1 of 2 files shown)


src/avram/database_validations.cr
```diff
@@ -21,11 +24,11 @@ module Avram::DatabaseValidations(T)
   # pass and Avram tries to save the record.
   private def validate_uniqueness_of(
     attribute : Avram::Attribute,
-    query : Avram::Criteria,
+    query : Avram::Criteria(T::BaseQuery, _),
```
matthewmcgarvey, line 27: I had added generics to the Attribute and the Criteria to connect them. I think it might have worked but with my recent work on enums, I'm not sure if the types of the two would be the same so I took it off to be safe.

jwoertink (approved):
This looks good to me! I saw your note about the generic on the Criteria. I'm not sure either if the types of the two would be the same, but it's probably fine how you have it. We will know better for sure after running through some full testing building out a sample app and stuff.

PR #702, opened 2021-07-10
Allow using cascade when truncating tables

### LINKED ISSUE #648: Add a CASCADE option to TRUNCATE
https://github.com/luckyframework/avram/blob/5e29f75371dca5a2bc16858163dc3790cabfbfcb/src/avram/queryable.cr#L36-L39

It would be cool to have an overload that allows us to pass an optional `cascade` option to `truncate`. Right now, if you have a table with a foreign key constraint, and you try to truncate it, you'll get a [postgres error](https://stackoverflow.com/questions/56700466/postgresql-truncate-table-with-foreign-key-constraint/56700574) mentioning 

> cannot truncate a table referenced in a foreign key constraint

To get around this, you have 2 options.

1. call `TRUNCATE TABLE the_table, other_table, one_more_related`
2. call `TRUNCATE TABLE the_table CASCADE` which will handle the first for you.

The second option could be doable just by adding a flag.

For a work around, just call `MyQuery.new.delete` which will just delete all of the records. 

Note: We are already including the `CASCADE` when truncating the whole db https://github.com/luckyframework/avram/blob/master/src/avram/database.cr#L202... I'm not sure the impact, but maybe we can just always CASCADE?

jaitaiwan:
In cases where a data in a database has foreign key references against
the table we're wanting to truncate, we can use the CASCADE key to
ensure that all data referencing the rows we're deleting are also
deleted to ensure the database keeps it's integrity.

 - Added in-code documentation as per crystal spec for the truncate
   command
 - Added new cascade flag to truncate method, set to "false" by default
   (potentially dangerous)
 - Added migration ensuring foreign key constraint on posts by comments
 - Added specification to test cascade deletes data referenced by
   foreign keys

 Fixes #648
 
 Some possible considerations:
- Perhaps we should turn on cascade by default considering it's potentially a dangerous state to truncate ignoring foreign key constraints? However I know that cascade by default would delete even more data so perhaps that's unwarranted
- Technically speaking the spec I added is really only testing if Postgres is doing a cascading delete, I'd feel much more comfortable if the resulting ExecResult was inspectable to see what the final query that was run was so that I could do an assertion against it.


src/avram/queryable.cr
```diff
@@ -36,9 +36,19 @@ module Avram::Queryable(T)
...
+    # Removes all data from a table using the TRUNCATE postgres SQL command.
+    #
+    # You should not run this command without `cascade: true` if your table
+    # columns are referenced by other foreign key constraints. Use *delete*
+    # instead if you don't want to accidentally delete rows referenced
+    # elsewhere.
+    #
+    # To delete all data referenced by foreign keys as well, set *cascade*
+    # to true.
+    def self.truncate(cascade : Bool = false)
```
jwoertink, line 48:
```suggestion
    def self.truncate(*, cascade : Bool = false)
```

With this, it enforces that a named arg is always used. This will prevent someone using `Thing.truncate(true)` in their codebase.

src/avram/queryable.cr
```diff
@@ -36,9 +36,19 @@ module Avram::Queryable(T)
       new.last?
     end
 
-    def self.truncate
+    # Removes all data from a table using the TRUNCATE postgres SQL command.
+    #
+    # You should not run this command without `cascade: true` if your table
```
jwoertink, line 41:
```suggestion
    # You should run this command with `cascade: true` if your table
```

jwoertink (changes requested):
Sweet! Great job on this. I think you just need to run the formatter `crystal tool format spec src`, and make sure ameba is happy `./bin/ameba`. I made a few suggestion, but once those are done, I think this will be good!

src/avram/queryable.cr
jaitaiwan: Happy to make the change. Would you mind explaining what the rule or rationale is for this?

jwoertink (approved):
Sweet! Thanks for the contribution.

PR #703, opened 2021-07-10
Add any? and none? as query methods

### LINKED ISSUE #680: Add any? and none? query methods
These would be methods to check if there are any/no records in the database.

```crystal
if UserQuery.new.email("admin@example.com").none?
  CreateUser.create!(email: "admin@example.com", password: "test123")
end
```

https://guides.rubyonrails.org/active_record_querying.html#existence-of-objects

`none?` would just be `!any?`

matthewmcgarvey:
Fixes #680 

Provides two simple methods to check if there are any records in the database matching the specified conditions. Within avram's codebase, it simplifies the uniqueness checks.

Usage looks like:

```crystal
UserQuery.any?
UserQuery.new.name("Sally").any?
```

### DIFF
--- a/src/avram/database.cr
+++ b/src/avram/database.cr
@@ -74,19 +74,19 @@ abstract class Avram::Database
   end
 
   # Methods without a block
-  {% for crystal_db_alias in [:exec, :scalar, :query, :query_all, :query_one, :query_each] %}
+  {% for crystal_db_alias in [:exec, :scalar, :query, :query_all, :query_one, :query_one?, :query_each] %}
     # Same as crystal-db's `DB::QueryMethods#{{ crystal_db_alias.id }}` but with instrumentation
     def {{ crystal_db_alias.id }}(query, *args_, args : Array? = nil, queryable : String? = nil, **named_args)
       publish_query_event(query, args_, args, queryable) do
         run do |db|
-          db.{{ crystal_db_alias.id }}(query, *args_, args: args)
+          db.{{ crystal_db_alias.id }}(query, *args_, **named_args, args: args)
         end
       end
     end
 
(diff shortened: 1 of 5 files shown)


src/avram/database.cr
```diff
@@ -74,19 +74,19 @@ abstract class Avram::Database
...
 
   # Methods without a block
-  {% for crystal_db_alias in [:exec, :scalar, :query, :query_all, :query_one, :query_each] %}
+  {% for crystal_db_alias in [:exec, :scalar, :query, :query_all, :query_one, :query_one?, :query_each] %}
     # Same as crystal-db's `DB::QueryMethods#{{ crystal_db_alias.id }}` but with instrumentation
     def {{ crystal_db_alias.id }}(query, *args_, args : Array? = nil, queryable : String? = nil, **named_args)
       publish_query_event(query, args_, args, queryable) do
         run do |db|
-          db.{{ crystal_db_alias.id }}(query, *args_, args: args)
+          db.{{ crystal_db_alias.id }}(query, *args_, **named_args, args: args)
```
matthewmcgarvey, line 82: Because we haven't used them before in the codebase, there was a bug here that meant we couldn't call `query_all` or `query_one` and the added `query_one?`. It expects an `as` named argument that passes in what it should be converted into

jwoertink (approved):
Sweet! This is a pretty nice method. It seems like it would actually run faster than doing `.first?` checks. Out of curiosity, have you tried any benchmarks?
Kind of a bummer we will have to start disabling the ameba checks around this, but I get it.

This all looks good 👍

matthewmcgarvey:
Ran a benchmark comparing `first?` to `any?`

CleanShot 2021-07-10 at 15 29 34@2x

PR #704, opened 2021-07-10
Consolidate nothing classes into one

### LINKED ISSUE #673: Move Nothing to a top level Avram namespace
Getting an error that looks like this is a bit jarring, especially to newcomers:

```
| Error: no overload matches 'SaveFortune.new' with types Fortune, Lucky::Params, id: Avram::NeedyInitializerAndSaveMethods::Nothing, created_at: Avram::NeedyInitializerAndSaveMethods::Nothing, updated_at: Avram::NeedyInitializerAndSaveMethods::Nothing, text: Avram::NeedyInitializerAndSaveMethods::Nothing, user_id: Avram::NeedyInitializerAndSaveMethods::Nothing
web          |
```

At least if it just says `Avram::Nothing`, it'll be a bit easier to parse

matthewmcgarvey:
Fixes #673 

Hopefully the comment on the `Avram::Nothing` class does an ok job at explaining its existence.

### DIFF
--- a/src/avram/nothing.cr
+++ b/src/avram/nothing.cr
@@ -1,5 +1,6 @@
 # This class is used in various places
 # where the question of "Did I not pass in anything or did I pass in nil?"
 # needs to be answered.
+# :nodoc:
 class Avram::Nothing
 end
--- a/src/avram/nothing.cr
+++ b/src/avram/nothing.cr
@@ -1,0 +1,5 @@
+# This class is used in various places
+# where the question of "Did I not pass in anything or did I pass in nil?"
+# needs to be answered.
+class Avram::Nothing
+end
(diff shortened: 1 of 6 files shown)


src/avram/nothing.cr
```diff
@@ -0,0 +1,5 @@
+# This class is used in various places
+# where the question of "Did I not pass in anything or did I pass in nil?"
+# needs to be answered.
+class Avram::Nothing
```
jwoertink, line 4: Lets add a `:nodoc:` here so this doesn't get generated in the docs. This really should be a private class since it's not meant to be used externally, but making it `private` I think would break..

jwoertink (commented):
This is a nice cleanup. I think just the nodoc tag and it's good to merge when you're ready.

Issue #699, opened 2021-07-09
Migrations don't like when a view model is empty


jwoertink:
I generated a model, then changed the `table` to `view`, but left the columns empty. Then I wrote out my view migrations, and when I tried to migrate, it blew up.

```
❯ lucky db.migrate
There was a problem expanding macro 'setup_db_mapping'

Code in macro 'setup'

 1 | Avram::Model.setup_db_mapping(
     ^
Called macro defined in lib/avram/src/avram/model.cr:163:3

 163 | macro setup_db_mapping(columns, *args, **named_args)

Which expanded to:

 > 1 | DB.mapping({
                  ^
Error: for empty hashes use '{} of KeyType => ValueType'
```

```crystal
class Report < BaseModel
  view do
  end
end
```

PR #706, opened 2021-07-12
Require at least one column defined on models

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/699

CleanShot 2021-07-12 at 16 35 35@2x

### DIFF
--- a/src/avram/model.cr
+++ b/src/avram/model.cr
@@ -50,6 +50,7 @@ abstract class Avram::Model
 
     {{ yield }}
 
+    validate_columns
     validate_primary_key
 
     {% if table_name %}
@@ -72,6 +73,7 @@ abstract class Avram::Model
   macro view(view_name = nil)
     {{ yield }}
 
+    validate_columns
     {% if view_name %}
       class_getter table_name : String = {{ view_name.id.stringify }}
     {% else %}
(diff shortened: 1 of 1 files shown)


src/avram/model.cr
```diff
@@ -72,6 +73,7 @@ abstract class Avram::Model
   macro view(view_name = nil)
     {{ yield }}
 
+    validate_columns
```
matthewmcgarvey, line 76: I considered moving this macro into the `setup` list below but if I did that it would have added a layer to the error message that pointed at macro code so I thought this makes the error easier to understand

src/avram/model.cr
```diff
@@ -147,6 +149,12 @@ abstract class Avram::Model
     )
   end
 
+  macro validate_columns
+    {% if COLUMNS.empty? %}
+      {% raise "#{@type} must define at least one column." %}
```
jwoertink, line 154: Can we add the typical `Try this...` example here too? Like `column my_column : String` would be fine I think.

matthewmcgarvey: CleanShot 2021-07-12 at 18 16 50@2x

Updated it to match the [`validate_primary_key` error](https://github.com/luckyframework/avram/blob/98ed8b7924356a2b2999c5ef34db637d3856f8a3/src/avram/model.cr#L111-L125)

jwoertink (approved):
Looks good!

Issue #694, opened 2021-07-01
Change DeleteOperation#destroy to #delete


paulcsmith:
https://github.com/luckyframework/lucky/discussions/1533

PR #707, opened 2021-07-13
Rename DeleteOperation#destroy to #delete

matthewmcgarvey:
Fixes https://github.com/luckyframework/avram/issues/694

### DIFF
--- a/src/avram/needy_initializer_and_delete_methods.cr
+++ b/src/avram/needy_initializer_and_delete_methods.cr
@@ -30,14 +30,14 @@ module Avram::NeedyInitializerAndDeleteMethods
     macro finished
       # This is called at the end so @type will be of the subclass,
       # and not the parent abstract class.
-      generate_initializer_and_destroy_methods
+      generate_initializer_and_delete_methods
     end
   end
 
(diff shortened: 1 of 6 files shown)


src/avram/needy_initializer_and_delete_methods.cr
```diff
@@ -91,22 +91,8 @@ module Avram::NeedyInitializerAndDeleteMethods
...
-        DeleteOperations do not have a 'delete' method.
-
-        Try this...
-
-          ▸ Use 'destroy' to delete a record
-
-        ERROR
-        %}
-      {% end %}
-    end
```
jwoertink, line 107: We should leave this in, but name it `destroy` with a message like "destroy has been renamed to delete". Maybe with the type name or something so it's easy to find which ones you have to replace?

matthewmcgarvey: I added it in, unfortunately I had to move the compilation error because `@type` was not the inherited type and it was a mess of macro code in the backtrace.

This is something that I hope can be removed in the future so I don't think it's a big deal.

jwoertink (approved):
👍

PR #723, opened 2021-08-07
fix: add the operation name to the unpermittable column error

### LINKED ISSUE #719: Add the operation name to the unpermittable column error
https://github.com/luckyframework/avram/blob/8f6301c169f8ca80b65a08de7f60a00d11cc791e/src/avram/save_operation.cr#L263

This error doesn't really tell you *which* model is missing "title", or which operation you're trying to permit this column in. Specifically in Avram, this requires a lot of digging around. It seems we can use `@type` to at least get the operation name and that at least narrows it down a bit.

We could update this error to:

```diff
- Error: Can't permit 'title' because the column has not been defined on the model.
+ Error: Can't permit 'title' because the column has not been defined on the model for #{@type}.
```

RhnSharma:
Fixes #719
Hi @jwoertink, how does this look?
Let me know if this needs any changes.
Thanks

### DIFF
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -260,7 +260,7 @@ abstract class Avram::SaveOperation(T)
         @@permitted_param_keys << "{{ attribute_name.id }}"
       {% else %}
         {% raise <<-ERROR
-          Can't permit '#{attribute_name}' because the column has not been defined on the model #{@type}.
+          Can't permit '#{attribute_name}' because the column has not been defined on the model for #{@type}.
 
           Try this...
 
--- a/src/avram/save_operation.cr
+++ b/src/avram/save_operation.cr
@@ -260,7 +260,7 @@ abstract class Avram::SaveOperation(T)
         @@permitted_param_keys << "{{ attribute_name.id }}"
       {% else %}
         {% raise <<-ERROR
-          Can't permit '#{attribute_name}' because the column has not been defined on the model.
+          Can't permit '#{attribute_name}' because the column has not been defined on the model #{@type}.
 
           Try this...
 


src/avram/save_operation.cr
```diff
@@ -260,7 +260,7 @@ abstract class Avram::SaveOperation(T)
         @@permitted_param_keys << "{{ attribute_name.id }}"
       {% else %}
         {% raise <<-ERROR
-          Can't permit '#{attribute_name}' because the column has not been defined on the model.
+          Can't permit '#{attribute_name}' because the column has not been defined on the model #{@type}.
```
jwoertink, line 263: This isn't quite right since `@type` isn't a model, it's the SaveOperation class. To test the error message, look for any SaveOperation in the specs, and add any random word to the `permit_columns` like `banana` or something. Then try to run the specs, and you'll see the error message.

This will need to say something more like `defined on the model for SaveUser`

jwoertink (changes requested):
Hey! Thanks for taking this on. I left a comment just saying that simply adding the type makes the error message a little confusing. I'd recommend giving it a shot to raise the error so you can see what I mean. After that update, it should be good 👍

jwoertink:
Hey @RhnSharma  just wanted to ping you. There's just a small change to your PR, then we can merge it in when you get a free moment. Thanks!

src/avram/save_operation.cr
RhnSharma: avram-SaveUser-error
@jwoertink Is this a valid error? Or do I need to make any changes?  Could you please guide me a little?
Thanks

jwoertink: That's a good error 👍

RhnSharma: I have made the changes. Please take a look at it and let me know if this needs any changes.
Thanks

jwoertink (approved):
Awesome! Thank you for making this change.

PR #724, opened 2021-08-24
Fix wrong command in help messages

akadusei:
Correct command is `db.migrations.status` ("migration**s**" <-- plural)

Originates from commit f3d6328d25732aaa6359b8e9f7a670df2aef5b41.


jwoertink (approved):
Oops! Good catch :joy:  Thanks for fixing this.

PR #725, opened 2021-09-02
Only set default UUID value on the SQL side

### LINKED ISSUE #714: Remove crystal side UUID generation
https://github.com/luckyframework/avram/blob/e638ffb8eb5e8088449dfa63cb61a0739212f38d/src/avram/save_operation_template.cr#L15-L22

The migrator will set the default UUID value to `gen_random_uuid()` now. This means we can remove it from the crystal side which will fix the issue of `new_record?` always returning false when using UUID.

Ref: https://github.com/luckyframework/avram/issues/393

jwoertink:
Fixes #714

### DIFF
--- a/spec/schema_enforcer/ensure_uuid_primary_key_has_default_spec.cr
+++ b/spec/schema_enforcer/ensure_uuid_primary_key_has_default_spec.cr
@@ -6,7 +6,7 @@ private class TableWithoutUUIDPrimaryKey < BaseModel
 end
 
 private class TableWithoutDefaultUUID < BaseModel
-  table :prices do
+  table :unused_uuid_table do
   end
 end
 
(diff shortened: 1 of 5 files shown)


spec/schema_enforcer/ensure_uuid_primary_key_has_default_spec.cr
```diff
@@ -23,7 +23,7 @@ describe Avram::SchemaEnforcer::EnsureUUIDPrimaryKeyHasDefault do
   end
 
   it "raises when table's uuid primary key does not have default" do
-    expect_schema_mismatch "Primary key on the 'prices' table has the type set as uuid but does not have a default value." do
+    expect_schema_mismatch "Primary key on the 'unused_uuid_table' table has the type set as uuid but does not have a default value." do
```
jwoertink, line 26: I had to make a new table to test this because `prices` gets used in other instances, and was throwing some invalid errors for not having an `id` value.

matthewmcgarvey (approved):

PR #733, opened 2021-09-30
Add new includes method for Array queries.

### LINKED ISSUE #640: New method to query for value inside of array column
If you have an array column, you may want to search for records that have a specific value inside of that array. Right now we have to use a raw `where` to do this.

```crystal
# currently
UserQuery.new.where("? = ANY (users.tags)", "lucky")

# proposed
UserQuery.new.tags.contains("lucky")
# or maybe
UserQuery.new.tags.includes("lucky")
```

This method should raise a compile-time error when called against any column that's not an array.

**EDIT:** I should also note that depending on the array generic type, you should only be allowed to pass that type in...

```crystal
# assumes column friends : Array(UUID)
UserQuery.new.friends.contains(UUID.random) # this is ok

UserQuery.new.friends.contains(9) # this should fail
```

We could possibly make an exception for `Array(String)`, though, that might be too magical

jwoertink:
Fixes #640

I'm only adding the `includes` method for now, and with somewhat limited support. One issue I would like to figure out before going full support is how to catch this method at compile-time so you only call it on Array columns. I tried several different patterns, but there's a few issues with it that I'll open up in a separate issue.

This PR allows you query for records with Array columns where a specific value exists in that column.

```crystal
# column blocked_on_pages : Array(Int64)
UserQuery.new.blocked_on_pages.includes(current_page.id).any?
```

If you try to call this on a non-array column, Postgres will throw a runtime error when you run the query.


src/avram/criteria_extensions/includes_criteria.cr
```diff
@@ -0,0 +1,27 @@
...
+      value = V.adapter.to_db!(value)
+      add_clause(Avram::Where::Includes.new(column, value))
+    end
+
+    # TODO: Figure out how we can ensure this method
+    # is only called on array columns at compile-time.
+    private def check_using_array!(_value : Array.class)
+    end
+
+    private def check_using_array!(_value)
```
jwoertink, line 14: These aren't actually used right now, but I wanted to leave them in as a template to the compile-time pattern. I do plan on eventually figuring out how to use it.

src/avram/criteria_extensions/includes_criteria.cr
```diff
@@ -0,0 +1,27 @@
+module Avram::IncludesCriteria(T, V)
+  macro included
+    # WHERE `value` = ANY(column)
+    def includes(value) : T
```
jwoertink, line 4: I also debated on `contains`, but with `includes`, we could also add an `excludes` method... What would be the opposite of `contains`, `omits`? 🤷‍♂️

robacarp: The Avram DSL here is close enough to sql that I think includes makes sense.

src/avram/criteria_extensions/includes_criteria.cr
```diff
@@ -0,0 +1,27 @@
...
+      value = V.adapter.to_db!(value)
+      add_clause(Avram::Where::Includes.new(column, value))
+    end
+
+    # TODO: Figure out how we can ensure this method
+    # is only called on array columns at compile-time.
```
robacarp, line 10:
```suggestion
    # is only called on array values at compile-time.
```

right?

src/avram/criteria_extensions/includes_criteria.cr
robacarp: Actually, I misunderstood the purpose of this when I said that. But I still think that includes is as good a name as any. Sorry for the email spam from my github comments 🤪

jwoertink: 😂  cool. Second confirmation is what I needed. Thanks

PR #737, opened 2021-10-03
If there are pending migrations, tell the app to shut down

### LINKED ISSUE #671: Make pending migration error a little more noticable
With so much going on booting a Lucky app, it's easy to miss that there's some pending migrations. 

I think we can do better.

Screen Shot 2021-04-27 at 2 02 16 PM

jwoertink:
Fixes #671
Fixes https://github.com/luckyframework/lucky/issues/1142

This PR adds a front and center error when there's pending migrations.

## After PR:
Screen Shot 2021-10-03 at 9 24 56 AM

there is one minor issue still here which is if you boot your app locally without a process manager (e.g. `crystal src/start_server.cr`), it will try to send interrupt signals to the parent process.
Screen Shot 2021-10-03 at 9 25 35 AM

Most likely this isn't a big issue. You can even just run `lucky watch` and it's fine. It's only when booting the app manually through crystal.

### DIFF
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -172,8 +172,23 @@ class Avram::Migrator::Runner
 
   def ensure_migrated!
     if !pending_migrations.empty?
-      raise "There are pending migrations. Please run lucky db.migrate"
+      puts ""
+      puts error_background
+      puts pending_migrations_error.colorize.on_red.white
+      puts error_background
+      puts ""
+      Process.signal(:term, Process.ppid)
+      Process.signal(:hup, Process.ppid)
+      exit 1
     end
+  end
+
+  private def pending_migrations_error
+    " There are pending migrations. Please run lucky db.migrate "
+  end
+
+  private def error_background
+    (" " * pending_migrations_error.size).colorize.on_red
   end
 
   private def migrated_migrations


src/avram/migrator/runner.cr
```diff
@@ -172,10 +172,25 @@ class Avram::Migrator::Runner
...
     if !pending_migrations.empty?
-      raise "There are pending migrations. Please run lucky db.migrate"
+      puts ""
+      puts error_background
+      puts pending_migrations_error.colorize.on_red.white
+      puts error_background
+      puts ""
+      Process.signal(:term, Process.ppid)
+      Process.signal(:hup, Process.ppid)
+      exit 1
```
jwoertink, line 182: This tells the process manager to shut down, then exit the program with an error code 1. If the parent process doesn't respond to the `term` and `hup` signals, it will call the exit, but then hang. However, you'll still see the bright red banner which is still better than it currently is.

robacarp: The idea here is that we send both term and hup, so that one of those might make a difference? Do we know that both are needed? I would expect term to be sufficient.

src/avram/migrator/runner.cr
```diff
@@ -172,10 +172,25 @@ class Avram::Migrator::Runner
 
   def ensure_migrated!
     if !pending_migrations.empty?
-      raise "There are pending migrations. Please run lucky db.migrate"
+      puts ""
+      puts error_background
+      puts pending_migrations_error.colorize.on_red.white
+      puts error_background
+      puts ""
```
robacarp, line 179: Should this just be a heredoc?

src/avram/migrator/runner.cr
jwoertink: I actually got the suggestion from someone else, I can try just term and see if that's sufficient though.

src/avram/migrator/runner.cr
jwoertink: I'm not sure... I'll have to see what `puts " ".colorize.on_red.white` does 🤔

jwoertink: lol 😂  my editor keeps stripping out the blank space in the heredoc. But it does work ok if I do something like this:

```crystal
def blank_space
  " " * 50
end

puts <<-MESSAGE.colorize.on_red.white
#{blank_space}
 WARNING THERE IS A THING GOING ON HERE
 #{blank_space}
MESSAGE
```

What are your thoughts? I'm indifferent on this.

robacarp: If your editor is doing it, it's probably something to do with crystal fmt -- that tool annihilates whitespace wherever it can. I'm just nitpicking, I don't think it matters! 😆 Crystal tool fmt and I don't get along most of the time...

src/avram/migrator/runner.cr
jwoertink: ok, looks like term is enough. It seems `TERM` and `HUP` are pretty much the same in this scenario since no one would be booting an app in a screen then trying to close their terminal 😂

matthewmcgarvey (approved):

robacarp:
Does this set of changes alter how a booting production-mode app would behave when database migrations are pending?

jwoertink:
Well, by default, this would only [run in development](https://github.com/luckyframework/lucky_cli/blob/8a9c089e9ad7dcc32102bd8c544cbee7bb092f20/src/web_app_skeleton/src/start_server.cr.ecr#L5-L8). For production, you have a few different options of either building the migrate task and manually running that, or changing the linked code to look more like:

```crystal
if LuckyEnv.development?
  Avram::Migrator::Runner.new.ensure_migrated!
  Avram::SchemaEnforcer.ensure_correct_column_mappings!
else
  Avram::Migrator::Runner.new.run_pending_migrations
end
```

but the `ensure_migrated!` method would still only run for development.

Issue #727, opened 2021-09-10
Time columns don't include milliseconds


jwoertink:
When we set the `created_at` and `updated_at` fields, we just called `created_at.value = Time.utc`. That value in the DB looks like `2021-09-10 18:08:39+00`, but if you actually inspect `Time.utc` you'll get something that looks like `2021-09-10 18:03:38.655286917 UTC`. Storing the milliseconds makes a HUGE difference when you're sorting by created_at in your specs especially if you're doing any window function using created_at.

We need to figure out how to include milliseconds on these fields by default.

Related: https://github.com/luckyframework/avram/issues/393 if we just let SQL handle it via `NOW()`, it would be fine.

*EDIT*: turns out it's all Time columns, not just `created_at` and `updated_at`.

jwoertink:
This happens here:
https://github.com/luckyframework/avram/blob/a3d830cac03f035a299e9c8fa6d099e43348b11b/src/avram/charms/time_extensions.cr#L74-L76

```crystal
Time.utc.to_s                            #=> "2021-09-10 19:21:22 UTC"
Time.utc.to_s("%F %X.%6N") #=> "2021-09-10 19:21:22.872793"
```

We account for time formats when pulling the value out of the DB, but not for inserting the value.

PR #738, opened 2021-10-03
Add compile-time catches for common method call mistakes with attributes

### LINKED ISSUE #731: Better error message when calling a method on an attribute but meant to call on the value
Especially when it comes to dealing with JSON

```
web          | Error: undefined method '[]?' for Avram::Attribute(JSON::Any) (compile-time type is Avram::Attribute(JSON::Any)+)
```

You might get this error, meaning you called `some_attr["whatever"]`, but you mean to call `some_attr.value["whatever"]`. It would be cool if we could catch a few common ones like this. Maybe we define `def []` on Attribute that just does a compile-time error?

Here's a list of ones I think might be common

* `to_s` (if you want to inspect an attribute call `inspect`)... NOTE: this could be dangerous, would need confirmation that it doesn't mess with Crystal directly.
* `to_i`, `to_i32`, `to_i64`
* `[]`
* `to_f`, `to_f64`

jwoertink:
Fixes #731

This doesn't really catch *all* possibilities, but should catch the bulk of what you might call on your attributes inside of an operation. We found ourselves mainly doing this on JSON attributes.. `metadata["key"]?` instead of `metadata.value["key"]?`.

### DIFF
--- a/src/avram/attribute.cr
+++ b/src/avram/attribute.cr
@@ -102,7 +102,7 @@ class Avram::Attribute(T)
   # These methods may accidentally get called on attributes
   # inside of operations. Since these methods don't exist,
   # chances are, you meant to call them on the value.
-  # ```crystal
+  # ```
   # username.to_s
   # # VS
   # username.value.to_s
--- a/src/avram/attribute.cr
+++ b/src/avram/attribute.cr
@@ -98,4 +98,64 @@ class Avram::Attribute(T)
       add_error("is invalid")
     end
   end
+
+  # These methods may accidentally get called on attributes
+  # inside of operations. Since these methods don't exist,
+  # chances are, you meant to call them on the value.
+  # ```crystal
+  # username.to_s
+  # # VS
+  # username.value.to_s
+  # ```
+  def to_s
+    call_value_instead_error_message(".to_s")
+  end
+
+  # NOTE: to_s(io : IO) is used when passing an object
+  # in to string interpolation. Don't override that method.
+  def to_s(time_format : String)
+    call_value_instead_error_message(".to_s(time_format)")
+  end
+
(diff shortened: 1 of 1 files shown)


jwoertink:
I'm 99% sure the spec failure is a race condition related to https://github.com/luckyframework/avram/issues/727

robacarp (approved):

wontruefree (approved):

src/avram/attribute.cr
```diff
@@ -98,4 +98,64 @@ class Avram::Attribute(T)
...
+  # ```
+  def to_s
+    call_value_instead_error_message(".to_s")
+  end
+
+  # NOTE: to_s(io : IO) is used when passing an object
+  # in to string interpolation. Don't override that method.
+  def to_s(time_format : String)
+    call_value_instead_error_message(".to_s(time_format)")
+  end
```
robacarp, line 118: I think these are a great addition. If there's one mistake I make _all the time_ it's calling these methods accidentally. 

This specific overload makes me a bit nervous because of the way the name of the string parameter would confuse someone who happened upon this error message through some secondary path that had nothing to do with a time format. 

What if the stringified method suggestion in the error message is just `.to_s(...)` instead? That seems like it would lead people down an almost identical path without lending confusion where they landed here without doing anything with time formats.

jwoertink: I like that 👍  Makes sense.

jwoertink:
So today I actually called `if attr.nil?` when I meant `attr.value.nil?`... However, what if you really wanted to know if the attribute itself was nil? I'm not sure if I should add that one or not 🤔

matthewmcgarvey (approved):

PR #749, opened 2021-10-15
Custom setting for Time format

### LINKED ISSUE #727: Time columns don't include milliseconds

jwoertink:
Fixes #727

This PR allows you to set a custom format that will be used on all `Time` columns. As of right now, we have no way to do this on a per-column level. It's all or nothing.

I'm opening this as a draft because I'm still not sold on going this route. The fear is this gives people the wide open ability to just do `default_time_format = "%Y"` which would set all `created_at`, and `updated_at` to just the year, and would break all html forms using time inputs. Maybe we could just have a `include_milliseconds = true`? I'm not sure, but I'll keep thinking about it.

### DIFF
--- a/src/avram.cr
+++ b/src/avram.cr
@@ -21,6 +21,8 @@ module Avram
     setting lazy_load_enabled : Bool = true
     setting database_to_migrate : Avram::Database.class, example: "AppDatabase"
     setting time_formats : Array(String) = [] of String
+    # This is the default format for storing the Time
+    setting default_time_format : String? = nil, example: "%F %X.%6N %:z"
   end
 
   Log            = ::Log.for(Avram)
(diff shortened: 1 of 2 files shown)


jwoertink:
Decided I'm going to close this. I'll open a separate PR with a different update

PR #751, opened 2021-10-16
Added new default_validations to Operations

### LINKED ISSUE #645: polymorphic validator runs before before_save methods
The code for validating that polmorphics are set correctly doesn't run in the intended order, because of the use of `finished` here. It runs before a `before_save` method in my operation where I set the value of the polymorphic, and thus the error is already set before I set the polymorphic.
https://github.com/luckyframework/avram/blob/master/src/avram/polymorphic.cr#L141
```crystal
macro finished
      class SaveOperation
        before_save do
          {% list_of_foreign_keys = associations.map(&.id).map { |assoc| "#{assoc.id}_id".id } %}

          # TODO: Needs to actually get the foreign key from the ASSOCIATIONS constant
          {% if optional %}
            validate_at_most_one_filled {{ list_of_foreign_keys.map(&.id).join(", ").id }}
          {% else %}
            validate_exactly_one_filled {{ list_of_foreign_keys.map(&.id).join(", ").id }},
              message: "at least one '{{ polymorphic_name.id }}' must be filled"
          {% end %}
        end
```

jwoertink:
Fixes #645

The original issue was that our polymorphic `before_save` callback would run validations before any that you would set in your own SaveOperation. This prevented you from being able to assign your association in your SaveOperation.

This PR adds a new functionality to operations (SaveOperation, Operation, and DeleteOperation) to set your own `default_validations`. It works similar to `before_save` in that it takes a block, and you can run some code in there. When the operations call `valid?` it will run any required validations (as it always has), as well as any default validations you might need. This means you can also abstract them to a module to include them in several different operations if you need.

```crystal
module DefaultThingValidations
  macro included
    default_validations do
      validate_required thing
    end
  end
end

class SaveThing1 < Thing1::SaveOperation
  include DefaultThingValidations
  before_save do
    # this is called before the default validations
    # so you can do any setup you need
    thing.value = "thing 1"
  end
end

class SaveThing2 < Thing2::SaveOperation
  include DefaultThingValidations
  before_save do
    thing.value = "thing 2"
  end
end
```

### DIFF
--- a/spec/operations/save_operation_spec.cr
+++ b/spec/operations/save_operation_spec.cr
@@ -87,7 +87,34 @@ private class AllowBlankComment < Comment::SaveOperation
   end
 end
 
+module DefaultUserValidations
+  macro included
+    property starts_nil : String? = nil
+
+    default_validations do
+      self.starts_nil = self.starts_nil.to_s + "!"
+      validate_required nickname
+    end
+  end
+end
+
+private class UserWithDefaultValidations < User::SaveOperation
+  include DefaultUserValidations
+
+  before_save do
+    self.starts_nil = "not nil"
+  end
+end
+
 describe "Avram::SaveOperation" do
+  it "calls the default validations after the before_save" do
+    UserWithDefaultValidations.create(name: "TestName", nickname: "TestNickname", joined_at: Time.utc, age: 400) do |op, u|
+      op.starts_nil.should eq("not nil!")
+      u.should_not be_nil
+      u.not_nil!.nickname.should eq("TestNickname")
+    end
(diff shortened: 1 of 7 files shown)


paulcsmith:
I think this is a goo idea to fix this issue. One thing that may be helpful is to add something to the docs and/org guides to explain when to use this. I read through: https://luckyframework.org/guides/database/callbacks-and-validations#default-validations and I wasn't sure when to use this over `before_save` until I came to this PR. What do you think? 

Definitely not a blocker, but just something I was thinking when I was going through the docs

PR #753, opened 2021-10-21
New Query Cache Feature

### LINKED ISSUE #63: Add a query cache

jwoertink:
Fixes #63 

This PR adds in query cache. The idea is that if your app runs a query more than once, it shouldn't need to hit the database after the first time. As apps get larger, it becomes a lot more difficult to keep queries all in a single spot. With this, as soon as you make a query, we store that using the SQL as the key, and the ResultSet we get from the database as the value. Then anytime you make that same query, we can instantly return the ResultSet and re-hydrate the object using `SomeModel.from_rs(result_from_cache)`.

Ultimately, if this all gets figured out and sorted, there will be another PR added in to Lucky that would set this up on a per request basis. I don't have any plans for making it pluggable with an alternate cache storage back-end. Maybe in the future, but for now, just in-memory will be a nice first step.

Query cache is not straightforward. I've gone through several different iterations on this, and so far this is the version to come closest. I'll comment on the PR to better explain what I'm doing here. For now, this will be a WIP just to get some more eyes on it.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -34,6 +34,9 @@ dependencies:
   pulsar:
     github: luckyframework/pulsar
     version: ~> 0.2.2
+  splay_tree_map:
+    github: wyhaines/splay_tree_map.cr
+    version: ~> 0.1.3
 
 development_dependencies:
   ameba:
--- a/shard.yml
+++ b/shard.yml
@@ -17,8 +17,9 @@ dependencies:
     github: luckyframework/lucky_task
     version: ~> 0.1.0
   pg:
-    github: will/crystal-pg
-    version: ~> 0.24.0
+    # TODO: https://github.com/will/crystal-pg/pull/236
+    github: jwoertink/crystal-pg
+    branch: register_uuid_array
   habitat:
     github: luckyframework/habitat
     version: ~> 0.4.7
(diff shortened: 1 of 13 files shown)


robacarp:
There's a ton of room in a database layer for query caching. When it works, you only notice the speed, you don't notice the cache. When it doesn't work, you notice terrible behavior all day every day and you don't know why. That just means it's important to land on a good implementation.

My favorite opt-in caching mechanism is built like this:

```crystal
def retrieve_value
   cache :cache_key, expire: 1.minute do
     actually_retrieve_value
   end
end
```

I have a ton to praise about this, but primarily:

- The caching interface is generic, so it can be used to store anything rather than just database results. Generated images or the results of a third party api call, etc.
- When the time comes the underlying storage mechanism can be swapped out so that multiple app containers can share a warmed cache rather than all needing to warm independently. This reduces warming time by N where N is the number of containers running. This also means that a fresh spawn container has a hot cache rather than needing to warm up itself before performing well. The storage mechanism isn't implied by the DSL.
- It gives me the power to shoot myself in the foot, rather than some code I don't control shooting me in the foot. That means I (as the implementing developer) need to level up a bit to cache well, but it also means that when the cache breaks it's not the fault of the Framework team.

jwoertink:
closing in favor of #763

PR #759, opened 2021-10-31
Rename Status enums in operations.

### LINKED ISSUE #758: Naming your model Status causes named collisions
https://github.com/luckyframework/avram/blob/76e34c0e46d984c037eae9c8607514dc0b51eeec/src/avram/save_operation.cr#L26

https://github.com/luckyframework/avram/blob/76e34c0e46d984c037eae9c8607514dc0b51eeec/src/avram/delete_operation.cr#L18

If you have a model named `Status`, we will generate `SaveStatus` and `DeleteStatus` operations, but since those also exist as enums, Crystal will throw some errors.

We should just rename these enums since they're used internally anyway. Maybe `SaveOperationStatus` and `DeleteOperationStatus`... I think there's a less likely chance you'll name a model "OperationStatus"... or maybe we just call it `OperationStatus` on both? 🤔

jwoertink:
Fixes #758

This helps to avoid name collisions if you happen to create a model named `Status`

### DIFF
--- a/spec/operations/delete_operation_spec.cr
+++ b/spec/operations/delete_operation_spec.cr
@@ -30,7 +30,7 @@ describe "Avram::DeleteOperation" do
 
       BasicDeleteUser.delete(user) do |operation, deleted_user|
         operation.valid?.should be_true
-        operation.delete_status.should eq BasicDeleteUser::DeleteStatus::Deleted
+        operation.delete_status.should eq BasicDeleteUser::OperationStatus::Deleted
         deleted_user.not_nil!.name.should eq user.name
         UserQuery.new.select_count.should eq 0
       end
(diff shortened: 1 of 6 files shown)


matthewmcgarvey (approved):

robacarp:
I think this just makes the collision less likely? If I happen to make an enum called OperationStatus in my model, think it will still conflict.

jwoertink:
No, I don't think it would. The reason they collided before was because we generated operations called `SaveStatus` or `DeleteStatus`. It didn't really collide with the model itself, but just the generated operations.

Now, if you created an operation called `OperationStatus`, or tried to create some constant within the scope of an operation, it would... That's probably a super custom edge case though. I'll do a quick test just to be sure, but I think we'll be ok 😬

Issue #755, opened 2021-10-25
Add random_order query method


jwoertink:
If you want your results to come back in a random order, you can currently do

```
UserQuery.new.order_by("RANDOM()", :asc)
```

But it would be neat if it was just built in with `UserQuery.new.random_order`.

wout:
I'll take care of this!

jwoertink:
Sweet!

jwoertink:
Closed by https://github.com/luckyframework/avram/pull/765

PR #765, opened 2021-11-19
Add random_order method to query objects

wout:
Here's a proposal for the `random_order` implementation (https://github.com/luckyframework/avram/issues/755).

Initially, I tried to add a `:random` option to `Avram::OrderBy` but a few things were bad about that approach:
- too many if statements to switch between _order by random_ and _order by column_ clauses
- unnecessary attributes for the random option
- unused `column` argument in the initializer, requiring the `column` attribute to be nilable
- clunky internal API (e.g. `Avram::OrderBy.new(nil, :random)`)

That's why I decided to create an abstract class called `Avram::OrderByClause` and split out the two different order methods into two subclasses called `Avram::OrderBy` and `Avram::OrderByRandom`. The `Avram::OrderBy` class could have been renamed to `Avram::OrderByColumn` for consistency, but by keeping the old name, we're not breaking anything.

Having multiple _order by_ classes also will make it easier to add more order methods later on. Like ordering by an array for example, which is something I've needed in the past. For example, we could also add an `Avram::OrderByArray` class, generating the following clause:

```sql
ORDER BY array_position('{5,1,4,3,2}'::int[], item.id::int)
```

Which could look like:

```crystal
Avram::OrderByArray.new(:id, [5,1,4,3,2], :nulls_last)
```

### DIFF
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -8,7 +8,7 @@ class Avram::QueryBuilder
   @offset : Int32?
   @wheres = [] of Avram::Where::Condition
   @joins = [] of Avram::Join::SqlClause
-  @orders = [] of Avram::OrderBy
+  @orders = [] of Avram::OrderByClause
   @groups = [] of ColumnName
   @selections : String = "*"
   @prepared_statement_placeholder = 0
@@ -150,8 +150,15 @@ class Avram::QueryBuilder
     self
   end
 
-  def order_by(order : OrderBy)
+  def order_by(order : OrderByClause)
+    reset_order if ordered_randomly?
     @orders << order
+    self
+  end
+
+  def random_order
+    reset_order
+    @orders << Avram::OrderByRandom.new
     self
   end
 
@@ -165,7 +172,7 @@ class Avram::QueryBuilder
   end
 
   def reverse_order
-    @orders = @orders.map(&.reversed).reverse!
+    @orders = @orders.map(&.reversed).reverse! unless ordered_randomly?
     self
   end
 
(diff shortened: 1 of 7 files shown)


src/avram/query_builder.cr
```diff
@@ -176,7 +183,7 @@ class Avram::QueryBuilder
   end
 
   def orders
-    @orders.uniq!(&.column)
+    @orders.uniq!(&.uid)
```
wout, line 186: Since `Avram::OrderByRandom` doesn't have a `column` method, I had to find a way to select unique order clauses. I'm not 100% sure if `uid` is the best method name here. Suggestions are welcome. :smile:

jwoertink:
wow! This is a lot more code than I was thinking 😂  This is pretty great though. It does feel like this will allow things to be more extensible for the future

wout:
Yeah, I didn't expect to add this amount of code either. But you know how those things go...

src/avram/query_builder.cr
jwoertink: Does this value matter for the random order? Is it just to ensure we don't have 2 `ORDER BY RANDOM()`? If that's the case, can we just make a `column` method on the random that returns `"*"`? I was thinking since you'll see "*" as meaning "all" that maybe that's the "column".... Not sure if that would work the same or not

wout:
You're right, reusing the column attribute is much more elegant.

jwoertink (approved):
Looks good!

wout:
Shall I update the docs as well?

jwoertink:
We should at least open an issue to update docs. It won't go out until next release

PR #767, opened 2021-11-30
Release 0.22.0

jwoertink:


src/avram/migrator/runner.cr
```diff
@@ -201,7 +201,7 @@ class Avram::Migrator::Runner
   end
 
   private def sorted_migrations
-    self.class.migrations.sort_by(&.new.version)
+    self.class.migrations.sort_by(&.new.version.as(Int64))
```
jwoertink, line 204: LuckCli keeps failing here:

```
In lib/avram/src/avram/migrator/runner.cr:204:27

 204 | self.class.migrations.sort_by(&.new.version)
                             ^------
Error: Error: can't infer block return type, try to cast the block body with `as`.
```

https://github.com/luckyframework/lucky_cli/runs/4374304374?check_suite_focus=true#step:9:511

PR #770, opened 2021-12-10
Adding latest and nightly checks for CI

jwoertink:
Ensuring that nothing breaks with latest Lucky or Crystal


jwoertink:
ummm... what? 🤨  that's weird they all failed like that.

matthewmcgarvey:
That's also way too many spec runs

PR #772, opened 2021-12-24
Add optional lucky extensions for avram

matthewmcgarvey:
This is where the code that combines Lucky and Avram will live that was taken from https://github.com/luckyframework/lucky/pull/1620.

To get the same behavior in apps as there was before, requires will be changed.

```crystal
require "lucky"
```

becomes

```crystal
require "lucky"
require "avram"
require "avram/lucky"
```

That should be the only breaking change. The tasks should continue to work, but the ones removed from lucky will no longer be preloaded (`lucky gen.model` and `lucky gen.resource.browser`)

I organized the code in a way that Avram can be run without the lucky dependency if desired. For instance, lucky is not included as a dependency of Avram (it is a development dependency, though). So, in order to use `require "avram/lucky"`, you must add lucky as a dependency or use in a lucky app.


jwoertink (approved):
I wonder if for the precompiled tasks, we just remove all of them. Then we somehow create a single command that builds all of them at development time. This would allow you to precompile some additional tasks like `db.console` since we'd know your final config at that time :thinking:  

In any case, I'll merge this now to get started on cleaning things up. I think this is going to help solve the param issue too, but that'll come a bit later.

PR #778, opened 2021-12-31
Add Queryable#group_count functionality

grepsedawk:


src/avram/query_builder.cr
```diff
@@ -254,6 +254,12 @@ class Avram::QueryBuilder
       .map(&.split('.').last)
   end
 
+  # TODO: rename/refactor
+  def select_direct(selection : Array(ColumnName))
```
grepsedawk, line 258: Could use some help with naming/using existing methods to do this

The reason select(Array(ColumnName)) [right below] doesn't work is because it re-prefixes the @table and our groups from previous method have the table already included in the selection, as well as COUNT(*) needing no table prefix.

src/avram/queryable.cr
```diff
@@ -243,6 +243,22 @@ module Avram::Queryable(T)
...
   end
 
+  def group_count
+    database.query_all(
+      query.select_direct(query.groups + ["COUNT(*)"]).statement,
+      args: query.args,
+      queryable: schema_class.name,
+    ) do |rs|
+      k = query.groups.map do
+        rs.read PG::PGValue | Int64 # TODO: Why isn't Int64 part of PGValue?
```
grepsedawk, line 253: @jwoertink @robcole and I were talking about this, not sure why it's not a part of PG::PGValue. Maybe @will knows?

jwoertink: I think it may just be an oversight. I'd open up a PR adding that, plus the other supported Ints.

src/avram/queryable.cr
```diff
@@ -243,6 +243,22 @@ module Avram::Queryable(T)
     end
   end
 
+  def group_count
+    database.query_all(
+      query.select_direct(query.groups + ["COUNT(*)"]).statement,
+      args: query.args,
+      queryable: schema_class.name,
+    ) do |rs|
+      k = query.groups.map do
```
jwoertink, line 252: Nice! 👍

src/avram/queryable.cr
```diff
@@ -243,6 +243,22 @@ module Avram::Queryable(T)
     end
   end
 
+  def group_count
+    database.query_all(
```
jwoertink, line 247: We will want to make sure we raise an error here if you try to call it when there's no groups. Would be even better if we could make it a compile-time error  somehow...

src/avram/query_builder.cr
jwoertink: I think `select_direct` sounds fine. `unscoped_select` could be another one... Chances are this will be more of an escape hatch than a direct API usage.

src/avram/queryable.cr
```diff
@@ -243,6 +243,22 @@ module Avram::Queryable(T)
     end
   end
 
+  def group_count
```
jwoertink, line 246: If that issue PGValue gets updated with the Ints, we can type this out to be `Hash(Array(PG::PGValue), Int64)`.

jwoertink (commented):
This is definitely a great start. We can get some more eyes on this, and then some specs and stuff, but good work!

src/avram/queryable.cr
robcole: It still works if there aren't groups. It will return a count of all records with an empty array as the grouping key.

grepsedawk: Yep! It becomes 1 group of `[]`, which is canonically correct!

src/avram/queryable.cr
grepsedawk:
```suggestion
  def group_count : Hash(Array(PG::PGValue | Int64), Int64)
```

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,95 @@ describe Avram::Queryable do
...
+        query = UserQuery.new.group(&.name)
+
+        query.group_count.should eq({ [] of String => 0 })
+      end
+
+      it "2 grouped columns returns { [] => 0 }" do # TODO: What should we do here?
+        query = UserQuery.new.group(&.name)
+
+        query.group_count.should eq({ [] of String => 0 })
+      end
```
grepsedawk, line 877: These 2 fail because the query returns an entirely empty `Hash`

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,95 @@ describe Avram::Queryable do
...
+
+      it "1 grouped columns (age, name) returns { [42, \"Gumby\"] => 1 }" do
+
+        query = UserQuery.new.group(&.age).group(&.name)
+
+        query.group_count.should eq({ [32, "Taylor"] => 1 })
+      end
+    end
+
+    context "when matrix [32, Daniel] [32, Taylor] [32, Taylor] [44, Shakira]" do
```
grepsedawk, line 908: All of these tests might be overkill or maybe they're not enough, I'm not really sure.
I tried to express the various returns that could happen with `zero | one | many` groups

src/avram/query_builder.cr
```diff
@@ -254,6 +254,12 @@ class Avram::QueryBuilder
       .map(&.split('.').last)
   end
 
+  # TODO: rename/refactor
```
grepsedawk, line 257:
```suggestion
```

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,95 @@ describe Avram::Queryable do
...
+        end
+      end
+
+      it "0 grouped columns returns { [] => 1 }" do
+        query = UserQuery.new
+
+        query.group_count.should eq({ [] of String => 1 })
+      end
+
+      it "1 grouped columns (age) returns { [42] => 1 }" do
```
matthewmcgarvey, line 894: Could you update these spec/context strings that are really specific to be more generic? This one is incorrect, for example

matthewmcgarvey (changes requested):
Nice! Looks like you made something pretty great!

spec/avram/queryable_spec.cr
grepsedawk: image

spec/avram/queryable_spec.cr
grepsedawk: Yea I was thinking about that! Will do

matthewmcgarvey (approved):
I'm happy with this 🎉

grepsedawk:
@matthewmcgarvey Still a draft but thanks for your support!
Where would I go about adding docs? (or is that seperately, on the lucky website repo?)

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,97 @@ describe Avram::Queryable do
     end
   end
 
+  describe "#group_count" do
+    context "when no records exist" do
+      it "0 grouped columns returns { [] => 0 }" do
+        query = UserQuery.new
+
+        query.group_count.should eq({ [] of String => 0 })
```
jwoertink, line 864: I'm not so sure about this... To me, this seems like a mistake. You're saying "Give me a grouped count on a non-grouped query". I get that canonically it's correct, but there's a few things we do in Lucky where we say "even though it's semantically correct, it's still more likely to cause issues". For example with empty strings. We treat those the same as `nil` even though they're not.

@matthewmcgarvey what do you think about this raising an error instead? If your query isn't grouped, should you still be allowed to call this method?

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,97 @@ describe Avram::Queryable do
...
+        UserFactory.create do |user|
+          user.age(32)
+          user.name("Taylor")
+        end
+      end
+
+      it "0 grouped columns returns { [] => 1 }" do
+        query = UserQuery.new
+
+        query.group_count.should eq({ [] of String => 1 })
```
jwoertink, line 893: Yeah, same as I mentioned above. This looks even more like an error here since you actually get a value back. It reads to me like "You have one of nothing".... If you were displaying this on a page in some table, your label would end up being blank.

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,97 @@ describe Avram::Queryable do
...
+        query = UserQuery.new.group &.age
+
+        query.group_count.should eq({ [32] => 1 })
+      end
+
+      it "1 grouped columns (age, name) returns grouping" do
+
+        query = UserQuery.new.group(&.age).group(&.name)
+
+        query.group_count.should eq({ [32, "Taylor"] => 1 })
```
jwoertink, line 906: I love this!

src/avram/queryable.cr
```diff
@@ -243,6 +243,22 @@ module Avram::Queryable(T)
     end
   end
 
+  # Remove when PG::PGValue matches rs.read possibilities
+  alias PGValue = Bool | Float32 | Float64 | Int32 | Int64 | PG::Numeric | String | Time | UUID | Nil
```
jwoertink, line 247: 👍  good call. Can we add `Int16` too? That's the only other simple type that Avram supports missing from here.

src/avram/queryable.cr
```diff
@@ -243,6 +243,22 @@ module Avram::Queryable(T)
...
+  def group_count : Hash(Array(PGValue), Int64)
+    database.query_all(
+      query.select_direct(query.groups + ["COUNT(*)"]).statement,
+      args: query.args,
+      queryable: schema_class.name,
+    ) do |rs|
+      {
+        query.groups.map { rs.read PGValue },
+        rs.read Int64,
+      }
```
jwoertink, line 258: oh sweet! Looks like you were able to clean this up quite a bit.

jwoertink (commented):
Overall I think this is great. Really great work on this.

My only concern is the issue when there's no grouping. It seems that when there's no group, you get an empty array. Then `([] + ["COUNT(*)"]).statement` would just end up being the same as the `select_count` method. If you feel this is a case that may be wanted, we could have an escape hatch method `group_count?` that would just give the value. Then `group_count` would do the raise. I say `group_count` for raising instead of `group_count!` because I would want it to be the default use. What are your thoughts?

spec/avram/queryable_spec.cr
matthewmcgarvey: I'm fine with that. Just checked and Rails does not allow you to call group without a column https://github.com/rails/rails/blob/9a0d951f3849ac84a2a9baa9e8080ede4c912183/activerecord/lib/active_record/relation/query_methods.rb#L352-L355

src/avram/query_builder.cr
robacarp: `select_direct` is opaque to me. This returns a list of columns, can it just be `selected_columns`?

src/avram/queryable.cr
robacarp: Yes, this method looks great

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,97 @@ describe Avram::Queryable do
...
+        query.group_count.should eq({ [] of String => 1 })
+      end
+
+      it "1 grouped columns (age) returns grouping" do
+        query = UserQuery.new.group &.age
+
+        query.group_count.should eq({ [32] => 1 })
+      end
+
+      it "1 grouped columns (age, name) returns grouping" do
```
robacarp, line 902: Should this be:

```suggestion
      it "2 grouped columns (age, name) returns grouping" do
```

spec/avram/queryable_spec.cr
```diff
@@ -856,6 +856,97 @@ describe Avram::Queryable do
...
+        end
+      end
+
+      it "0 grouped columns returns { [] => 1 }" do
+        query = UserQuery.new
+
+        query.group_count.should eq({ [] of String => 1 })
+      end
+
+      it "1 grouped columns (age) returns grouping" do
```
robacarp, line 896:
```suggestion
      it "1 grouped column (age) returns grouping" do
```

src/avram/query_builder.cr
jwoertink: Maybe we need to see why using the normal `select` method prepends the table twice.... I think that was the reason for this second method anyway, right?

`query.select(query.groups)` was causing something like `GROUP BY table.table.column` or something 🤔 so we had to come up with a different method that didn't pre-scope the columns. I think this method is more of a hack, and really shouldn't be exposed externally. I just forget why it was needed.

spec/avram/queryable_spec.cr
robacarp: I think this should mirror what postgres does as well. I think the intent is to make the code receiving a grouped select less complex, right? That being said, I don't think I know what it means to have 1 of nothing.

spec/avram/queryable_spec.cr
grepsedawk: Rails doesn't allow you to call a group with no columns, and neither do we -- this is only allowing you to return the same type without requiring a group

src/avram/query_builder.cr
grepsedawk: https://github.com/luckyframework/avram/pull/778/files#diff-df659060f00cb38deae2b7b48c5172fa24e3fd5b7f3652d35f96f992b35b0116R263
select itself prepends the table name

src/avram/queryable.cr
grepsedawk:
```suggestion
  alias PGValue = Bool | Float32 | Float64 | Int16 | Int32 | Int64 | PG::Numeric | String | Time | UUID | Nil
```

spec/avram/queryable_spec.cr
grepsedawk: @robacarp see https://github.com/luckyframework/avram/pull/778/files#r780806822 -- I actually do believe this mirrors PG's handling of this

src/avram/query_builder.cr
grepsedawk: One option might be to have a `prepend : Boolean = true` default arg on the select function itself

spec/avram/queryable_spec.cr
jwoertink: I still think we need to make it an error, but I might have a way to make it a compile-time error... We can roll with what you have now, and I'll do a PR a bit later to try it out. Thanks for the explanation, and all the work on this!

Basically I want to try setting a nilable bit once you've added a group, then do a check on the args like [rails does](https://github.com/rails/rails/blob/9a0d951f3849ac84a2a9baa9e8080ede4c912183/activerecord/lib/active_record/relation/query_methods.rb#L1677-L1678) but it'll check it at compile-time forcing you to call `group()` at least once before calling this method. 😄

jwoertink (approved):

matthewmcgarvey:
@grepsedawk just need to fix the formatting issue

spec/avram/queryable_spec.cr
grepsedawk: If it's compile time I feel that's a lot more tolerable than what I initially had which I believe is a runtime exception

jwoertink:
🎉

PR #780, opened 2022-01-01
Run specs in a transaction instead of truncating db before every spec

matthewmcgarvey:
Before making this re-usable, I'd like my work to be verified. This changes how avram works and monkey patches crystal-db so we'll need to be careful.

I've been referencing Rails' implementation. Want to see what I'm looking at? Here are some of the files of interest:
- https://github.com/rails/rails/blob/main/activerecord/lib/active_record/test_fixtures.rb
-- This is where the code to run tests within transactions is. It's roughly equivalent to what I've written in the `spec_helper.rb`
-- `setup_fixtures` and `teardown_fixtures` are the two main methods
- https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
-- Where I was referencing for a lot of the code around managing transactions


spec/avram/instrumentation_spec.cr
```diff
@@ -2,7 +2,8 @@ require "../spec_helper"
 
 describe "Instrumentation" do
   it "publishes the query and args" do
-    TestDatabase.query "SELECT * FROM users"
+    # using block form to make sure the ResultSet is closed
+    TestDatabase.query "SELECT * FROM users" { }
```
matthewmcgarvey, line 6: This caused an error when rolling back the transaction. Surprised this didn't come up before.

spec/avram/model_spec.cr
```diff
@@ -33,7 +33,6 @@ describe Avram::Model do
     user = UserFactory.create
     post = PostFactory.create
 
-    user.id.should eq(post.id)
```
matthewmcgarvey, line 36: The id incrementing does not get reset with transaction rollbacks so there's no way to do this check. This spec seems a bit pointless but I did the minimal change.

spec/avram/query_logging_spec.cr
```diff
@@ -13,7 +13,7 @@ describe "Query logging" do
   it "does not log truncate statements" do
     Avram::QueryLog.dexter.temp_config do |log_io|
       TestDatabase.truncate
-      log_io.to_s.should eq("")
+      log_io.to_s.should_not contain("TRUNCATE TABLE")
```
matthewmcgarvey, line 16: Since we're not truncating the db before the spec, the DatabaseInfo is not loaded/memoized and it will be fetched on this truncate call and show up in the logs https://github.com/luckyframework/avram/blob/master/src/avram/database/database_info.cr

src/avram/database.cr
```diff
@@ -3,7 +3,8 @@ abstract class Avram::Database
 
   @@db : DB::Database? = nil
   @@lock = Mutex.new
-  class_getter transactions = {} of FiberId => DB::Transaction
+  class_getter connections = {} of FiberId => DB::Connection
```
matthewmcgarvey, line 6: We maintain the connections reference instead of transactions because new sub-transactions can be made which would have overwrote the parent transaction. By holding onto the connection (and the added monkeypatch) we can  still get to the transaction

src/avram/database.cr
```diff
@@ -154,16 +155,24 @@ abstract class Avram::Database
...
+    connections[current_object_id] ||= db.checkout
+  end
+
   private def current_transaction : DB::Transaction?
-    transactions[Fiber.current.object_id]?
+    current_connection.stack.last?
+  end
+
+  def current_object_id : FiberId
+    @@lock_id || Fiber.current.object_id
```
matthewmcgarvey, line 175: I'm not sure this was required for this specific work so far. It will be more useful in the future when we're testing an app that will be running on a different fiber. We can control which connections are used by doing this

src/avram/database.cr
```diff
@@ -180,7 +189,7 @@ abstract class Avram::Database
 
   # :nodoc:
   def transaction : Bool
-    if current_transaction
+    if current_transaction.try(&.joinable?)
```
matthewmcgarvey, line 192: transaction is "joinable" if it does not require a sub-transaction to be made. We have always considered nested transactions to be "joinable" though it wasn't encoded. The new thing is specifying that a transaction is not joinable and a sub-transaction should be made.

This was added so that specs verifying failed operations and other transaction rollback scenarios would still work. The "root" transaction that is added in the spec helper is not joinable so the transaction made in an operation is forced to be sub

src/avram/database.cr
```diff
@@ -190,20 +199,19 @@ abstract class Avram::Database
...
-  private def transactions
-    self.class.transactions
+  private def connections
+    self.class.connections
   end
 
   private def wrap_in_transaction
-    db.transaction do |tx|
-      transactions[Fiber.current.object_id] ||= tx
+    (current_transaction || current_connection).transaction do
```
matthewmcgarvey, line 207: Use the current transaction to make a sub transaction if one is available to use before using the connection

src/avram/database.cr
```diff
@@ -190,20 +199,23 @@ abstract class Avram::Database
...
     true
   rescue e : Avram::Rollback
     false
   ensure
-    transactions.delete(Fiber.current.object_id)
+    # TODO: not sure of this
+    if current_connection.stack.empty?
+      current_connection.release
+      connections.delete(current_object_id)
+    end
```
matthewmcgarvey, line 218: I was referencing Rails' implementation for all of this. They had the tie between the connection and the thread id in the database pool which made this a bit confusing. Not sure if this is correct.

src/ext/db/connection.cr
```diff
@@ -0,0 +1,11 @@
+module DB
+  abstract class Connection
+    getter stack = [] of DB::Transaction
+
+    def begin_transaction : Transaction
+      return stack.last if @transaction
+      @transaction = true
+      create_transaction
+    end
```
matthewmcgarvey, line 9: https://github.com/crystal-lang/crystal-db/blob/5a7d27e0c58510458af3894783649d53529753cc/src/db/connection.cr#L49-L53

This needs to be thought about quite a bit. The regular implementation raises an error if it's called while there's already a transaction. Instead of doing that, I return the current transaction.

Rails' implementation returns a new nested transaction if one exists https://github.com/rails/rails/blob/b7e1ea2eeb524224dc06c08fbb779f99d20dde2f/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L233-L263

src/ext/db/pool.cr
```diff
@@ -0,0 +1,28 @@
...
+      @total
+    end
+
+    private def build_resource : T
+      resource = @factory.call
+      @total << resource
+      @idle << resource
+      if resource.is_a?(DB::Connection)
+        ConnectionStartedEvent.publish(connection: resource)
+      end
```
matthewmcgarvey, line 13: Just added the pubsub to this so that we could go ahead and start/track a transaaction on any new connections.

spec/avram/instrumentation_spec.cr
jwoertink: Wouldn't this use https://github.com/crystal-lang/crystal-db/blob/5a7d27e0c58510458af3894783649d53529753cc/src/db/query_methods.cr#L45 without the block? That method is defined https://github.com/luckyframework/avram/blob/9b184d4b586d783b096a5d377eafd98c78e32026/src/avram/database.cr#L77

spec/avram/model_spec.cr
jwoertink: 🤔  lol... is this just assuming that `user.id` and `post.id` will both be `1` here?  Yeah, that doesn't seem very helpful.. did I do that? 😂

spec/avram/instrumentation_spec.cr
matthewmcgarvey: Yes, but check out the comment above the method. It returns a `ResultSet` that must be closed (which is why I'm surprised it didn't cause issues before this). Because we weren't closing the result set, an error occurred when rolling back the transaction because there was data still in the pipe left to parse. None of that is official lingo, just how I understand it.

src/avram/database.cr
jwoertink: When would the stack on the current_connection be empty? Like, what would cause that case?

spec/avram/instrumentation_spec.cr
jwoertink: oh, I see. Sort of like doing `File.open` without the block... Gotcha.

src/avram/database.cr
matthewmcgarvey: Outside of specs where a root transaction is created, once an operation is complete, the transaction is done and the stack will be empty so we release the connection back to the thread pool. This is where things are pretty confusing to me at the moment

jwoertink (commented):
So far it seems ok. I may need to come back and really read through the extension stuff. That's the only part that has me a bit foggy. But great start!

bcardiff:
I think I have a different implementation in mind. Would it be enough/simpler?

The `TestDatabase` setups a `DB::Database`, right? That `Database` can be configured to start always a transaction using `Database#setup_connection`.

We could track connection->transaction in `TestDatabase` (probably using a `Hash(WeakRef(Connection), Transaction))`)

I'm not sure how you determine which connection was used on a test, or we assume any connection could have been used. Either way, all connections that could have been used can be reset in the tear down, and a new transaction could be created for the next test. And the connection->transaction hash updated.

How does that sound?

--

I would like that adding events/subscribers on DB to have their own discussion. Same as the idea of adding joinable or other properties to transaction. It seems a lot to be added in one shoot IMO and I am not sure the best design and implementation.

matthewmcgarvey:
@bcardiff Glad you're taking a look!

Really like your suggestion of using `Database#setup_connection`! I updated the code to use that instead of my thrown together solution using Pulsar.

As I think more about your hash suggestion, I think it's starting to make some sense, but I'll have to spend some more time on it.

I'm not sure there is a way to avoid monkey patching in the `joinable?` piece, though.

matthewmcgarvey:
One thing to note... migrations now have a ton of logs since it's using `Avram::Database` instead of `DB::Database` directly now.

src/avram/database.cr
bcardiff: How is this hash cleared? I think it might prevent Connection objects from being released. WeakRef can help in that regard if there is no good place to clear the hash.

bcardiff:
> I'm not sure there is a way to avoid monkey patching in the joinable? piece, though.

I see. You need `DB::Connection#begin_transaction` to be able to create a nested transaction and not raise if the connection is already on a transaction mode. So other users of the connection can do that without needing the transaction explicitly.

I will check some code around and see if the existing restriction is because it was simpler and smaller that way and if so if it can be lifted.

---

On other aspects I don't know a lot about Avram and the lifecycle of connections, but... why they are mapped to FiberId and how you identify the connections used during a spec?

matthewmcgarvey:
@bcardiff That's not quite it. By default, Avram didn't use `SavePointTransaction`s. If there was a transaction open, it didn't try to open a new one. But with this change, The transactions opened in the spec helper are getting set to `joinable = false` so that Avram still behaves as it did before and opens new transactions. These just so happen to be `SavePointTransaction`s instead of the root ones. Without that, if the spec was testing what happens if the change to the database fails, it would rollback the root transaction and the spec would leak data if it continued.

matthewmcgarvey:
At this point, I'm somewhat confident this is merge-able. There's nothing crazy going on with the monkey-patches and I've been able to test it in a Lucky app.

bcardiff:
Thanks for the clarification @matthewmcgarvey . I think joinable is True iff the transaction is in the trackable_transaction list, right? If this is true the joinable patch could even be removed. When running the code outside of spec the list/set would be empty.

If this is not preferred, I would like to encourage prefixing joinable with avram_ and maybe put a `:nodoc`.

src/avram/database.cr
jwoertink: I don't know if the connections hash was ever being cleared. I think the individual connections inside were just being deleted somewhere... 🤔  I'd like to learn more about this `WeakRef`. I'm still not really sure what it does, or how to use it

src/avram/test.cr
```diff
@@ -0,0 +1,45 @@
+module Avram::Test
```
jwoertink, line 1: Is this the official name, or just for testing purposes?

jwoertink:
Does this have any affect on multiple databases? I know we've never really supported migrations with multiple dbs, so that's fine.... but just the connection stuff, that should all still be good, right? I'd really like to hear more about what @bcardiff is suggesting with the `WeakRef`, and maybe even removing the patch if possible.

src/avram/test.cr
matthewmcgarvey: I didn't put much consideration into this name. What do you think would be better?

matthewmcgarvey:
My understanding of what maintaining a hash like `Hash(WeakRef(Connection), Transaction)` would entail would be that before we create a transaction, we get the current transaction and put it in a temp variable and then use that to make a transaction and set it as the current value to the connection key. When we leave the transaction block, we put back the transaction in the temp variable as the current.

I am able to get working code if I make a transactions hash like `Hash(Connection, Transaction)`. I'm not sure what the intended way of holding on to a reference of `WeakRef(Connection)` would work. Either way, I don't find the code easy to understand this way. The monkey-patched stack makes more sense to me, personally.

matthewmcgarvey:
@jwoertink I think I've updated it to work with multiple databases.

src/avram/test.cr
jwoertink: hmm... `Avram::SpecTransaction`... `Avram::TransactionSpecHelper` ? 🤔  I just thought `Test` was you trying things out and you were gonna rip this out when you were done. I wasn't sure if it was actual code or not 😅

matthewmcgarvey:
Making a note of this:

So far I know of 3 ways your tests could break with this used in your specs:
- You use `Avram::Database.listen` (it will throw an error if used in a transaction)
- You use `DB::Database` directly since it won't know about the existing transaction
- You have database interactions outside of `it` blocks. Like accidentally writing specs in a `describe`

bcardiff:
Using prefixed properties will be good in case stack lands in crystal-db. I am less confident of joinable getting into crystal-db, I need to understand it better outside Avram.

Regarding the `WeakRef`, the concern is about memory footprint. So in the context of specs it might not be a big deal. Anyway, if during the execution of the specs the pool creates multiple connections and those connections are referenced somewhere, then they will not be GC-ed. The connection socket should be closed though. 

When using a `WeakRef`, that is almost a boxed refrence, you are telling the GC to not count the inner reference. Allowing the Connection object to be GC-ed. The cost is that when dereferencing a `WeakRef`, the inner value could be a `nil`, because the object was GC-ed.

Another approach would be to have events on the DB and have a more strict lifecycle of the foreign references to connection. But these does not exist.

PR #785, opened 2022-01-11
Updates ci.yml to test crystal 1.3

robacarp:
Also provides implicit "latest patch" crystal versions so that any patches to releases are picked up. Given the change in luckyframework/lucky, I'm not sure if the inbetweenst versions should be kept or just 1.0 and latest.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,9 +12,10 @@ jobs:
       fail-fast: false
       matrix:
         crystal_version:
-          - 1.0.0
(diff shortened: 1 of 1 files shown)


jwoertink:
I actually have this already https://github.com/luckyframework/avram/pull/770 😬  But I do sort of like the implicit... Though, I guess it doesn't matter much when saying `latest` 🤔

robacarp:
I thought for sure I'd seen it somewhere!

PR #786, opened 2022-01-11
Integrating AvramSlugify

jwoertink:
Closes https://github.com/luckyframework/avram_slugify/issues/13

This pulls the entire `AvramSlugify` shard directly in to Avram as `Avram::Slugify`. No usage or features have changed aside from the name scope. The main considerations for bringing this over are

* 1 less shard to maintain. Each Avram release required making sure AvramSlugify was also in-sync
* It's a super small shard with only 1 file (~ 45 LOC)
* New features become easier to implement with Avram being readily available

The specs weren't copy/paste since the shard was using a `User` model that wasn't setup the same. I decided to just make a separate model for testing that.

The only other main thing to consider was whether or not to make this an optional require. On the one hand, we could add it as its own module where you'd have to manually require it:

```crystal
require "avram"
# explicit require
require "avram/slugify"
```

But then I thought... why? It doesn't interfere with your code by always having it. Crystal is pretty good at optimizing for code that's not being used.

### DIFF
--- a/db/migrations/20220111192510_create_articles.cr
+++ b/db/migrations/20220111192510_create_articles.cr
@@ -1,0 +1,15 @@
+class CreateArticles::V20220111192510 < Avram::Migrator::Migration::V1
+  def migrate
+    create table_for(Article) do
+      primary_key id : Int64
+      add_timestamps
+      add title : String
+      add sub_heading : String?
+      add slug : String
+    end
+  end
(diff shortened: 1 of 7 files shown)


matthewmcgarvey (approved):

paulcsmith:
Good idea! 👍

PR #789, opened 2022-01-12
Add bulk upsert to lucky

robcole:
This is some WIP that @grepsedawk and I have been working on to bring in bulk upserts. Sharing this primarily to get some early feedback - there are a couple of things we've discovered that would be helpful to discuss:

1 - Postgre has a cap of 32768 bind variables that can be used. This would limit bulk inserts to a smallish number of updates, and it would probably be preferable to interpolate SQL values in the statement directly. Any ideas for things we'd need to look out for here?

2. This type of UPSERT requires unique constraints, which the test database doesn't currently have. The current `upsert` behavior looks like it's an upsert, but it's really more of an analog to `find_or_create_by` in Rails than a DB-level upsert. We'd love to eventually overhaul bulk upsert to work as a `.upsert` overload, and ideally move `.upsert` to something like `.create_or_find_by` etc if everyone is amenable. 

Any comments would be welcome - will begin adding some more tests, etc., later this week!


src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,70 @@
...
+    record.insert_values.map_with_index(1) do |_value, index|
+      "$#{index}"
+    end.join(", ")
+  end
+
+  private def value_placeholders
+    i = 0
+
+    @records.map do |record|
+      "($#{i += 1}, $#{i += 1}, $#{i += 1}, $#{i += 1})"
```
robcole, line 67: 🤷

grepsedawk: We were testing with 4 columns, this will need to be dynamic based off of columns.count

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,70 @@
...
+
+  private def returning
+    T.column_names.join(", ")
+  end
+
+  private def fields
+    @records.first.insert_values.keys.map do |key|
+      <<-TEXT
+      "#{key}"
+      TEXT
```
grepsedawk, line 47:
```suggestion
      <<-SQL
      "#{key}"
      SQL
```

spec/avram/bulk_upsert_spec.cr
```diff
@@ -0,0 +1,19 @@
+require "../spec_helper"
+
+describe Avram::BulkUpsert do
+  describe "bulk upserts" do
+    # Ideally, compiler can catch this / this should be impossible..
+    context "when collections mismatch" do
```
robcole, line 6: Won't be able to catch this with the compiler, so we probably want to raise some sort of RunTimeError when this happens. The fields for all records need to match, otherwise you can have mismatches between the number of fields and the number of values for a given record.

src/avram/bulk_upsert.cr
robcole: This was to test a theory about the right way to pass bind args -- will be removed, since bind args seem like a non-starter due to the 32k var cap.

spec/avram/bulk_upsert_spec.cr
grepsedawk: Ooh, we might be able to just pad out missing values with nil!

robcole: I think that would overwrite to NULL for those fields.

grepsedawk: In an update scenario, probably -- but we could make our CONFLICT UPDATE part use coalesce to retain old values when null is passed in a bull scenario!

grepsedawk: ON CONFLICT (XXX) DO UPDATE SET col = COALESCE(excluded.col, col)

src/avram/upsert.cr
```diff
@@ -81,6 +81,27 @@ module Avram::Upsert
...
     end
 
+    def self.bulk_upsert(upserts)
+      operations = upserts.map do |upsert_args|
+        new(**upsert_args)
+      end
+
+      upsert = Avram::BulkUpsert(self).new(
+        operations,
+        {{ attribute_names }}.to_a
```
robcole, line 91: We're going to want a similar API, but something that hooks into database column constraints, since this isn't used for DB-level upsert currently.

spec/avram/bulk_upsert_spec.cr
robcole: Ooooh that could work ❤️

robcole: Oh wait, that would prevent setting things to NULL, since it’d always coalesce back.

src/avram/upsert.cr
grepsedawk: I believe this can be either a list of columns or a constraint name

spec/avram/bulk_upsert_spec.cr
robacarp: Could you just use a named tuple as the input for each row?

src/avram/bulk_upsert.cr
robacarp: if the cap is 32k variables per query, could you do some basic math and split a statement into multiple queries? I would really like to avoid jumping into sql injection land.

src/avram/upsert.cr
```diff
@@ -92,7 +113,7 @@ module Avram::Upsert
 
   # :nodoc:
   macro included
-    {% for method in ["upsert", "upsert!"] %}
+    {% for method in ["upsert", "upsert!", "bulk_upsert"] %}
```
robacarp, line 116: Can / should this functionality just live in an overload to upsert rather than a dedicated method name?

src/avram/bulk_upsert.cr
robacarp: Something like:

```crystal
field_count = record.columns.count
records.each_slice(32767 / field_count) do |record_set|
   upsert(record_set)
end
```

I don't think I'd have a problem with raising an exception if you hit this maximum and just requiring the API to take a max of rows where 32767 > field_count * row_count.

src/avram/upsert.cr
```diff
@@ -81,6 +81,27 @@ module Avram::Upsert
...
+        {{ attribute_names }}.to_a
+      )
+
+      pp upsert.statement
+      pp upsert.args
+      records = [] of T
+
+      new.database.query upsert.statement, args: upsert.args do |rs|
+        records << T.from_rs(rs).first
+      end
```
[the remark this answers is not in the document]
grepsedawk: Interpolation doesn't have this limit and is actually much faster than this, but this might be safer/better

src/avram/upsert.cr
grepsedawk: From our DM's last night:

> [7:31 PM] grepsedawk: Also, it might be a hot take, but I really like the idea of doing method overloading over a new method here

Definitely a strong take, but we didn't change our path forward to open up discussion about this pattern.

I'm 💯 on board with an overload

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  private def conflicts
+    @column_names.join(", ")
+  end
+
+  private def set_timestamps(collection)
+    collection.map do |record|
+      record.created_at.value ||= Time.utc if record.responds_to?(:created_at)
+      record.updated_at.value = Time.utc if record.responds_to?(:updated_at)
```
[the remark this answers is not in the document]
grepsedawk: Yeah we were considering this! The one thing we needed to work through for this was a MAJOR data changeup -- these would then need to be NOW() instead of bound $n params which changes up the models a lot

spec/avram/bulk_upsert_spec.cr
grepsedawk: @robacarp If we used a named tuple, I think we run into a few other issues:

1. BulkUpsert class would need to be generated with macros I believe to support each class that uses it
2. You would ALWAYS need to pass ALL columns into bulk operations, which I'm not sure is something we'd want.

src/avram/bulk_upsert.cr
grepsedawk: robacarp example, for prosperity: https://github.com/luckyframework/avram/pull/789#discussion_r783094748

spec/avram/bulk_upsert_spec.cr
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    context "when mixed new records and updated records" do
+      # Insert spec example.
+      it "inserts with a hash of String" do
```
jwoertink, line 11: raw hashes are actually disallowed on purpose because they're not safe. They are easier to work with, but also impose security risk. This is why you can't pass a hash in to `create` or `update` https://github.com/luckyframework/avram/blob/ac281f490a85283eaf8b74fd23357dd70a6d7f22/src/avram/needy_initializer_and_save_methods.cr#L104

src/avram/bulk_upsert.cr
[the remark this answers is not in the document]
jwoertink: Related: https://github.com/luckyframework/avram/issues/393 (note the issue on `updated_at` here that would require a trigger to know when to change)
Somewhat related: https://github.com/luckyframework/lucky/discussions/1583

spec/avram/operations/save_operation_spec.cr
```diff
@@ -307,6 +307,24 @@ describe "Avram::SaveOperation" do
...
+    context "when the records are not persisted" do
+      it "should create the records" do
+        record_args = (1..2).to_a.map do |i|
+          {
+            name:     "Test #{i}",
+            nickname: "Test Nickname #{i}",
+          }
+        end
+
+        UpsertUserOperation.bulk_upsert(record_args)
```
jwoertink, line 320: Since this is just taking an Array of NamedTuple, there doesn't seem to be any safety here. For example, what if I add a key for a column that doesn't exist?

jwoertink (commented):
I LOVE that you guys are tackling this. I know it's a super difficult problem to solve, and to make it extra hard, we have to take additional considerations in to mind with the Lucky core goals. 

* Is this type-safe? (as in providing more safety than just what Crystal gives you)
* Does this help me catch bugs at compile-time?
* Does it add security by ensuring bad data can't be used?

I know the current `upsert` doesn't quite work like a normal DB upsert. @paulcsmith wanted to add `ON CONFLICT` support directly in to the query builder which would add some extra support. You can read up on the original PR https://github.com/luckyframework/avram/pull/334 

Also while you're working on this, keep in mind on how this would look coming from an HTML form. How do the params get built and passed over to this? One tip I've learned while working on Lucky is don't be afraid to break away from the normal "rails mindset" coding styles. Sometimes an API that looks a bit funky / offputting at first may end up being a better solution in the long run. A little extra typing characters could mean a few less specs and more confidence!

Feel free to hit me up if you have any questions I can answer. I'll keep an eye on the PR and try to chime in when I can. Stoke to see this! 🚀

spec/avram/bulk_upsert_spec.cr
robcole: Yeah this is placeholder code stolen from insert to remind me about some of the tests I want to add, need to update the tests f’sho.

src/avram/upsert.cr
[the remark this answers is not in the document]
robcole: Speed is my primary concern here, in a lot of models that would be a few thousand records per query max, which is TINY compared to the work I’ve done in other tooling with bulk upserts. I’ll keep mulling this over as I really want to keep sql injection away as an issue if possible.

spec/avram/bulk_upsert_spec.cr
robcole: Yeah the issue is mostly about the need for all keys we insert to be available in all records, otherwise some records will lose values.

grepsedawk:
> keep in mind on how this would look coming from an HTML form. How do the params get built and passed over to this?
@jwoertink 

anything that is manual (from a form) I'd generally consider "small enough" to not care about utilizing bulk operations directly from an html form

To me, bulk operations are more for:

- CSV imports
- API requests/syncs (in our production use case, syncing entire gmail inboxes from the gmail API)
- generative data (generate 10k records for a seed DB)

src/avram/upsert.cr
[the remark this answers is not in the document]
robcole: https://klotzandrew.com/blog/postgres-passing-65535-parameter-limit seems worth reviewing here.

spec/avram/operations/save_operation_spec.cr
jwoertink: oh sweet! Ok

src/avram/upsert.cr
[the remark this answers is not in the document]
grepsedawk: Assuming each row has 10 columns, that takes us to 6.5k records per bulk insert
I don't really even think that's worth it, especially when you consider the cost of binding 25k fields being 3 seconds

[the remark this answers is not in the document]
robacarp: @robcole that's a great link, thanks for sharing. That approach will not work for streams, but maybe thats okay.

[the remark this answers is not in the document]
grepsedawk: What is a stream?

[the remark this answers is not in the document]
robacarp: A stream of never-ending data. Examples:

- Real time event data
- Market data (stocks, crypto, etc)
- Sensor data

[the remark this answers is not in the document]
grepsedawk: I don't see why unnest wouldn't work for streams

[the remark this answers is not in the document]
robcole: Same. Can you explain more about this use case @robacarp? It seems like you could do the array building in something called from the stream, and then pass those records in to upsert at whatever interval or size makes most sense.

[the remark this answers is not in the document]
robacarp: Sure, if you chop up the stream every so often you can pivot your dataset and do the insert no problem.

grepsedawk:
Nice to haves:

- [ ] Throw a clean error when UK does not exist

src/avram/bulk_upsert.cr
robcole: Went with an approach that won't limit our record count.

src/avram/upsert.cr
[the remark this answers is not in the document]
robcole: No cap + no worries about SQL injection = 💪 🚀 🤖

spec/avram/bulk_upsert_spec.cr
```diff
@@ -0,0 +1,19 @@
+require "../spec_helper"
```
grepsedawk, line 1: Delete this file

spec/avram/operations/save_operation_spec.cr
```diff
@@ -67,6 +67,7 @@ end
 
 private class UpsertUserOperation < User::SaveOperation
   upsert_lookup_columns :name, :nickname
+  upsert_unique_on :name, :nickname
```
grepsedawk, line 70: 🤔 why both?

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,82 @@
...
+      "#{column}=EXCLUDED.#{column}"
+    end.join(", ")
+  end
+
+  private def returning
+    T.column_names.join(", ")
+  end
+
+  private def changed_attributes
+    @records.first.changed_attributes
```
grepsedawk, line 49: I don't know how arrays work in crystal, but `@records.first` is used _a lot_ -- might be worth memoizing

I suggest the name `sample`

src/avram/save_operation.cr
```diff
@@ -379,6 +383,14 @@ abstract class Avram::SaveOperation(T)
     @record.try &.id
   end
 
+  def self.column_names
+    T.column_names
+  end
+
+  def self.database_table_info
+    T.database_table_info
+  end
```
grepsedawk, line 392: Might want to delegate here:
https://crystal-lang.org/api/1.3.0/Object.html#delegate(*methods,toobject)-macro

It seems it might make stack traces more direct

src/avram/upsert.cr
```diff
@@ -90,9 +90,26 @@ module Avram::Upsert
     end
   end
 
+  macro upsert_unique_on(*attribute_names)
```
grepsedawk, line 93: https://github.com/luckyframework/avram/pull/789/files#diff-a265a41187dff25dd916dbc390a4da9c8a879e991c7fad41300fcd7fc6200607R117 references the wrong method in the error

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,82 @@
+class Avram::BulkUpsert(T)
+  def initialize(@records : Array(T), @column_names : Array(Symbol))
+    @records = set_timestamps(records)
+  end
+
+  def statement
+    <<-SQL
+    INSERT INTO #{table}(#{fields})
+      (select * from unnest(#{value_placeholders}))
```
grepsedawk, line 9:
```suggestion
      (SELECT * FROM unnest(#{value_placeholders}))
```

grepsedawk:
```suggestion
      (SELECT unnest(#{value_placeholders}))
```
I think this works too, could be wrong.

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,82 @@
...
+      ON CONFLICT (#{conflicts}) DO UPDATE SET #{updates}
+      RETURNING #{returning}
+    SQL
+  end
+
+  def args
+    @records.map do |record|
+      record.changed_attributes.map(&.value)
+    end.transpose
```
grepsedawk, line 18: If the records have a mismatching count of keys, transpose will fail.

spec/avram/operations/save_operation_spec.cr
robcole: Different APIs - one indicates a unique constraint, the other one is used for the find or create by and doesn’t have any requirement for mapping to UKs.

robcole: They just happen to be the same, at least until we decide on which database migration is best to add for any unique keys.

spec/avram/bulk_upsert_spec.cr
robcole: I think we may want a few small tests around this.

src/avram/bulk_upsert.cr
robcole: Yeah, we still need to add some sort of Way of preventing mismatched arrays, and handling the null overwrite issues in the failing spec

src/avram/bulk_upsert.cr
```diff
@@ -14,10 +17,27 @@ class Avram::BulkUpsert(T)
...
   def args
     @records.map do |record|
-      record.changed_attributes.map(&.value)
+      changed_record_values(record, &.value)
     end.transpose
   end
 
+  private def build_permitted_fields(fields_arr)
+    record = @records.first
+    fields_arr.push(:created_at) if record.responds_to?(:created_at)
```
grepsedawk, line 26: What if a user includes this column in the changeset?

robcole: 👍 - will add a de-dup as well, though I think this approach is going to fall out of favor since I've learned there's no way to send mismatching tuples (compiler error)

src/avram/bulk_upsert.cr
robcole: Resolved! 🙇 compiler

robacarp:
After looking over this and watching the discussion, I'm aware that there were at least three different assumptions about what a bulk insert/upsert might be used for. I observed these assumptions:

- Bulk upsert would be used for consuming a CSV or other finite data source, outside any http request cycle. I think this is the assumption under which much of this code is written. This might be a one-time job (switching database engines) or it might be a daily job (ingesting some CSV from an LMS, etc).
- Stream consumption. An infinite data source which may be moving incredibly fast and never end. This was my assumption and [Twitter's Compliance Firehose](https://developer.twitter.com/en/docs/twitter-api/enterprise/compliance-firehose-api/overview) is one such datastream. 
- Backing for a REST API endpoint which would allow some api customer to post bulk data. For example marking a collection of emails as "Read" in an inbox app.

They're not fundamentally incompatible, nor is it absolutely necessary that whatever is implemented here facilitate all three of these. I do think it might be helpful to identify the use cases which need to be discussed. It might provide some clearer guidance around validations, compile time guards, etc.

grepsedawk:
> This might be a one-time job (switching database engines) or it might be a daily job (ingesting some CSV from an LMS, etc).

Or inside the req cycle, like a user uploads a csv

grepsedawk:
https://github.com/will/crystal-pg/pull/244 needs to be merged/released or monkeypatched into avram before this is viable
/cc @robcole

src/avram/upsert.cr
```diff
@@ -90,6 +90,24 @@ module Avram::Upsert
...
+    def self.bulk_upsert(upserts)
+      operations = upserts.map do |upsert_args|
+        new(**upsert_args)
```
grepsedawk, line 96: As of now, when a user passes a mismatching array, the compiler throws an error that double splats don't support unions
It would be _awesome_ if we could catch this and provide a clean message that all records must have matching keys

spec/avram/operations/save_operation_spec.cr
```diff
@@ -307,6 +308,126 @@ describe "Avram::SaveOperation" do
     end
   end
 
+  describe ".bulk_upsert" do
+    # TODO: Is this possible? How do we write this?
+    it "should only proc one query" do
```
robcole, line 313: @jwoertink @matthewmcgarvey Any thoughts on how we might be able to test this?

src/avram/save_operation.cr
robcole: Not sure given this is a class method and `delegate` is a macro, but happy to take a look into it more if you think it's worth investigating.

grepsedawk: I don't think this is a huge deal, it'd just be 1 stack trace jump out of "framework error full traces"

src/avram/bulk_upsert.cr
```diff
@@ -85,10 +85,6 @@ class Avram::BulkUpsert(T)
     "#{column_types[column.name.to_s]}[]"
   end
 
-  private def cast(column : Avram::Attribute(Time))
```
grepsedawk: <3

spec/avram/operations/save_operation_spec.cr
grepsedawk: https://github.com/luckyframework/avram/blob/ac281f490a85283eaf8b74fd23357dd70a6d7f22/spec/avram/queryable_spec.cr#L35-L42

src/avram/upsert.cr
grepsedawk: The error when it doesn't compile should be something like:
> You must pass matching NamedTuples into upsert

rather than complaining about a `**`, which isn't going to be clear to newcomers

src/avram/upsert.cr
```diff
@@ -90,6 +90,24 @@ module Avram::Upsert
     end
   end
 
+  macro upsert_unique_on(*attribute_names)
+    def self.bulk_upsert(upserts)
```
grepsedawk, line 94: if we rename (and add types) to this and the below bulk_upsert, overloading will work exactly as expected, even with upsert_unique_on and the other upsert_lookup macro being split for now!

src/avram/upsert.cr
grepsedawk: https://play.crystal-lang.org/#/r/cmgb is a viable approach here

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
...
+    end.to_h
+  end
+
+  private def cast(column)
+    "#{column_types[column.name.to_s]}[]"
+  end
+
+  private def value_placeholders
+    permitted_attributes.map_with_index(1) do |k, index|
+      "$#{index}::#{cast(k)}"
```
matthewmcgarvey, line 90: Should probably pass in `column_types` to `cast` and fetch `column_types` before this loop instead of rebuilding that hash on each iteration

matthewmcgarvey: In fact, `cast` doesn't seem to be providing much value anymore now that it looks like you got rid of the type restrictions

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
+class Avram::BulkUpsert(T)
+  def initialize(@records : Array(T),
+                 @conflicts : Array(Symbol),
+                 @permitted_fields : Array(Symbol))
```
matthewmcgarvey, line 4: Probably don't need to set this as the instance variable since you're overwriting it 3 lines later

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
+class Avram::BulkUpsert(T)
+  def initialize(@records : Array(T),
```
matthewmcgarvey, line 2: Don't need to set instance variable when re-setting it just below

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
...
+  def args
+    @records.map do |record|
+      permitted_attributes(record).map(&.value)
+    end.transpose
+  end
+
+  private def permitted_fields_for(fields : Array(Symbol))
+    fields.push(:created_at) if @sample_record.responds_to?(:created_at)
+    fields.push(:updated_at) if @sample_record.responds_to?(:updated_at)
+    fields.uniq
```
matthewmcgarvey, line 28: Already adding to the passed in array with the `.push`, why not go ahead and use `.uniq!` instead of making a new array

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
...
+  private def permitted_attributes
+    permitted_attributes(@sample_record)
+  end
+
+  private def conflicts
+    @conflicts.join(", ")
+  end
+
+  private def set_timestamps(collection)
+    collection.map do |record|
```
matthewmcgarvey, line 46: This could be a `.each` and won't need to have the last line of the block returning the record

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
...
+  private def permitted_attribute_column_names
+    permitted_attributes.map(&.name)
+  end
+
+  private def fields
+    permitted_attribute_column_names.map(&.to_s).join(", ")
+  end
+
+  private def column_types : Hash(String, String)
+    T.database_table_info.not_nil!.columns.map do |col_info|
```
matthewmcgarvey, line 76: The `.not_nil!` could probably be moved to `database_table_info`

src/avram/bulk_upsert.cr
```diff
@@ -0,0 +1,93 @@
...
+      [
+        col_info.column_name,
+        col_info.data_type,
+      ]
```
matthewmcgarvey, line 80: Let's make this a tuple instead of an array

src/avram/save_operation.cr
```diff
@@ -367,6 +367,10 @@ abstract class Avram::SaveOperation(T)
     {{ T.constant(:PRIMARY_KEY_NAME).id }}.value.nil?
   end
 
+  def changed_attributes
+    column_attributes.select(&.changed?)
+  end
```
matthewmcgarvey, line 372: Looks to be unused. Could it be removed?

src/avram/upsert.cr
```diff
@@ -90,6 +90,27 @@ module Avram::Upsert
     end
   end
 
+  macro upsert_unique_on(*attribute_names)
+    def self.upsert(upserts : Array(X)) forall X
+      \{%
+        if X.union?
```
matthewmcgarvey, line 96: Should another check/error be added in case `X` isn't a `NamedTuple`? Only if the error would be more helpful than the current default. The double splat below probably forces it to the type anyways

matthewmcgarvey (commented):
Catching up on all the work y'all have done and threw some comments in there 👍

src/avram/upsert.cr
robcole: Yeah, double splat will catch it.

src/avram/save_operation.cr
robcole: Yep, remnant of precious experiments

robcole:
> Nice to haves:
> 
> * [ ]  Throw a clean error when UK does not exist

@grepsedawk totally forgot about this one, lmk if you want to pair on it later this week too

src/avram/upsert.cr
```diff
@@ -119,8 +122,8 @@ module Avram::Upsert
       end
     {% end %}
 
-    def self.bulk_upsert(_upserts)
-      \{% raise "Please use the 'upsert_unique_on' macro in #{@type} before using '.bulk_upsert'" %}
+    def self.upsert(_upserts : Array)
+      \{% raise "Please use the 'upsert_unique_on' macro in #{@type} before using '.upsert'" %}
```
grepsedawk, line 126:
```suggestion
      \{% raise "Please use the 'upsert_unique_on' macro in #{@type} before using '.upsert(Array)'" %}
```

spec/avram/operations/save_operation_spec.cr
```diff
@@ -307,6 +312,126 @@ describe "Avram::SaveOperation" do
...
+          u.name("Name 1")
+          u.nickname("Nickname 1")
+          u.age(42)
+          u.year_born(1960)
+          u.joined_at(Time.utc)
+        end
+      end
+
+      it "allows manual passing of updated_at, but ignores created_at" do
+        some_time = Time.utc(2016, 2, 15, 10, 20, 30)
```
grepsedawk, line 332: Fun fact, I took this from the crystal docs for Time, and so did a lot of other people already: https://cs.github.com/?scopeName=All+repos&scope=&q=Time.utc%282016%2C+2%2C+15%2C+10%2C+20%2C+30%29

src/avram/upsert.cr
robcole: added better errors 👍

spec/avram/operations/save_operation_spec.cr
```diff
@@ -307,6 +314,149 @@ describe "Avram::SaveOperation" do
...
+
+        records = UpsertUserOperation.upsert([insert])
+        records.first.id.should_not eq nil
+        records.first.created_at.should eq some_time
+        records.first.updated_at.should eq some_time
+      end
+    end
+
+    # Crystal takes care of ordering our tupled array for us, but this helps
+    # protect for the future: https://play.crystal-lang.org/#/r/cm53
```
robacarp, line 433: This is not something I'd have thought of, but it seems like a good test. This comment here is _probably_ unnecessary, and I have a vague notion that the link will become invalid in a short time.

spec/avram/operations/save_operation_spec.cr
```diff
@@ -307,6 +314,149 @@ describe "Avram::SaveOperation" do
...
+    context "when the tuples are passed in different orders" do
```
robacarp, line 434:
```suggestion
    context "when the tuple values are passed in different orders" do
```

Nitpicky English

src/avram/upsert.cr
```diff
@@ -90,6 +90,29 @@ module Avram::Upsert
     end
   end
 
+  macro upsert_unique_on(*attribute_names)
+    def self.upsert(upserts : Array(X)) forall X
+      \{%
+        if X > NamedTuple
+          raise("All array elements for #{@type}.upsert must be NamedTuples. Given: #{X}")
```
robacarp, line 97:
```suggestion
          raise("All array elements for #{@type}.upsert must be NamedTuples. You provided: #{X}")
```

I'm not sure what comparative exceptions around Avram might be. "You provided:" isn't my favorite but "Given:" feels like it doesn't lead me in the right direction. Similarly below.

robacarp (commented):
I was expecting to see a bunch of pivoty maxtrixy nonsense here to accomplish this, but it's all readable! I left a few nitpicky comments. 

I'm not familiar enough with the upsert pattern in postgres to really attest to the approach here, but there are no red flags in the way it's built to my eye.

src/avram/upsert.cr
robcole: Makes sense, I didn’t love Given either but was aiming for terseness - “you provided” def reads better

spec/avram/operations/save_operation_spec.cr
robcole: Removed the comment, thanks!

jwoertink (commented):
Honestly, this is pretty chonky, but it doesn't look "crazy".... I don't have any immediate use for it, but from what it sounds like, @grepsedawk used it and it seems to work?

The only thing I'm not seeing (and it may be ok for now) is no way to use this with Lucky::Params, right?

@robcole Can you just update the original post for this PR with an example code of how someone would use this? That will give me (and other readers) an easy overview of what this PR is really doing.

robcole:
> Honestly, this is pretty chonky, but it doesn't look "crazy".... I don't have any immediate use for it, but from what it sounds like, @grepsedawk used it and it seems to work?
> 
> The only thing I'm not seeing (and it may be ok for now) is no way to use this with Lucky::Params, right?
> 
> @robcole Can you just update the original post for this PR with an example code of how someone would use this? That will give me (and other readers) an easy overview of what this PR is really doing.

It was designed for bulk importing which isn’t something I often do with params. I’ll see if I can find an example where it might be a fit, but would usually be iterating over a collection of responses from some source (API, CSV, or internal model building typically). Will add the examples in using it for today or tomorrow. 

I think you’d just need an array of params since we use the same general interface as upsert and it should work fine, but I haven’t tested that yet.

jwoertink:
> I think you’d just need an array of params since we use the same general interface as upsert 

Ok, I think that's fine. We can worry about adding in support from params later then. I think your usage makes sense, so just a quick code example. Just a few lines, nothing crazy, and that would be good!

matthewmcgarvey (approved):

jwoertink:
Overall, I love the direction, and I think this is a super useful feature. There's just a few things that concern me, so I'd like to talk about possible solutions, or direction we can take. And please forgive me if any of this is incorrect and I missed something.

Right now, this implementation does a piggy back off of SaveOperations to help keep things a bit more consistent, and make use of some existing methods so we don't end up duplicating things :+1:  However, this leads to several new issues...

For example, say I have a normal `SaveUser` operation.

```crystal
  upsert_lookup_columns :email, :username
  upsert_unique_on :email, :username
  needs passcode : Int32
  attribute password : String

  before_save do
     validate_required(password)
      validate_email_format(email)
     validate_unique(username)
   end
   after_commit do |user|
     EmailUser.new(user, passcode).deliver
   end
end

# This works fine
SaveUser.upsert(passcode: 1234, email: "...", username: "...") do |op, user|
end

# This leads to some issues
SaveUser.upsert([{email: "...", username: "..."}])
```

* The compiler would fail because `needs` isn't passed in
* None of the validations or callbacks would run
* The interface is different which could make it confusing.
* It doesn't currently support `params`

So at this level, it's almost not even a SaveOperation.... The lack of validations and callbacks is the biggest part I think. Assuming you're pulling in bulk data from a CSV in which you don't control, there's no way to guarantee that a malformed email won't get inserted in to your DB. Since postgres won't handle validations other than NULL or not, and possible unique (if you have the index), then your app would be at-risk of bad data.

Here's my suggestions, but I'm totally open to others as well.

1. We rename this `upsert` method to something else. 
  - PRO: Less confusion with the current upsert method
  - PRO: Allows us to mentally separate the interface so we don't necessarily need to conform to the `method() {}` `method!` API
  - CON: It doesn't solve the SaveOperation issues with callbacks, validations, etc...
2. We create a new operation type like `Avram::SQLOperation` or something... `Avram::BulkOperation`...
  - PRO: We completely remove ourselves from trying to cram in to a current setup that doesn't really work
  - PRO: We could keep the current name, and usage since it's a whole different object type
  - PRO: It could be expanded later for other uses...
  - CON: It's just another operation type... Which also means even more code to manage
  - CON: It's going to require duplicating a lot of code. (to me, this isn't completely horrible provided we have enough specs...)
3. We leave this PR as-is, but we add the `@[Experimental()]` annotation, and ensure that it's documented as a "at your own risk" feature.
  - PRO: We can merge this sooner than later.
  - CON: We have to come back to this later since it's incomplete.

I'm sure there's other options I may not be thinking of....

@robcole, I know you and @grepsedawk have put in a ton of work in to this, and I really appreciate it :heart: . I want to make sure it's clear that I'm not shooting down all your efforts and hard work. I think having this feature in Avram will be huge! We just need to make sure that features like this live up to Lucky's core goals. I'd be happy to also hop on a voice chat in discord if either of you (or anyone else) would like to chat more about potential solutions, or just to hash out any concerns.

grepsedawk:
I side towards 3.
I really do feel the proper name for this is in a `SaveOperation.upsert(Array)` and this would allow us to vet out real life uses as time moves on.
Also, WORST CASE with 3 is a user has 2 save operations: 1 for bulk things and 1 for normal things

It allows us all to feel out the API a bit more before making any big changes

robcole:
> I side towards 3. I really do feel the proper name for this is in a `SaveOperation.upsert(Array)` and this would allow us to vet out real life uses as time moves on. Also, WORST CASE with 3 is a user has 2 save operations: 1 for bulk things and 1 for normal things
> 
> It allows us all to feel out the API a bit more before making any big changes
> 
> Also, I could be wrong but I think the way @robcole did the splat it would actually support the `needs` of the save operation!

I also side with 3, but with a note that when we bring in sql operations or bulk operations, I’m 💯 down to work on refactors to bring the two interfaces into more uniformity. I’d like to have a sensible way to have it still optionally yield an operation, but for the current code, it doesn’t feel like it brings enough value to change the API around that much. 

A few things I’m less certain of for bulk upsert in general based on how I’ve used it:

- I’ve never wanted callbacks to run when using it. By default they don’t in Rails, so this is something that could be an option that gets added, but by default, it isn’t something I’ve ever wanted. Frequently these updates have been 500k+ inserts over a matter of seconds, so having callbacks could really be problematic on some common workloads. 

- I think validations should always be optional. I can understand those being a lot more important and seems like an easy extension to add, once there’s some notion of how we want to handle “3 of 50 objects failed validation — what do you want to do?” Similar arguments around collection size might apply here. 

The way I designed it should be a somewhat easy port for things like params once we have the following concepts in Lucky to make sure everything feels smooth.

robcole:
@jwoertink In the event you're curious - https://github.com/zdennis/activerecord-import is where I got a lot of the API design inspiration (on top of your existing upsert code) as it's what I used up until Rails finally added it. Validations are optional in it, but the interface around reporting and fixing those validations can be a bit wonky (basically you're doing `collection.select(&:invalid?).map(&:errors)`.

For the recent Rails additions, AFAIK these are all intended on being bulk operations, so callbacks and validations are skipped as well: https://www.bigbinary.com/blog/bulk-insert-support-in-rails-6

Given this LMK if you've got more thoughts here - definitely appreciating the feedback as it'll help me make sure I'm on the same page in general with Lucky API goals.

jwoertink:
Oh sweet. Thanks for the links @robcole. I'll take a look at those.

Pulling things from rails can be nice since it's so well established, but it's also a double edged sword. Since rails isn't type-safe, and doesn't enforce any sort of data integrity, you're more likely to have bad data which leads to bugs in production. For example, in rails, an empty string is fine. If the column is `String` and set to not null in postgres, an empty string would be valid, but in Lucky that's not the case. Empty strings are treated like `nil`. We also make required column validations required. They're more "opt-out" than they are "opt-in".

I just had another idea.... Since one of the goals for Lucky is to try and always provide an escape hatch for when there's something we can't do at a high-level, what if we just dropped this whole API down a level?

For example, instead of doing this off the SaveOperation, it becomes lower level like

```crystal
upsert = Avram::BulkUpsertt(Thing).new(array_of_named_tuples_for_thing)
upsert.run
```

After writing that, I didn't really like it, but the premise is still good... At a lower level, we don't need to necessarily follow Lucky convention. This is similar to when you do something like `UserQuery.new.where("complex sql ?", "no safety")`. That's a small snippet, but the same idea. By doing that, then later on we could create a whole new high level wrapper around it that includes all the other stuff we want!

What are your thoughts on that?

src/avram/upsert.cr
```diff
@@ -100,5 +123,9 @@ module Avram::Upsert
         \{% raise "Please use the 'upsert_lookup_columns' macro in #{@type} before using '{{ method.id }}'" %}
```
grepsedawk, line 123: image
This is an interesting case
I was passing a map directly, which didn't fall through to self.upsert(Array), OR the standard upsert, but was caught up by this method's error when I passed an Enumerable (from the map)

src/ext/db/param.cr
```diff
@@ -0,0 +1,8 @@
+# Can be removed once https://github.com/will/crystal-pg/pull/244 is merged.
```
grepsedawk, line 1: This can be removed now https://github.com/luckyframework/avram/pull/807

jwoertink:
@robcole After talking it over with the team, there's been some suggestions on what if we made this a separate shard extension similar to how we did https://github.com/luckyframework/avram_slugify? If we go that route, you could keep the current implementation as-is. This would allow people to use the feature, and give feedback / contribute to really flesh out the different use cases. If you're cool with that, I could even start a new repo under Lucky and I think I can give you access to that. Let me know your thoughts, or you can hit me up on discord if you want :+1:

robcole:
@jwoertink I think it makes more sense to have this in Avram, but I agree that it isn’t ready yet. Have you done any work on the concept of a BulkOperation or SQLOperation?

My plan was to refactor this and add a BulkOperation which yields the operation in the same way as others, but would have a concept of multiple records, and initially wouldn’t handle validations and callbacks, but could down the road. Haven’t had time to get to the refactor yet but it’s on my list still.

jwoertink:
Oh, ok. Yeah, I agree that it should be in Avram core too, but just not with the current implementation. If you're open to some refactors, then we can keep this open :+1:  

I haven't started on any of the bulk stuff. Work has just been way too crazy lately, so I've fallen behind a bit on my Lucky duties :frowning_face:  Once I get back in the swing of things, I do want to tackle bulk stuff in general. I have a few ideas on how I think we can handle it, I just need to put that stuff down on paper :sweat_smile:

jwoertink:
Sweet! Stoked. And no rush on any of that. I don't plan on another release until around summer, and if it's not ready, we can always do the next release :smile:  I just wanted to make sure you weren't waiting on me for anything. Sounds like we're on the same page though :+1:  Thanks for being understanding!

Issue #672, opened 2021-04-28
Issue preloading soft deleted records


jwoertink:
I'll need to investigate this more, but we just ran in to an issue with some preloads.

Take this model setup

```crystal
class User < BaseModel
  table do
    # ...
    column soft_deleted_at : Time?
    has_many transactions : Transaction
  end
end

class Transaction < BaseModel
  table do
    belongs_to user : User
  end
end

class User::BaseQuery
  include Avram::SoftDelete::Query

  def initialize
    defaults &.only_kept
  end
end

class UserQuery < User::BaseQuery
end

class TransactionQuery < Transaction::BaseQuery
end
```

Then we had some code like this:

```crystal
# in the action
transactions = TransactionQuery.new.preload_user

# in the page

transactions.each do |transaction|
  td do
    link(transaction.user.username, to: Users::Show.with(transaction.user_id))
  end
end
```

We kept getting a 500 error saying `NilAssertionError` which generally relates to calling `not_nil!` somewhere. In this case, our `user` should never be nil, but calling `transaction.user` says otherwise. We printed out the record, and saw this:

```
#<Transaction:0x7f9cdd05a8f0 @_user_preloaded=true, @_preloaded_user=nil, @id=1, @created_at=2021-04-28 18:02:47.0 UTC, @updated_at=2021-04-28 18:02:47.0 UTC, @user_id=4>
```

Looking up user 4, we saw the record existed, but it was soft deleted. 

Avram should probably log some sort of warning that a required record could not be preloaded, or maybe when you preload an association that's been soft deleted, that record isn't returned at all?

matthewmcgarvey:
Another option would be to raise an error. Transaction specifies that user association will be there but it isn't so something is wrong

jwoertink:
Yeah, that's a good idea. It took a long time to figure out that error, so throwing an error that's easier to spot what's going on would make things a lot nicer.

AND.... to get around this, the proper query was

```crystal
transactions = TransactionQuery.new.preload_user.where_user(UserQuery.new)
```

So maybe when we raise the error, we can mention something like that.

jwoertink:
Just ran in to this again, and it it gets a bit more complicated. 

```crystal
class User < BaseModel
  table do
    # ...
    has_many following_follows : Follow, foreign_key: :follower_id
    has_many follower_follows : Follow, foreign_key: :followee_id
    has_many followers : User, through: [:follower_follows, :follower]
[...]

PR #812, opened 2022-02-15
Raise error if required association not loaded in preload

matthewmcgarvey:
Related #672 

If a model defines a required belongs to association and the record isn't retrieved in a preload, this raises an appropriate error. Also added for required has one associations.

Applying this to the related issue, in the scenario where you have soft deletes, the solution now becomes making what was a required association optional since it might not always be loaded. This doesn't change whether or not it is loaded but will guarantee an error unless you change the association. Before, the error only occurred if you tried to access the association on the particular record.

### DIFF
--- a/src/avram/errors.cr
+++ b/src/avram/errors.cr
@@ -30,6 +30,12 @@ module Avram
     end
   end
 
+  class MissingRequiredAssociationError < AvramError
+    def initialize(model : Avram::Model.class, association : Avram::Model.class)
+      super "Expected #{model} to have an association with #{association} but one was not found."
+    end
+  end
+
   # Raised when a validation is expecting an impossible constraint
   class ImpossibleValidation < AvramError
     def initialize(attribute : Symbol, message = "an impossible validation")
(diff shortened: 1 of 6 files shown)


src/avram/errors.cr
```diff
@@ -30,6 +30,12 @@ module Avram
     end
   end
 
+  class MissingRequiredAssociationError < AvramError
+    def initialize(model : Avram::Model.class, association : Avram::Model.class)
+      super "Expected #{model} to have an association with #{association} but one was not found."
```
matthewmcgarvey, line 35: Error message could be better. Any ideas?

Current:
Screen Shot 2022-02-14 at 11 33 36 PM

jwoertink (approved):
Ok, this wasn't exactly what I was thinking, but I think this is still pretty good :+1:

src/avram/errors.cr
robacarp: How about `Expected Comment to be assigned to a Post but no association was supplied`?

Or `Expected Comment to belong to a Post but no Post was supplied`

matthewmcgarvey: Whoops, forgot to wait for feedback on this 😬 Feel free to open a PR with an error message that you think is better. Otherwise I _might_ remember to make one at some point

robacarp: Well. I could probably have also looked to see if it was merged before trying to be helpful...

PR #820, opened 2022-03-06
Add a default value to the default timestamp columns.

### LINKED ISSUE #816: add_timestamps should have defaults, for manual database manipulation
https://github.com/luckyframework/avram/blob/9251a965372d319977558bcbe3d0696f60aeb45e/src/avram/migrator/create_table_statement.cr#L84

This should provide defaults of `CURRENT_TIMESTAMP()` for the two times. It makes it easier for someone using `psql` to add a record manually, as they won't have to formulate a time constant.

jwoertink:
Fixes #816

When generating a new table and using the `add_timestamps`, this will now generate the `created_at` and `updated_at` columns defaulting to `NOW()`. The values will still come from within Crystal since the SaveOperation will set them, however, this allows you to hop into psql real quick and create a new record without having to worry about setting these.

Updating a record will still require you to update `updated_at` manually since that would require setting up a trigger.

### DIFF
--- a/spec/avram/migrator/create_table_statement_spec.cr
+++ b/spec/avram/migrator/create_table_statement_spec.cr
@@ -29,8 +29,8 @@ describe Avram::Migrator::CreateTableStatement do
     built.statements.first.should eq <<-SQL
     CREATE TABLE users (
       id serial PRIMARY KEY,
-      created_at timestamptz NOT NULL,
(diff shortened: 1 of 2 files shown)


robcole:
This will simplify a lot of the extra logic I needed for bulk inserts as well. 🙏

PR #821, opened 2022-03-07
Adding new Slugify.generate method to generate a slug value

### LINKED ISSUE #813: Slugify doesn't allow you to update the slug
https://github.com/luckyframework/avram/blob/9251a965372d319977558bcbe3d0696f60aeb45e/src/avram/slugify.cr#L38

Not sure if this was intentional or not... I can see there being some issues where a slug could potentially clash.

In my case, sometimes users will mistype their username during signup. We don't allow them to change it, but if they request a fix, we can manually update it for them. In that instance, we would want to make sure the slug matches. It's pretty confusing when you add in the slugify, and nothing happens... no errors, nothing. It just doesn't update because the `slug.value` is not blank here. I'd even be ok with an error saying slug is already filled or something. Then you could at least `slug.value = nil` before hand.

jwoertink:
Fixes #813

This method will generate a String value that can be used for a slug, but without setting any attribute value for you. This would be more used for an escape hatch than anything. Since it's not passing in the slug column, there's no way to query to know if it's been taken already. This means we can't do the query lookup to ensure the value is unique.

```crystal
before_save do
  if title.changed?
    slug.value = Avram::Slugify.generate(using: title)
  end

  validate_uniqueness slug, query: ArticleQuery.new.slug
end
```

### DIFF
--- a/spec/avram/slugify_spec.cr
+++ b/spec/avram/slugify_spec.cr
@@ -34,14 +34,6 @@ describe Avram::Slugify do
 
         op.slug.value.should eq("software-developer")
       end
-
-      it "it sets slug from a single string" do
-        op = build_op
-
-        slugify(op.slug, "Software Developer")
-
-        op.slug.value.should eq("software-developer")
-      end
     end
 
     describe "with an array of slug candidates" do
@@ -121,6 +113,40 @@ describe Avram::Slugify do
       op = build_op(title: "The Boss")
       slugify(op.slug, op.title, ArticleQuery.new.title("A"))
       op.slug.value.to_s.should start_with("the-boss-") # Has UUID appended
+    end
+  end
+
+  describe ".generate" do
+    it "skips blank slug candidates" do
+      slug = Avram::Slugify.generate(["", "Software Developer"])
+
+      slug.should eq("software-developer")
+    end
+
+    describe "with a single slug candidate" do
+      it "sets slug from a single attribute" do
+        op = build_op(title: "Software Developer")
+        slug = Avram::Slugify.generate(op.title)
+
+        slug.should eq("software-developer")
+      end
+
+      it "sets slug from a single string" do
+        slug = Avram::Slugify.generate("Software Developer")
+
+        slug.should eq("software-developer")
+      end
+    end
+
+    describe "with an array of slug candidates" do
+      it "sets when using multiple attributes" do
+        op = build_op(title: "How Do Magnets Work?", sub_heading: "And Why?")
+
+        slugify(op.slug, [[op.title, op.sub_heading]])
+        slug = Avram::Slugify.generate([[op.title, op.sub_heading]])
+
+        slug.should eq("how-do-magnets-work-and-why")
+      end
     end
   end
 end
--- a/spec/avram/slugify_spec.cr
+++ b/spec/avram/slugify_spec.cr
@@ -118,7 +118,8 @@ describe Avram::Slugify do
 
   describe ".generate" do
     it "skips blank slug candidates" do
-      slug = Avram::Slugify.generate(["", "Software Developer"])
+      op = build_op(title: "Software Developer")
+      slug = Avram::Slugify.generate(op.slug, ["", "Software Developer"], ArticleQuery.new)
 
       slug.should eq("software-developer")
     end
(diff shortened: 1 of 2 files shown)


spec/avram/slugify_spec.cr
```diff
@@ -34,14 +34,6 @@ describe Avram::Slugify do
...
         op.slug.value.should eq("software-developer")
       end
-
-      it "it sets slug from a single string" do
-        op = build_op
-
-        slugify(op.slug, "Software Developer")
-
-        op.slug.value.should eq("software-developer")
-      end
```
jwoertink, line 44: turns out this was a duplicate spec with the one right above it.

matthewmcgarvey (commented):
Not sure how useful it would be without the ability to check the slug in the database. another option would be to pass in everything like the other method but not to check the value. Or maybe just the column name instead of the whole attribute but that might be confusing

jwoertink:
Yeah, I was kind of stumped on this one too. I went back and forth with it a few times and finally just said I'll push something up so we can hash it out. 

I agree passing in just the column name might be confusing, and we'd lose the type-safety there anyway.

Maybe what we could do is something like..

```crystal
def set(...)
  if slug.value.blank?
    slug.value = generate(...)
  end
end
```

and just have generate handle the logic ... :thinking:

src/avram/slugify.cr
```diff
@@ -49,6 +64,22 @@ module Avram::Slugify
...
+  def generate(using slug_candidates : Array(String | Avram::Attribute(String) | Array(Avram::Attribute(String)))) : String
+    slug_candidates = format_candidates(slug_candidates)
+
+    slug_candidates.first? || UUID.random.to_s
+  end
+
+  private def format_candidates(slug_candidates : Array(String | Avram::Attribute(String) | Array(Avram::Attribute(String)))) : Array(String)
+    slug_candidates.map do |candidate|
+      parameterize(candidate)
+    end.reject(&.blank?)
```
robacarp, line 80: Should this include a call to [`#strip`](https://crystal-lang.org/api/1.2.2/String.html#strip%3AString-instance-method) as well?

src/avram/slugify.cr
```diff
@@ -49,6 +64,22 @@ module Avram::Slugify
...
   end
 
+  def generate(using slug_candidate : Avram::Attribute(String) | String) : String
+    generate([slug_candidate])
+  end
+
+  def generate(using slug_candidates : Array(String | Avram::Attribute(String) | Array(Avram::Attribute(String)))) : String
+    slug_candidates = format_candidates(slug_candidates)
+
+    slug_candidates.first? || UUID.random.to_s
```
robacarp, line 74: A uuid fallback would not be my expectation here, but I'm not sure what would have been my expectation. A slug is usually designed to be more "human friendly" than a UUID... so I think I'd just expect it to raise if it couldn't build a slug.

spec/avram/slugify_spec.cr
```diff
@@ -123,6 +115,40 @@ describe Avram::Slugify do
...
+
+        slug.should eq("software-developer")
+      end
+    end
+
+    describe "with an array of slug candidates" do
+      it "sets when using multiple attributes" do
+        op = build_op(title: "How Do Magnets Work?", sub_heading: "And Why?")
+
+        slugify(op.slug, [[op.title, op.sub_heading]])
```
robacarp, line 145: What's this line do? Is it proving a compile time check?

jwoertink: I put it there just to make sure you were awake :stuck_out_tongue:  actually, it's just a copy/paste error. I need to remove it :sweat_smile:

jwoertink:
Alright, here's an update that moves all of the logic to `generate`. Then `set` just looks to see if the slug has a value already and if not, calls generate to set it.

I started to add in an error and have `generate` raise the error if no slug could be generated, but then I saw this spec

https://github.com/luckyframework/avram/blob/c798fbbda4cd9bd40ad9800022b4c6203e6f4fe7/spec/avram/slugify_spec.cr#L101-L111

basically this allows it to be `nil`. I decided that does make sense because it would be up to you to add in the validations if this field was required or not. So I turned generate in to `String?` which just returns nil if no slug is generated.

src/avram/slugify.cr
```diff
@@ -36,19 +53,40 @@ module Avram::Slugify
...
+    result = slug_candidates.find { |candidate|
+      query.where(slug.name, candidate).none?
+    }
+
+    if result.nil?
+      if candidate = slug_candidates.first?
+        "#{candidate}-#{UUID.random}"
+      end
+    else
+      result
```
matthewmcgarvey, line 80:
```suggestion
    if result
      result
    elsif candidate = slug_candidates.first?
      "#{candidate}-#{UUID.random}"
```

To remove some nesting

src/avram/slugify.cr
```diff
@@ -23,6 +23,23 @@
...
+#
+# ```
+# class SaveArticle < Article::SaveOperation
+#   before_save do
+#     if title.changed?
+#       slug.value = Avram::Slugify.generate(using: title)
+#       validate_uniqueness_of slug, query: ArticleQuery.new.slug
+#     end
+#   end
+# end
```
matthewmcgarvey, line 41: This needs to be updated I think

matthewmcgarvey (approved):

PR #822, opened 2022-03-07
Adding an overload for this validation with a Range

### LINKED ISSUE #563: validates_inclusion_of(parameter, in: Range) doesn't compile
Possibly related: #537 

Validate inclusion of technically takes a Range, because it's typed to an enumerable. However, this fails to compile: `validate_inclusion_of http_status_code, in: 100..599`

```
In src/operations/monitor/save_http.cr:14:1
 14 | validate_inclusion_of http_status_code, in: 100..599
  ^--------------------
Error: instantiating 'validate_inclusion_of(Avram::PermittedAttribute(Int32 | Nil))'

In lib/avram/src/avram/validations.cr:102:27
 102 | unless allowed_values.includes? attribute.value
                             ^--------
Error: instantiating 'Range(Int32, Int32)#includes?((Int32 | Nil))'
In /usr/local/Cellar/crystal/0.35.1_1/src/range.cr:282:32
 282 | (begin_value.nil? || value >= begin_value) &&
                                  ^
Error: undefined method '>=' for Nil (compile-time type is (Int32 | Nil))

Nil trace:
  /usr/local/Cellar/crystal/0.35.1_1/src/range.cr:282
        (begin_value.nil? || value >= begin_value) &&
  /usr/local/Cellar/crystal/0.35.1_1/src/range.cr:277
      def includes?(value)
```

It fails because Range#includes? doesn't take a nil and know what to do with it. Over on #537 there's demonstration that numeric range validation is desired. 

Depending on what is the desired API, something like this overload could be added to resolve the problem. In the mean time I've added it to my SaveOperation.

```
  def validate_inclusion_of(attribute, in allowed_values : Range, allow_nil : Bool = false)
    nil_message = "is required"
    range_message = "must be between #{allowed_values.begin} and #{allowed_values.end}"

    value = attribute.value
    if value.nil?
      attribute.add_error nil_message unless allow_nil
      return
    end

    unless allowed_values.includes? value
      attribute.add_error range_message
    end
  end
```

jwoertink:
Fixes #563

Since you can't allow nil, I've added a method overload that raises a compile-time error letting you know that you can't allow nil with ranges.

### DIFF
--- a/spec/avram/validations_spec.cr
+++ b/spec/avram/validations_spec.cr
@@ -260,6 +260,11 @@ describe Avram::Validations do
       result.should eq(true)
       allowed_name.valid?.should be_true
 
+      allowed_number = attribute(4)
+      result = Avram::Validations.validate_inclusion_of(allowed_number, in: 0..5)
+      result.should eq(true)
+      allowed_number.valid?.should eq(true)
+
       forbidden_name = attribute("123123123")
       result = Avram::Validations.validate_inclusion_of(forbidden_name, in: ["Jamie"])
       result.should eq(false)
--- a/src/avram/validations.cr
+++ b/src/avram/validations.cr
@@ -145,6 +145,43 @@ module Avram::Validations
     no_errors
   end
 
+  def validate_inclusion_of(
+    attribute : Avram::Attribute(T),
+    in allowed_values : Range(T, T),
+    message : Avram::Attribute::ErrorMessage = Avram.settings.i18n_backend.get(:validate_inclusion_of)
+  ) : Bool forall T
+    no_errors = true
+    if value = attribute.value
+      if !allowed_values.includes?(value)
+        attribute.add_error(message)
+        no_errors = false
+      end
+    else
+      attribute.add_error(message)
+      no_errors = false
+    end
+
+    no_errors
+  end
+
+  def validate_inclusion_of(
+    attribute : Avram::Attribute(T),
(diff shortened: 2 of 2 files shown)


matthewmcgarvey (commented):
What if in the overload for Range we chec if attribute is nil and we set an error if `allow_nil` is false or return if it's allowed. Then we can check as normal now that the value is guaranteed to not be nil.

I'm wary of the compile time error because it seems like we can work around it and do what is expected.

robacarp:
The rails validation pattern [matches up](https://guides.rubyonrails.org/active_record_validations.html#allow-nil) with that definition of `allow_nil: true`. In the case of Rails, the allow_nil: true check happens elsewhere and skips the validation entirely, but I don't think that matters.

matthewmcgarvey (commented):
I'm finding this difficult to explain, let me know if this doesn't make sense:

The way we define `Avram::Attribute`, it's never `Avram::Attribute(String?)` it's always `Avram::Attribute(String)`. So the `T` shared between `Avram::Attribute` and `Enumerable` means that the enumerable can't have a nil value. So, I think it might make sense to make this new implementation the default (Just making allowed_values an enumerable). Thoughts?

jwoertink:
That's interesting. I think you might be right. I'm down to give it a shot. So instead of having this separate overload, we just make the current method implementation use this instead.

matthewmcgarvey (approved):
🎉

Issue #825, opened 2022-03-08
Support other Int types for enums


jwoertink:
https://github.com/luckyframework/avram/blob/c798fbbda4cd9bd40ad9800022b4c6203e6f4fe7/src/avram/charms/enum_extensions.cr#L10

If you're using cockroachdb, you may have to use Int64. Right now, enums only support Int32, but it would be awesome if we had a way to get that type value from the enum

```crystal
enum Type : Int64
end
```

Then as long as we support the Int, things should all work.

/cc @wontruefree

PR #826, opened 2022-03-08
support enums in cockroachdb

wontruefree:


matthewmcgarvey:
I would like to see an enum added or an enum modified (only if there's more than one so that Int32 and Int64 are covered) in specs. Just the process of adding it should reveal any edge cases we might have missed

jwoertink (changes requested):
We should for sure add a spec to make sure this doesn't break anything. I'm thinking maybe we can adjust https://github.com/luckyframework/avram/blob/main/spec/support/models/issue.cr so that way one of these is default, and the other is Int64 like maybe `enum Status : Int64`, and migration would have to be setup for Int64 https://github.com/luckyframework/avram/blob/abdfbd51acb1d69c2ca482d582c145ea6e9cf37d/db/migrations/20200324211956_create_issue.cr#L5

Then in theory, this file https://github.com/luckyframework/avram/blob/main/spec/avram/type_extensions/enum_spec.cr should just all work.

jwoertink:
Ok, so that theory worked... We can set the enum to Int64, and it seems to work.... Now the question is, do we have enough spec coverage to say with certainty that this won't break stuff? :joy:

jwoertink:
This is actually done in https://github.com/luckyframework/avram/issues/825 So next release will support enums with Int64. I'll close this out now. Thanks @wontruefree!

PR #828, opened 2022-03-12
Make all default times store in the DB with milliseconds down to 6point.

### LINKED ISSUE #727: Time columns don't include milliseconds

jwoertink:
Fixes #727

This makes all `Time` columns store with milliseconds. The tough part is if you're trying to do exact comparison with time, it's a bit harder now, but the benefit is if you need to sort by a time column, you'll get a more accurate sorting. This is especially common with UUID based tables that sort by `created_at`.

Also to note that out of the box default with postgres seems to use this same format for time columns:

```
development=# select now();
              now              
-------------------------------
 2022-03-12 18:02:14.760713+00
(1 row)
```

### DIFF
--- a/spec/avram/queryable_spec.cr
+++ b/spec/avram/queryable_spec.cr
@@ -1460,7 +1460,7 @@ describe Avram::Queryable do
       posts = Post::BaseQuery.new.published_at.between(start_date, end_date)
 
       posts.query.statement.should eq "SELECT posts.custom_id, posts.created_at, posts.updated_at, posts.title, posts.published_at FROM posts WHERE posts.published_at >= $1 AND posts.published_at <= $2"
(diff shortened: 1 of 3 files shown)


wontruefree:
does this work with all the time datatypes in postgres?

jwoertink:
Probably not with the `date` field, but we only map `Time` to `timestampz`, so for now that's the only one I'm looking at. I think for the other date/time column types, you'd probably have to hand roll some custom stuff.

PR #847, opened 2022-06-18
Array param support for operations

### LINKED ISSUE #408: Support for (virtual) array attributes

jwoertink:
Fixes #408

This PR adds in the ability to pass arrays through params in to operations. Prior to this, you could only set array values through named args, or manually by splitting a string or something like that.

Now you can do this:

```crystal
# post:tags[]=Crystal&post:tags[]=Lucky&post:title=Woot!&post:content=beepboop
SavePost.create(params) do |op, post|
  # ...
end
```

In doing this, there was quite a bit that needed some refactoring. One major breaking change is that `Avram::Params` now requires that all Hash values are `Array(String)` instead of `String`.

```crystal
# before
Avram::Params.new({"key" => "value"})

# after
Avram::Params.new({"key" => ["value"]})
```

-----------------------------------------------------------
Commit 15d7741 causes this error when I run specs:

This is the spec that causes the crash.
spec/avram/operations/define_attribute_spec.cr:227

```
Invalid memory access (signal 11) at address 0x18
[0x5590191bc336] *Exception::CallStack::print_backtrace:Nil +118 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559018f9fd4a] ~procProc(Int32, Pointer(LibC::SiginfoT), Pointer(Void), Nil) +330 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x7f9fb73dc520] ?? +140323950806304 in /lib/x86_64-linux-gnu/libc.so.6
[0x5590195a7717] *Avram::Params+ +39 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590197ae1df] */home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes +143 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590197aeed6] */home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes +22 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590197aeeb6] */home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes#set_attributes<Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing, Avram::Nothing>:Nil +6 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590197b5121] */home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes#initialize<Avram::UploadParams>:Nil +257 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590197b500d] */home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes::new<Avram::UploadParams>:/home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes +125 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019864c4e] */home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::upload_save_operation<Hash(String, Avram::UploadedFile)>:/home/jeremy/Development/lucky_org/avram/spec/avram/operations/define_attribute_spec.cr::SaveOperationWithAttributes +14 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590191427e9] ~procProc(Nil) +89 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019408864] *Spec::Example#internal_run<Time::Span, Proc(Nil)>:(Array(Spec::Result) | Nil) +244 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559018fa294e] ~procProc(Nil) +30 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x55901940b121] *Spec::Example::Procsy#run:Nil +49 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559018fa29d2] ~procProc(Nil) +98 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x55901940b121] *Spec::Example::Procsy#run:Nil +49 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019613b08] *Avram::SpecHelper::wrap_spec_in_transaction<Spec::Example::Procsy, TestDatabase.class>:Nil +568 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x55901911213e] ~procProc(Spec::Example::Procsy, Nil) +78 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x55901940af7a] *Spec::Context+ +442 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x55901940adb6] *Spec::Context+ +118 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x55901940ad2d] *Spec::Context+ +77 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590194071b2] *Spec::ExampleGroup#run_around_each_hooks<Spec::Example::Procsy>:Bool +290 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590194086a8] *Spec::Example#run:Nil +968 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019407769] *Spec::ExampleGroup +137 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019407616] *Spec::ExampleGroup#run:Nil +422 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590193b7e5b] *Spec::RootContext +187 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x5590193b7d96] *Spec::RootContext#run:Nil +6 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559018fa2659] ~procProc(Int32, (Exception | Nil), Nil) +57 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019351e16] *Crystal::AtExitHandlers::run<Int32, (Exception+ | Nil)>:Int32 +166 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019ac4b23] *Crystal::exit<Int32, (Exception+ | Nil)>:Int32 +35 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559019ac4a8f] *Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32 +111 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x559018f92f46] main +6 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x7f9fb73c3fd0] ?? +140323950706640 in /lib/x86_64-linux-gnu/libc.so.6
[0x7f9fb73c407d] __libc_start_main +125 in /lib/x86_64-linux-gnu/libc.so.6
[0x559018f7ee85] _start +37 in /home/jeremy/.cache/crystal/crystal-run-spec.tmp
[0x0] ???
```


spec/avram/type_extensions/enum_spec.cr
```diff
@@ -1,6 +1,6 @@
 require "../../spec_helper"
 
-include LazyLoadHelpers
```
jwoertink, line 3: This wasn't being used.

spec/support/upload_params.cr
```diff
@@ -4,6 +4,7 @@ class Avram::UploadParams < Avram::Params
   @uploads : Hash(String, Avram::UploadedFile) = {} of String => Avram::UploadedFile
 
   def initialize(@uploads)
+    @hash = {} of String => Array(String)
```
jwoertink, line 7: The specs wouldn't compile without this... How it was working before is magic....

src/avram/add_column_attributes.cr
```diff
@@ -0,0 +1,125 @@
+# This module creates methods for each column in a model that map
+# to an `Avram::Attribute` as well as methods that fill those attributes
+# with values that comes from params.
+module Avram::AddColumnAttributes
```
jwoertink, line 4: This code was in both SaveOperation and DeleteOperation, but seemed to have a few slight differences. One being that it was generating a `permitted_params` method for every column, but that method never did anything at the compiler level. So this becomes a small refactor since I had to change how things are checked.

src/avram/add_column_attributes.cr
```diff
@@ -0,0 +1,125 @@
...
+          {% end %}
+        {% end %}
+      end
+    end
+
+    def permitted_params
+      single_values = @params.nested(self.class.param_key).reject {|k,v| k.ends_with?("[]")}
+      array_values = @params.nested_arrays?(self.class.param_key) || {} of String => Array(String)
+      new_params = single_values.merge(array_values)
+      new_params.select(@@permitted_param_keys)
```
jwoertink, line 27: This will use some autocast magic which becomes `Hash(String, Array(String) | String)`. Maybe I should just add the return type?

src/avram/add_column_attributes.cr
```diff
@@ -0,0 +1,125 @@
...
+      {% COLUMN_ATTRIBUTES << attribute %}
+    {% end %}
+
+    private def extract_changes_from_params
+      permitted_params.each do |key, value|
+        {% for attribute in attributes %}
+          {% if attribute[:type].is_a?(Generic) %}
+            set_{{ attribute[:name] }}_from_param(value.as(Array(String))) if key == {{ attribute[:name].stringify }}
+          {% else %}
+            set_{{ attribute[:name] }}_from_param(value.as(String)) if key == {{ attribute[:name].stringify }}
```
jwoertink, line 17: If we don't cast these here, then `typeof(value)` is `Array(String) | String`, and the compiler starts to get confused

src/avram/add_column_attributes.cr
```diff
@@ -0,0 +1,125 @@
...
+        end
+
+        parse_result = {{ attribute[:type] }}.adapter.parse(_value)
+
+        if parse_result.is_a? Avram::Type::SuccessfulCast
+          {{ attribute[:name] }}.value = parse_result.value.as({{ attribute[:type] }})
+        else
+          {{ attribute[:name] }}.add_error "is invalid"
+        end
+      end
```
jwoertink, line 99: This was just 1 method before that did a quick check if the type was Generic or not before parsing. However, I think there was a bug doing it that way. We call `if _value.blank?` because if it's an empty string, treat it as `nil`. But in the case of an empty array, that could be a very valid value, and shouldn't be treated as nil.

Now, if we still want that same functionality (treating `[]` as `nil`), then I can add that back in, but I think that may have been a mistake.

spec/lucky/ext/avram/operation_params_spec.cr
```diff
@@ -0,0 +1,67 @@
...
+
+  param_key :beep
+end
+
+private class AMixedBag < Bucket::SaveOperation
+  permit_columns :names
+  attribute brand : String
+end
+
+describe "OperationParams" do
```
jwoertink, line 24: This spec file is intended to utilize Lucky directly now that Lucky is a dependency of Avram. We should be able to test real legit params instead of stubbing out params all over.

PR #859, opened 2022-07-20
Use actual names for Generics instead of T

### LINKED ISSUE #836: Using the generic name T in an operation name will case after_save to fail silently
The `T` generic silently clashes with the one used in `after_save`.

So this does not work:

```crystal
class SaveSomeModel(T) < SomeModel::SaveOperation
  after_save do |some_model|
    # use T...
  end
end
```

But this does:

```crystal
class SaveSomeModel(Some) < SomeModel::SaveOperation
  after_save do |some_model|
    # use Some...
  end
end
```

As proposed in the Discord, it would be better not never use `T` inside the framework, but use something like `LuckyT` instead.

jwoertink:
Fixes #836

There's a ton of places where `T` gets used, so I'll have to take some time to think out if all of them are necessary to change, or if just the ones where named collisions can happen....

### DIFF
--- a/src/avram/database_validations.cr
+++ b/src/avram/database_validations.cr
@@ -1,6 +1,6 @@
 require "./validations/**"
 
-module Avram::DatabaseValidations(T)
+module Avram::DatabaseValidations(AvramModel)
   # Validates that the given attribute is unique in the database
   # All validation methods return `Bool`. `false` if any error is added, otherwise `true`
   #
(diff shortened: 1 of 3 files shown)


matthewmcgarvey (commented):
I'm not so sure about `AvramModel` because it hides that it's a generic.

Is it possible to use `_T` or `_Avram_T` or something like that?

robacarp (approved):
I have no problem with this. I don't think it's opaque to call it AvramModel, thats what it would look like if it were a type alias, and that's essentially what this is.

PR #860, opened 2022-07-20
Adding support for bytea columns with Bytes

### LINKED ISSUE #762: Adding the byte array data type
An app I'm building is using two databases. The first one is an internal database for authentication and aggregated data. The other is provided by [cardano-db-sync](https://github.com/input-output-hk/cardano-db-sync), which is a component that syncs the Cardano blockchain into a Postgres database.

In their schema, they're using the Postgres' `bytea` datatype to store block and transaction hashes. I've been able to make this work with Avram using the following modifications.

First, there's the adapter for `Bytes`, which is an alias of `Slice(UInt8)`:

```cr
struct Slice(T)
  def self.adapter
    Lucky
  end

  module Lucky
    alias ColumnType = Bytes
    include Avram::Type

    def self.criteria(query : T, column) forall T
      Criteria(T, Bytes).new(query, column)
    end

    def from_db!(value : Bytes)
      value
    end

    def parse(value : Bytes)
      SuccessfulCast(Bytes).new(value)
    end

    def to_db(value : Bytes)
      value
    end

    class Criteria(T, V) < String::Lucky::Criteria(T, V)
    end
  end
end
```
I'm not really familiar with the `Criteria` object in the adapter, so maybe I'm overlooking something there.

Next, there's this modification to allow querying with an instance of `Bytes`:

```cr
abstract class Avram::Where::ValueHoldingSqlClause < Avram::Where::SqlClause
  getter value : String | Array(String) | Array(Int32) | Slice(UInt8)
end
```

So now this works as expected:

```cr
class Cardano::Block < Cardano::BaseModel
  table :block do
    # ...
    column hash : Bytes
    # ...
  end
end
```

One thing I worry about is that the adapter is added specifically for `Slice(UInt8)`, but the type is `Slice(T)`. Not sure if that's going to cause trouble in the future, or not.

I'm wondering now, is it worthwhile adding this data type to Avram? In my case, I have a read-only database, so I'm good with the two modifications. If we were to add it as an officially supported type, this should also work with operations and migrations.

If you think it's a good addition, I'll gladly prepare a PR. 🤓️

jwoertink:
Fixes #762

Questions I need an answer for:

* How do you query for this in raw postgres? Postgres will let you query by the original string, or a hexbytes representation. I couldn't get the hexbytes query to work, but regular bytes and the raw string both work.
* What does an insert SQL with this column type look like? `values ('raw string')` or `values ('\xabc123')`. 
* How would you use this in a HTML form or API? I assume you'd pass a String and then call `to_slice`?  You can just use a normal text field and insert a string like a normal String field. Postgres will handle the conversion under the hood.

### DIFF
--- a/spec/avram/type_extensions/bytes_spec.cr
+++ b/spec/avram/type_extensions/bytes_spec.cr
@@ -11,4 +11,16 @@ describe "Bytes" do
     result.should be_a(Avram::Type::SuccessfulCast(Bytes))
     result.value.should eq(Bytes[10, 10, 10])
   end
+
+  describe "blank?" do
+    it "returns true" do
+      result = Bytes.adapter.parse(Bytes.empty)
+      result.value.blank?.should eq(true)
+    end
+
+    it "returns false" do
+      result = Bytes.adapter.parse("100".to_slice)
+      result.value.blank?.should eq(false)
+    end
+  end
 end
--- a/spec/avram/type_extensions/bytes_spec.cr
+++ b/spec/avram/type_extensions/bytes_spec.cr
@@ -3,8 +3,12 @@ require "../../spec_helper"
 describe "Bytes" do
   it "parses Bytes from String" do
     result = Bytes.adapter.parse("10")
-    result.value.should eq(10)
+    result.value.should eq(Bytes[49, 48])
   end
 
-
+  it "returns SuccessfulCast for Bytes" do
+    result = Bytes.adapter.parse(Bytes[10, 10, 10])
+    result.should be_a(Avram::Type::SuccessfulCast(Bytes))
+    result.value.should eq(Bytes[10, 10, 10])
+  end
 end
--- a/spec/avram/type_extensions/bytes_spec.cr
+++ b/spec/avram/type_extensions/bytes_spec.cr
@@ -1,0 +1,10 @@
+require "../../spec_helper"
+
+describe "Bytes" do
+  it "parses Bytes from String" do
+    result = Bytes.adapter.parse("10")
+    result.value.should eq(10)
+  end
+
+
+end
--- a/db/migrations/20220720013254_create_beats.cr
+++ b/db/migrations/20220720013254_create_beats.cr
@@ -1,0 +1,13 @@
+class CreateBeats::V20220720013254 < Avram::Migrator::Migration::V1
+  def migrate
+    create table_for(Beat) do
+      primary_key id : Int64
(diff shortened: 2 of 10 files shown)


spec/avram/type_extensions/bytes_spec.cr
```diff
@@ -0,0 +1,10 @@
+require "../../spec_helper"
+
+describe "Bytes" do
+  it "parses Bytes from String" do
+    result = Bytes.adapter.parse("10")
+    result.value.should eq(10)
```
hugopl, line 6: Looking at the FLoat64 and Int32 specs, should not the expected result be `"10".to_slice` instead of the integer 10?

Disconsider if the comment doesn't make sense since I didn't dig into Avram internals yet

jwoertink: Oh, I just copied the spec. Since it doesn't compile yet, I haven't ran it, so I have no idea what to actually expect here. I've never used Bytes for anything (or a bytea column), so I'm not sure what a `String` -> `Bytes` conversion looks like :sweat_smile:

mdwagner:
Is the goal to work with both unescaped and escaped bytes implementations in Postgre? Or just the unescaped? Unescaped is the preferred, according to postgres docs.

mdwagner:
Also, you don't _have_ to call to_slice method on a string, there's also the bytes method. Either will work here, since we're not passing any pointers down to C.

jwoertink:
> Is the goal to work with both unescaped and escaped bytes implementations in Postgres?

I have no clue. I guess whatever provides the biggest most common support? @hugopl and @wout may be able to answer this. I just wanted to get the PR moving so I can get this in for the next release.

wout:
Unfortunately, I had to give up on this route. After looking at the bigger picture for my goal, which is building infrastructure on Cardano, I had to port a massive amount of code to Crystal. Code that's readily available in Haskell and JS. Because that would be an overwhelming task to write and maintain for me alone, I decided to continue in Haskell.

hugopl:
My use case was far simple, I just wanted the data on a `bytea` column and I didn't need to use this column in any `WHERE` statements, so the first version of this PR that failed to pass on specs already worked for me :sweat_smile: 

Partial support is better than no support, and things can always be improved later.

jwoertink:
Ok, I got both query and save operations working! There's some specs and stuff, but I think this is decent enough for a start now.

Issue #857, opened 2022-07-19
Add custom date critera query methods


jwoertink:
Postgres has some cool date functions like `EXTRACT()` when working with times 

https://www.postgresqltutorial.com/postgresql-date-functions/postgresql-extract/

It would be cool if we had a type-safe way to query with these. Maybe it looks something like...

```crystal
# SELECT * FROM users WHERE EXTACT(DAY FROM birthdate) = ? AND EXTRACT(MONTH FROM birthday) = ?
UserQuery.new.birthdate.extract_day.eq(Time.utc.day).birthdate.extract_month.eq(Time.utc.month)
```

jwoertink:
Closed by #861

PR #861, opened 2022-07-29
Add extract functionality as a new criteria to timestamptz columns

jinn999:
This PR concerns the extract info from date mentioned in issue #857.

It provides a way to be able to extract specific fields from a `timestamptz` column and compare it with some numerical values.
For all the fields implemented till now (see `Avram::ChronoUnits` in the `extract_criteria.cr` or the relevant postgresql documentation), three possible methods are provided:

- `column.extract(Avram::ChronoUnits::<enum_value>)`
             e.g: `some_table.created_on.extract(Avram::ChronoUnits::Day)`
- `column.extract(<enum_value_as_symbol>)`
             e.g: `some_table.created_on.extract(:day)`
- `column.extract_<enum_value>`
             e.g: `some_table.created_on.extract_day`

Functionality can be used as:
- column.extract_year.eq(2022)

Most values returned are ints, but 4 of them which return floats
- julian
- epoch
- second
- milliseconds

I am still not able to do the whole script/setup script/test stuff due to the M1 usual issue, but relevant specs pass..

### DIFF
--- a/src/avram/criteria_extensions/extract_criteria.cr
+++ b/src/avram/criteria_extensions/extract_criteria.cr
@@ -1,0 +1,132 @@
+module Avram
+  enum ChronoUnits
+    Century
+    Day
+    Decade
+    Dow
+    Doy
+    Epoch
+    Hour
+    Isodow
+    Isoyear
+    Julian
+    Microseconds
+    Millennium
+    Milliseconds
+    Minute
+    Month
+    Quarter
+    Second
+    Timezone
+    Timezone_hour
+    Timezone_minute
+    Week
+    Year
+  end
+end
+
+module Avram::ExtractCriteria
+  macro included
+    def extract(chrono_unit : Avram::ChronoUnits)
+        applied_operation = "extract(#{chrono_unit.to_s.downcase} from #{@column})"
+        case chrono_unit
+        when .julian?, .second?, .milliseconds?, .epoch?
+            Criteria(T,Float64).new(rows, applied_operation)
+        else Criteria(T,Int32).new(rows, applied_operation)
+        end
+    end
+
+    def extract(symbol : Symbol)
+        extract(Avram::ChronoUnits.parse(symbol.to_s))
+    end
+
+    def extract_day
+        extract(Avram::ChronoUnits::Day)
+    end
+
+    def extract_month
+        extract(Avram::ChronoUnits::Month)
+    end
+
+    def extract_century
+        extract(Avram::ChronoUnits::Century)
+    end
+
+    def extract_decade
+        extract(Avram::ChronoUnits::Decade)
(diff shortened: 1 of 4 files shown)


src/avram/criteria_extensions/extract_criteria.cr
```diff
@@ -0,0 +1,132 @@
...
+    Microseconds
+    Millennium
+    Milliseconds
+    Minute
+    Month
+    Quarter
+    Second
+    Timezone
+    Timezone_hour
+    Timezone_minute
```
jwoertink, line 22: Curious about these. Is this because of the Symbol arg override? I'm assuming doing TimezoneHour then passing :timezone_hour fails?

src/avram/criteria_extensions/extract_criteria.cr
```diff
@@ -0,0 +1,132 @@
...
+    def extract_week
+        extract(Avram::ChronoUnits::Week)
+    end
+
+    def extract_year
+        extract(Avram::ChronoUnits::Year)
+    end
+
+end
+end
```
jwoertink, line 132: I can't tell if this is formatting or the GitHub mobile app 😅

jwoertink (commented):
This is amazing! Thanks so much for adding this in. A few small comments but I think it's great 😁

src/avram/criteria_extensions/extract_criteria.cr
jinn999: Now, this is funny. it was not the mobile app, it was formatting, but the funny thing is that neither tool format nor ameba had anything to complain :/

jinn999:
Ok, did a major revamp to avoid having all these extract_xxx methods to define and risking forgetting some of them. Switched to heavy macroing, and moved chronounits in their own file

src/avram/criteria_extensions/extract_criteria.cr
jinn999: Actually, it was a complete mishap from my part, dunno why but I was convinced that they were with no underscore :/ either way, I changed them to be correct now :) which also made me learn about camelcase and underscore methods XD

src/avram/criteria_extensions/extract_criteria.cr
jwoertink: That's strange. You'd think the formatter would have caught that 🤔

jinn999: Yup, pretty strange... But also the checks you have in the pipeline would have caught that :/ bug in the crystal tool format? Dunno :/

jwoertink (approved):
:+1:  Thanks!

paulcsmith:
This is super cool. Thanks @jinn999

PR #862, opened 2022-07-31
Allows named args to take precendence over param values.

### LINKED ISSUE #764: Values from named args should take precedence over values from params
When you're using a SaveOperation, there might be a time where you need to manipulate a value coming from params. Maybe a date, or CSV input or something...

Right now, if you pass both param value and named arg of the same field, the param value takes precedence. Since named args are controlled by you, it would make sense that this is the value you actually want.

```crystal
  permit_columns username
end

# user:username=ZeroCool
SaveUser.create(params, username: "CrashOverride") do |_,_|
  # user should be CrashOverride
end
```

jwoertink:
Fixes #764

This flips extracting param values first before setting named args. There could be a case where you might have a "admin" field that's technically permitted, but you want the initial value to be `false`. So you set a named arg value to be false. Currently, if a user updated the params to manually push that through, it would be set to what they set it to. Now it'll be set to your named arg value.

### DIFF
--- a/spec/avram/operations/save_operation_spec.cr
+++ b/spec/avram/operations/save_operation_spec.cr
@@ -619,6 +619,25 @@ describe "Avram::SaveOperation" do
         record.greeting.should eq "Hi"
         record.admin.should eq false
       end
+
+      it "lets named args take precedence over param values" do
+        params = build_params("model_with_default_values:greeting=Hi")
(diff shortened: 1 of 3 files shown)


matthewmcgarvey (approved):

paulcsmith:
Great idea 👍

PR #866, opened 2022-08-07
Factory callbacks now return self

### LINKED ISSUE #865: Should Factory callbacks return self instead of Nil?
https://github.com/luckyframework/avram/blob/e705cb3897dec614bd5331526c10d42c6339003b/src/avram/factory.cr#L41-L47

It would make sense that you might not run code that looks like

```crystal
before_save { do_thing }.some_other_code
```

However, I'm finding this to be a common pattern, and could eventually become something we add in:

```crystal
class UserFactory < Avram::Factory
  def initialize
    before_save do
      some_special_setup
    end
  end

  def with_friends
    after_save do |user|
       FriendFactory.create &.user(user)
     end
    # this must return `self` here otherwise it won't compile when calling this method
    self
  end
end

UserFactory.create

UserFactory.create(&.with_friends)
```

Without return `self` in the instance methods, any calls passed through the `create` block will raise the compile error

> yield(new).create
> Undefined method create for Nil

If anyone can think of why this might be bad, let me know.

jwoertink:
Fixes #865

You can now use these like this:

```crystal
class UserFactory < Avram::Factory
  def with_friends
    after_save do |user|
      FriendFactory.create &.user_id(user.id)
    end
  end
end

UserFactory.create &.with_friends
```

### DIFF
--- a/spec/avram/factory_spec.cr
+++ b/spec/avram/factory_spec.cr
@@ -70,6 +70,13 @@ describe Avram::Factory do
       scan = factory.create
       scan.line_item_id.should eq line_item_id
     end
+
+    it "returns self and can be chaind" do
+      factory = ScanFactory.new
+      line_item_id = LineItemFactory.create.id
+      scan = factory.before_save { factory.line_item_id(line_item_id) }.create
+      scan.line_item_id.should eq line_item_id
+    end
   end
 
   describe "after_save" do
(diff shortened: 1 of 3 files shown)


matthewmcgarvey (approved):

paulcsmith:
This is awesome!

PR #878, opened 2022-08-26
Don't wrap nil in JSON::Any on nilable columns

### LINKED ISSUE #877: Nilable JSON columns don't actually return nil
```crystal
table do
  column metadata : JSON::Any?
end

pp model.metadata #=> nil
pp model.metadata.nil? #=> false
pp typeof(metadata) #=> JSON::Any
```

It seems that it returns `JSON::Any.new(nil)`, so it's more of `null` json than it is nil.

matthewmcgarvey:
Fixes #877

This hasn't really caused issues for people because `JSON::Any` wrapping `nil` still returns true for `.nil?` which seems weird.


jwoertink (approved):
Nice!

PR #879, opened 2022-09-10
this update was missed in #868

jwoertink:
Right now the LuckyCLI is tanking on this class like it did before... I'm hoping it's because this was missed....

Issue #819, opened 2022-03-04
Errors should have the variables as properties


jwoertink:
https://github.com/luckyframework/avram/blob/9251a965372d319977558bcbe3d0696f60aeb45e/src/avram/errors.cr#L23-L31

You should be able to access these locals as properties so that way if you need to customize this error, you can.

This is just one class, but there's a few that could benefit here as well

PR #880, opened 2022-09-17
Fix compilation issue with InvalidOperationError

matthewmcgarvey:
Connected to:
- https://github.com/luckyframework/avram/pull/879
- https://github.com/luckyframework/avram/pull/868
- https://github.com/luckyframework/avram/issues/819

Before this change, I was getting one of those "you've found a bug in the Crystal compiler errors" with this as the first line

`BUG: trying to downcast Avram::InvalidOperationError(AvramOperationType) (Crystal::GenericClassType) <- Exception+ (Crystal::VirtualType) (Exception)`

Something about the generics just didn't seem to work, so I found a solution that works without generics. It's not ideal since there's very little on the `Avram::OperationErrors` interface, I just don't know of another way to solve this.

### DIFF
--- a/src/avram/errors.cr
+++ b/src/avram/errors.cr
@@ -56,11 +56,11 @@ module Avram
   end
 
   # Raised when using the create!, update!, or delete! methods on an operation when it does not have the proper attributes
-  class InvalidOperationError(AvramOperationType) < AvramError
+  class InvalidOperationError < AvramError
     getter errors : Hash(Symbol, Array(String))
-    getter operation : AvramOperationType
+    getter operation : OperationErrors
 
-    def initialize(@operation : AvramOperationType)
+    def initialize(@operation)
       message = String.build do |string|
         string << "Could not perform #{operation.class.name}."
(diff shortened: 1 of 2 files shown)


jwoertink:
Actually, this is fixed by https://github.com/luckyframework/lucky_cli/pull/777/files#diff-c0153183856c27bd5dc920235cd6b3e125cfeac7ff8abf50ba03e31c19b9eba6R32 

When the `InvalidOperation` is raised, it actually passes in the SaveOperation

https://github.com/luckyframework/avram/blob/main/src/avram/save_operation.cr#L240

so the Type for `operation` would need to be `Avram::SaveOperation | Avram::Operation`, but if we just pass that over in the Errors action, it actually fixes the Crystal issue.

Related: https://github.com/crystal-lang/crystal/issues/12472

matthewmcgarvey:
O ok! I tried updating places in Avram. Didn't think about looking in my own project, and that change did work for me 👍

PR #887, opened 2022-10-07
always pass record to block in a delete operation

### LINKED ISSUE #886: Always pass record to block in DeleteOperation.delete
Right now, DeleteOperation conditionally passes the record to the block

https://github.com/luckyframework/avram/blob/9c8caecf052641d15dbe027e0cafb1741d8af33f/src/avram/needy_initializer_and_delete_methods.cr#L127-L131

Unless there's a good reason for that, it seems more ✨ _ergonomic_ ✨ to always pass in the record.

jcads:
Closes #886

### DIFF
--- a/src/avram/needy_initializer_and_delete_methods.cr
+++ b/src/avram/needy_initializer_and_delete_methods.cr
@@ -125,7 +125,12 @@ module Avram::NeedyInitializerAndDeleteMethods
         operation.delete!
       {% else %}
         operation.delete
-        yield operation, operation.record.not_nil!
+        case operation.delete_status
+        when :DeleteFailed
+          yield operation, nil
+        when :Deleted
+          yield operation, operation.record
+        end
       {% end %}
     end
   end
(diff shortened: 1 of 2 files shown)


src/avram/needy_initializer_and_delete_methods.cr
```diff
@@ -124,11 +124,8 @@ module Avram::NeedyInitializerAndDeleteMethods
       {% if with_bang %}
         operation.delete!
       {% else %}
-        if operation.delete
-          yield operation, operation.record.not_nil!
-        else
-          yield operation, nil
-        end
+        operation.delete
+        yield operation, operation.record.not_nil!
```
matthewmcgarvey, line 128:
```suggestion
        yield operation, operation.record
```

I don't believe you even need `not_nil!`. It doesn't look like it's defined optionally in `DeleteOperation`. Could be wrong, though.

matthewmcgarvey (approved):

PR #895, opened 2022-10-14
Fix error --  cast from Array(String) to String

### LINKED ISSUE #884: cast from Array(String) to String failed from Avram::Params
In https://github.com/luckyframework/avram/pull/847 I had to change `Avram::Params` to take `Hash(String, Array(String))` instead of `Hash(String, String)` so that we can allow array params to be passed in. This also aligns with `URI::Params`.

I wasn't able to make it a union because this caused a LOT of compiler headache trying to determine which type a value was, and constantly having to cast it.

To use this, you would do

```crystal
params = Avram::Params({"username" => ["tacoguy"]})
SaveUser.create(params) do |_,_|
end
```

As it turns out, the specs never caught this because in the specs, we just inherit from Avram::Params

https://github.com/luckyframework/avram/blob/9c8caecf052641d15dbe027e0cafb1741d8af33f/spec/avram/params_spec.cr#L21-L37

The error actually borks on https://github.com/luckyframework/avram/blob/9c8caecf052641d15dbe027e0cafb1741d8af33f/src/avram/add_column_attributes.cr#L17

because in this case, `value` is `Array(String)` even though the column should just be `String`. The tricky part is `typeof(value)` here is  actually `Array(String) | String` which ends up coming back to the original union issue. We need some sort of way to say that if value gets to this branch, and value is still an array, it needs to call `.first`, but if it's not an array, then make sure to cast as a String.

akadusei:
Fixes #884

You can now instantiate `Avram::Params` passing a hash with array values or string values or a mixture of both.

### DIFF
--- a/src/avram/params.cr
+++ b/src/avram/params.cr
@@ -1,10 +1,12 @@
 class Avram::Params
   include Avram::Paramable
 
-  @hash : Hash(String, Array(String))
+  @hash : Hash(String, Array(String) | String) | \
+    Hash(String, Array(String)) | \
+    Hash(String, String)
 
   def initialize
-    @hash = {} of String => Array(String)
+    @hash = {} of String => String
   end
 
   def initialize(@hash)
@@ -15,9 +17,9 @@ class Avram::Params
   end
 
   def nested(key : String) : Hash(String, String)
-    Hash(String, String).new.tap do |h|
-      @hash.each do |k, v|
-        h[k] = v.first?.to_s
+    Hash(String, String).new.tap do |params|
+      @hash.each do |_key, _value|
+        params[_key] = _value if _value.is_a?(String)
       end
     end
   end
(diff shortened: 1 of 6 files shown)


src/avram/params.cr
```diff
@@ -1,10 +1,12 @@
 class Avram::Params
   include Avram::Paramable
 
-  @hash : Hash(String, Array(String))
+  @hash : Hash(String, Array(String) | String) |  \
+    Hash(String, Array(String)) |  \
+    Hash(String, String)
```
jwoertink, line 6: oh, interesting! I think this is where I got tripped up before. I didn't think to also do a union of each one. Nice 👍

jwoertink (commented):
Thanks for taking this on. One thing I'd like to see would be a spec that allows me to pass in an array value, and then one with a mixed value. So like...

```crystal
it "should work" do
  params = Avram::Params.new({"key" => ["val1", "val2"]})
  SomeOperation.create(params) do |o, v|
    v.should_not be_nil
  end
end

it "should also work" do
  params = Avram::Params.new({"key" => ["val1", "val2"], "key2" => "123"})
  SomeOperation.create(params) do |o, v|
    v.should_not be_nil
  end
end
```

I think having those two test cases, we should be solid enough to know this all works well.

jwoertink (approved):
Sweet! I think this looks good.

matthewmcgarvey (approved):

PR #900, opened 2022-10-20
Add support for CockrochDB query return

wontruefree:
CockrochDB returns Int64 when checking for email. This fixes that bug.

### DIFF
--- a/src/avram/queryable.cr
+++ b/src/avram/queryable.cr
@@ -221,7 +221,7 @@ module Avram::Queryable(T)
     cache_store.fetch(cache_key(:any?), as: Bool) do
       queryable = clone
       new_query = queryable.query.limit(1).select("1 AS one")
-      results = database.query_one?(new_query.statement, args: new_query.args, queryable: schema_class.name, as: Int32)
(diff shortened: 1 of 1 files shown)


src/avram/queryable.cr
```diff
@@ -221,7 +221,7 @@ module Avram::Queryable(T)
     cache_store.fetch(cache_key(:any?), as: Bool) do
       queryable = clone
       new_query = queryable.query.limit(1).select("1 AS one")
-      results = database.query_one?(new_query.statement, args: new_query.args, queryable: schema_class.name, as: Int32)
+      results = database.query_one?(new_query.statement, args: new_query.args, queryable: schema_class.name, as: (Int32 | Int64))
```
jwoertink, line 224: I think we can just use `as: Int64` here. The actual return value is irrelevant. We only really care if there is a value.

wontruefree: because of other databases this needs to be Int32 and Int64.

matthewmcgarvey (approved):

jwoertink (approved):

PR #901, opened 2022-10-21
Avoid re-fetching associations that have already been preloaded

matthewmcgarvey:
This updates the class-level preload functions to check if the association has already been loaded before redoing the process. Now you don't have to worry about making unnecessary SQL queries when you don't have to.

**TODO** Support has many through passing multiple records.

Honestly, has many through has been the most difficult part of the codebase **by far** and this is no different. The problem is when the intermediary association is loaded but not the one in the has many through. Since we pass that we want the has many through association preloaded in the query that loads the intermediary, if it is loaded that call to preload is completely ignored.

The fix is to rethink how we make that query. Instead of loading the intermediaries, just set up fancy joins to only query for the has many through association. But how do we do that? 🤷 

Also, this preload stuff is very broken for a has many through with more than one intermediary (a through list with more than 2 items)

I figure that this is good enough to include without that handling for the has many through. This stuff should be mostly invisible to end-users, anyways.

```crystal
user = UserQuery.find(123)
user = UserQuery.preload_subscription(user) # <--- Query happens here to load the subscription
user = UserQuery.preload_subscription(user) # <--- User already has subscription loaded, nothing happens
user = UserQuery.preload_subscription(user, force: true) # <--- Force query to refetch association
user = UserQuery.preload_subscription(user, &.renewal("monthly") # <--- The actual query we pass in doesn't matter, we still don't fetch the subscription again
```

### DIFF
--- a/db/migrations/20170127143151_create_sign_in_credentials.cr
+++ b/db/migrations/20170127143151_create_sign_in_credentials.cr
@@ -4,6 +4,7 @@ class CreateSignInCredentials::V20170127143151 < Avram::Migrator::Migration::V1
       primary_key id : Int64
       add_timestamps
       add user_id : Int64
+      add value : String
     end
   end
 
--- a/spec/avram/preloading/preloading_belongs_to_spec.cr
+++ b/spec/avram/preloading/preloading_belongs_to_spec.cr
@@ -153,5 +153,61 @@ describe "Preloading belongs_to associations" do
         end
       end
     end
+
+    it "does not refetch association from database if already loaded (even if association has changed)" do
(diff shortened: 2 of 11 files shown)


jwoertink (approved):
This was a bit tough to follow, but I didn't see anything glaring. This isn't a breaking change right?

matthewmcgarvey:
@jwoertink no, it should not be a breaking change. I added the `force` param but it has a default value and I maintained ordering of the overloads that take in an array

matthewmcgarvey:
I'll look over the code one more time and merge later today/tomorrow

Issue #893, opened 2022-10-10
Confusing error message when enum columns encounter unexpected data


robacarp:
If you have an enum column:

```crystal
class Model < BaseModel
  enum Status
    Unknown = -1
    Ok = 0
    NotOk = 1
  end

  table :whatever do
    column status_code : Model::Status = Model::Status::Unknown
  end
end
```

And you pull data from the database which has `status=2`, you'll get this runtime error:

cast from Avram::Type::FailedCast to Avram::Type::SuccessfulCast(T) failed, at /app/lib/avram/src/avram/type.cr:25:5:25

Which is not at all helpful. What's a better way to raise and inform developers about this data corruption?

PR #903, opened 2022-10-25
Talking errors in casting failures

jinn999:
Implements #893. There is a breaking change, though. Previously, calling FailedCast#value would always return a nil value, without providing any clue on which value was actually breaking.
Now, it raises a FailCastError with a talking error message, in the form of - for instance - "Value ->Invalid<- is not a valid constant for enum WhateverEnum".

Btw,  is there an "easy" way to test avram modifications in a dummy lucky app?

### DIFF
--- a/spec/avram/queryable_spec.cr
+++ b/spec/avram/queryable_spec.cr
@@ -374,7 +374,7 @@ describe Avram::Queryable do
     end
 
     it "raises PQ::PQError if no record is found with letter-only id (String)" do
-      expect_raises(Exception, "FailedCast") do
+      expect_raises(Avram::FailedCastError) do
         UserQuery.find("id")
       end
     end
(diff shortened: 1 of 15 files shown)


jwoertink:
Hmm... I'm not sure about this one :thinking:  The `FailedCast` was never intended to be an error. It's meant to just be a type check that it didn't know how to convert. By setting these values to `nil`, then your operation can handle the errors as you need. With this raising now, then you would never get to your operation in a full Lucky app. It would fail out before you had a chance to gather up all of the validation errors. And since exceptions are meant to be handled in your Lucky app's Error::Show action, this would end up breaking some of that convention.

I do see why you went this route though. I think I want to sit on this a little longer and see if there's an alternate way we can fix the issue without making some breaking changes here. Thanks for getting this started!

Issue #888, opened 2022-10-07
Upsert operations should tell you if a record was created or updated


jwoertink:
When you're using an upsert, you may want to know whether the record that comes back was an updated record or a new record.

```crystal
SaveGoal.upsert(title: "Some goal", amount: 100) do |operation, goal|
  # maybe something like this?
  operation.created? #=> false
  operation.updated? #=> true
end
```

This could really hook in to all SaveOperation too.

```crystal
SaveGoal.create(...) do |o, g|
  o.created? #=> true
  o.updated? #=> false
end

SaveGoal.update(...) do |o, g|
  o.created? #=> false
  o.updated? #=> true
end
```

We already have `operation.saved?` 

https://github.com/luckyframework/avram/blob/9c8caecf052641d15dbe027e0cafb1741d8af33f/src/avram/save_operation.cr#L27-L28

jwoertink:
One thing I forgot to consider on this.... What are the results when the save fails? I guess if you're calling `create` and it fails to create then `created?` just returns false, and same for `updated?` :thinking:

jwoertink:
This is now done.

PR #904, opened 2022-10-25
Add updated? and created? methods on SaveOperations

jinn999:
Implements #888.
Adds two new boolean methods to `SaveOperation`.
They work similarly to `new_record?` but always return false if the record was not persisted (e.g: failed save or not-yet-performed operation)

### DIFF
--- a/spec/avram/operations/save_operation_spec.cr
+++ b/spec/avram/operations/save_operation_spec.cr
@@ -220,7 +220,9 @@ describe "Avram::SaveOperation" do
           nickname: nil,
           age: 30,
           joined_at: joined_at
-        ) do |_operation, user|
+        ) do |operation, user|
+          operation.created?.should be_false
+          operation.updated?.should be_true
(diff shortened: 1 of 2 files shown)


jwoertink:
This is great! Thanks for doing this. I think the only part missing for it to complete #888 is that it needs to work on `upsert` as well. That was actually the one bit I really needed for this. But if you don't have time to add that, I'm willing to accept this as a partial implementation. I think it's already super useful.

jinn999:
I'm probably drunk or missing the point but... Doesn't it work for upsert as well? Have a look at the changed tests, but I'm pretty sure I added a check in there (iirc, it was a free-to-have in defining these two methods)

jwoertink:
Nope, I'm the drunk one! I totally missed that your specs cover upsert already :tada:  Thanks for pointing that out.

jwoertink (approved):

PR #908, opened 2022-11-19
Ensure matching column message when missing Enum

stan-slas:
First PR here :wave: 

### What
When we declare an Enum field in the models, it should be referenced as Int32 in the migration file (https://luckyframework.org/guides/database/models#using-enums).

The current message was instead suggesting to type it as Enum in the migration which doesn't work.

### How it has been tested
- Added one missing test for the missing column
- Added the specific UT for the Enum case

I didn't assert the full message for brevity of UT on hint messages. I thought it may become too verbose for a not critical function but feel free to request if needed.

Example of full message 
```
ModelWithMissingEnumColumn wants to use the column 'gender' but it does not exist.
Try adding the column to the table...
       
         ▸ Generate a migration:
       
             lucky gen.migration AddGenderToModelWithMissingEnumColumns
       
         ▸ Add the column to the migration:
       
             alter :users do
               # Add the column:
               add gender : Int32
       
               # Or if this is a column for a belongs_to relationship:
               add_belongs_to choices : Int32
             end
       
       Or, you can skip schema checks for this model:
       
           class ModelWithMissingEnumColumn < BaseModel
             # Great for models used in migrations, or for legacy schemas
             skip_schema_enforcer
           end
```

### DIFF
--- a/src/avram/schema_enforcer/ensure_matching_columns.cr
+++ b/src/avram/schema_enforcer/ensure_matching_columns.cr
@@ -49,6 +49,8 @@ class Avram::SchemaEnforcer::EnsureMatchingColumns < Avram::SchemaEnforcer::Vali
     if best_match
       message += " Did you mean '#{best_match.colorize.yellow.bold}'?\n\n"
     else
+      # Enum column are mapped to Int32 in the database layer
+      missing_attribute_type_hint = "#{missing_attribute[:name]} : #{missing_attribute[:type].sub("Enum", "Int32")}"
       message += <<-TEXT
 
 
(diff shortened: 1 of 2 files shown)


src/avram/schema_enforcer/ensure_matching_columns.cr
```diff
@@ -49,6 +49,8 @@ class Avram::SchemaEnforcer::EnsureMatchingColumns < Avram::SchemaEnforcer::Vali
     if best_match
       message += " Did you mean '#{best_match.colorize.yellow.bold}'?\n\n"
     else
+      # Enum column are mapped to Int32 in the database layer
+      missing_attribute_type_hint = "#{missing_attribute[:name]} : #{missing_attribute[:type].sub("Enum", "Int32")}"
```
jwoertink, line 53: This is nice, but it's possible to define an Enum as `Int64` as well. https://github.com/luckyframework/avram/pull/856

jwoertink (commented):
Thanks for the PR! I think this will help for sure, but I have a few questions. I'm not sure if it's always been the case, but `add_belongs_to choices : Int32` wouldn't ever make sense. This would assume `Int32` has a `id` method that would relate back to `choices_id`. We may want to think about just putting something more generic there. Any ideas?

Also Enums can technically be Int64 as well. Maybe there's a way we can detect if the Enum type is Int64?

src/avram/schema_enforcer/ensure_matching_columns.cr
jwoertink: Coming back to this, we may eventually want to add all of the `Int`. Like if you had a Colors Enum that you wanted to be `Int8` or maybe `Int16`. I'm not sure the best way to handle this just yet, but definitely having `Int32` static in here may be an issue.

PR #915, opened 2022-12-30
Adding in a spec that tests Lucky #1762

jwoertink:
@stephendolan Here's a spec setup for that issue you closed out... though, it seems to be failing for me :thinking:


jwoertink:
:man_facepalming:  my bad... it was because I didn't permit the column :grimacing:

PR #920, opened 2023-01-16
Fix `DB::ColumnTypeMismatchError` for `Int32` columns in CockroachDB

### LINKED ISSUE #919: Getting `DB::ColumnTypeMismatchError` for `Int32` columns in CockroachDB
I have an app that I'm currently testing on CockroachDB, and encounter this error:

```
In PG::ResultSet#read the column one returned a Int64 but a Int32 was expected. (DB::ColumnTypeMismatchError)
    from lib/db/src/db/result_set.cr:89:9 in 'read'
    from lib/db/src/db/query_methods.cr:202:9 in 'query_one?:as:args'
    from lib/avram/src/avram/database.cr:88:3 in 'query_one?:as:args:queryable'
    from lib/avram/src/avram/database.cr:88:3 in 'query_one?:args:queryable:as'
    from lib/avram/src/avram/queryable.cr:224:17 in 'any?'
    # ...
```

The problem appears to be that in CockroachDB (v22.2), `int` is an alias for `int8`. In PostgreSQL, `int` refers to `int4`. So for an `Int32` column in Avram, CockroachDB is creating an `int8` column whereas `crystal-pg` expects to read an `int4`.

Not tested, but this may be solved by changing the `Avram::Migrator::Columns::Int32Column#column_type` to `int4`:

https://github.com/luckyframework/avram/blob/ac949e463af90fd516e845e75744117c72ad4435/src/avram/migrator/columns/int32_column.cr#L11

While at it, we may as well change `Avram::Migrator::Columns::PrimaryKeys::Int32PrimaryKey#column_type` to `serial4`:

https://github.com/luckyframework/avram/blob/ac949e463af90fd516e845e75744117c72ad4435/src/avram/migrator/columns/primary_keys/int32_primary_key.cr#L9

These changes should make it explicit to both DB engines what the column type is. This should not be a breaking change.

akadusei:
Fixes #919

There's no specs, though; can't think of any.

### DIFF
--- a/spec/avram/migrator/alter_table_statement_spec.cr
+++ b/spec/avram/migrator/alter_table_statement_spec.cr
@@ -32,7 +32,7 @@ describe Avram::Migrator::AlterTableStatement do
       ADD name text,
       ADD email text,
       ADD nickname text NOT NULL,
-      ADD age int NOT NULL DEFAULT '1',
+      ADD age int4 NOT NULL DEFAULT '1',
       ADD num bigint NOT NULL DEFAULT '1',
       ADD amount_paid decimal(10,5) NOT NULL DEFAULT '1.0',
       ADD completed boolean NOT NULL DEFAULT 'false',
@@ -41,7 +41,7 @@ describe Avram::Migrator::AlterTableStatement do
       ADD updated_at timestamptz,
       ADD future_time timestamptz NOT NULL DEFAULT '#{Time.local.to_utc}',
       ADD new_id uuid NOT NULL DEFAULT '46d9b2f0-0718-4d4c-a5a1-5af81d5b11e0',
-      ADD numbers int[],
+      ADD numbers int4[],
       DROP old_column,
       DROP employee_id;
     SQL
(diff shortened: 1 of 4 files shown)


akadusei:
Don't merge yet. I would like to confirm first that this works on CockroachDB...

akadusei:
My bad, but this was actually fixed in #900. I'm running the same tests against Avram master, and the issue does not exist.

So this PR is not really necessary. You may still merge, though, to forestall any surprises in future.

jwoertink:
Ah, cool. That's good then. I'm trying to get another release out, but still hung up on a few things.

Looking at the pg docs, it seems they're just aliases, but maybe using `int4` specifically will just be more explicit for other PG-like adapters. I'm cool with merging it if it's not gonna affect anything. Like, if I have an existing app which uses `int` on one table and `int4` on another, it should still all work the same, right?

akadusei:
Sorry, again, but #900 did not fix this wholly, so this PR is still needed. Without this change, some specs in another app using [Bill](https://github.com/GrottoPress/bill) fail with the same error.

jwoertink (approved):

jwoertink:
I assume those specs are all passing with this? Do you think we need to change `bigint` to `int8` and `smallint` to `int2`? Or was it just that `int` was ambiguous and cockroach treated it differently?

jwoertink:
It would be great if we could just use CockroachDB in the CI specs. Looks like it may not be possible just yet?  https://github.com/cockroachdb/cockroach/issues/87043

akadusei:
> Do you think we need to change bigint to int8 and smallint to int2?

No, its just `int` and `serial` . All other types work OK, from what I've tested so far.

akadusei:
> It would be great if we could just use CockroachDB in the CI specs.

Yeah, I was thinking the same, but I wasn't sure. That should help a lot.

akadusei:
There's still one last issue with `#select_sum` methods. In CockroachDB, they return `PG::Numeric`  leading to this error:

```
cast from PG::Numeric to (Int64 | Nil) failed, at /path/to/lib/avram/src/avram/charms/int32_extensions.cr:51:9:51 (TypeCastError)
    from lib/avram/src/avram/charms/int32_extensions.cr:51:9 in 'select_sum'
    from lib/avram/src/avram/charms/int32_extensions.cr:55:9 in 'select_sum!'
    # ...
```

I'm pushing that as a separate PR.

PR #921, opened 2023-01-16
Fix `TypeCastError` with `#select_sum` in CockroachDB

akadusei:
Related to #920

```
cast from PG::Numeric to (Int64 | Nil) failed, at /path/to/lib/avram/src/avram/charms/int32_extensions.cr:51:9:51 (TypeCastError)
    from lib/avram/src/avram/charms/int32_extensions.cr:51:9 in 'select_sum'
    from lib/avram/src/avram/charms/int32_extensions.cr:55:9 in 'select_sum!'
    # ...
```


jwoertink (approved):
I dig it! Thanks for the PRs.

PR #924, opened 2023-01-17
Add CI specs for CockroachDB

akadusei:
Experimenting with specs for CockroachDB...

A redo of #923

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Lint
         run: ./bin/ameba
 
-  specs:
+  specs__postgresql:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


jwoertink:
With https://github.com/cockroachdb/cockroach/pull/98899 closed out, does this get easier?

akadusei:
This PR does not use docker; I'm installing and configuring CockroachDB directly in Ubuntu, and it works OK. (If docker is preferred, please let me know)

The problem is that we may have to check explicitly the DB type in places where CockroachDB incompatibility exists, and I'm not sure if this PR is the right approach? For instance, `citext` is not supported, so there should be a way to disable that for CockroachDB, at least in specs. There's also the lack of support for `RESTART IDENTITY` (cockroachdb/cockroach#38931), which may require that we remove it from `Avram::Database::DatabaseCleaner#truncate`. There's probably others to consider.

I have been meaning to create a separate issue to track incompatibilities, so we can come up with a definite design decision as to how to handle them in Avram, before moving ahead. Ideally, whatever decision we arrive at should be applicable to all other PostgreSQL-compatible databases, not just CockroachDB. I just haven't got around to doing it, yet.

jwoertink:
Ah, ok. That makes sense. I really wanted to avoid getting in to the DB switching game (which is why we don't support MySQL). I guess since Cockroach is meant to be postgres compatible, then it's ok... I agree though, we should think about the design changes we need so other postgres compatible DBs can be easily supported. I know there's a few of them out there now.

PR #925, opened 2023-01-17
Fix compile error when preloading optional associations

### LINKED ISSUE #922: Compile error for optional association preloads
I'm getting a compile error when preloading optional associations. It seems to originate from here:

https://github.com/luckyframework/avram/blob/6d5357eca70f63150e740f7ac6f3473a0c33da37/src/avram/associations/belongs_to.cr#L64

```
Error: 

The 'eq' method can't compare a column to a value that may be 'nil'.
# ...
```

It seems the `#eq` method does not accept nilable arguments.

akadusei:
Fixes #922

### DIFF
--- a/spec/avram/preloading/preloading_belongs_to_spec.cr
+++ b/spec/avram/preloading/preloading_belongs_to_spec.cr
@@ -209,5 +209,15 @@ describe "Preloading belongs_to associations" do
         comments[1].post.title.should eq("THIS IS CHANGED")
       end
     end
+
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
I'm gonna approve this, but @matthewmcgarvey has spent way more time in this section than me. So I'll leave it just for a second pair of eyes.

matthewmcgarvey (approved):
:shipit:

PR #927, opened 2023-01-28
Support String primary key

### LINKED ISSUE #21: Feature request: Add support for prefixed KSUID primary_key_type
It would be great if lucky_record could support prefixed [KSUID](https://github.com/Sija/ksuid.cr) as a _primary_key_type_. 

Currently the only supported types are bare _int_ and _UUID_.

A modern ID is short, optionally k-sortable and optionally prefixed with an abbreviated model id. Example (from Stripe): `inv_SojvcbC89I7uo5`

It would be great if lucky models could be configured to
generate IDs like the above by adding a line such
as this to the model file:

`primary_key_type KSUID, prefix: 'inv'`

albertorestifo:
This PR adds support for a `String` primary key, with a required user-specified generator `Proc`.

Closes #21


spec/support/models/discount.cr
```diff
@@ -0,0 +1,18 @@
+class Discount < BaseModel
+  skip_default_columns
+
+  table do
+    primary_key id : String, value_generator: ->{ Random::Secure.hex }
```
albertorestifo, line 5: `value_generator`

Better name suggestions are very welcome, I don't like this name

albertorestifo:
Opened the PR as draft to collect early feedback. If this is the right direction, then I'll implement more tests.

spec/support/models/discount.cr
albertorestifo: I was mostly looking for something shorter, but this is good enough for now I guess.

Thanks for the feedback, I'll write more tests to touch all the same paths that the UUID tests touch

src/avram/model.cr
```diff
@@ -107,6 +107,26 @@ abstract class Avram::Model
...
+                  primary_key id : String, value_generator: ->{ KSUID.new }
+                  ...
+                end
+              ERROR
+          %}
+      {% end %}
+
+      def self.primary_key_value_generator : String
+        {{value_generator}}.call
+      end
```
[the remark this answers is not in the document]
jwoertink: One way around that might be to pass this value to another method that does type checking.

```crystal
def self.primary_key_value_generator : String
  ensure_value_is_string!({{value_generator}}.call)
end

def self.ensure_value_is_string(value : String)
  value
end

def self.ensure_value_is_string(value)
  {% raise "Value is not a string" %}
end
```

spec/support/models/discount.cr
jwoertink: I think `value_generator` is ok. Maybe a bit verbose, but you wouldn't want to use `value` since that would indicate the return value of this.

Does this work with the `autogenerated` option? Say you want a String primary key, but the value is generated from postgres.... Are you able to just skip this?

albertorestifo:
I'll resume work on this MR, I've a bit more time to work on side-projects now :D

PR #930, opened 2023-02-19
Applying latest ameba suggestions

jwoertink:
Looks like we're pulling ameba 1.4 now that includes some new suggestions.


matthewmcgarvey (approved):

jwoertink:
I'm gonna merge this. The Crystal nightly is currently failing, so it may be a bit before this can fully be green. Tested locally, and it's all good.

PR #935, opened 2023-03-30
Refine CRUD templates

zw963:
1. links use non keyword argument form.
2. Remove unnecessary namespaces.

### DIFF
--- a/src/lucky/tasks/gen/templates/resource/pages/edit_page.cr.ecr
+++ b/src/lucky/tasks/gen/templates/resource/pages/edit_page.cr.ecr
@@ -4,7 +4,7 @@ class <%= pluralized_name %>::EditPage < MainLayout
   quick_def page_title, "Edit <%= resource %> with id: #{<%= underscored_resource %>.id}"
 
   def content
-    link "Back to all <%= pluralized_name %>", <%= pluralized_name %>::Index
+    link "Back to all <%= pluralized_name %>", to: <%= pluralized_name %>::Index
     h1 "Edit <%= resource %> with id: #{<%= underscored_resource %>.id}"
     render_<%= underscored_resource %>_form(operation)
   end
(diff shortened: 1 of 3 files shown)


jwoertink:
I think the namespaces are actually better to have. These help reduce ambiguity when glancing over code, as well as help searching when doing grep across an entire project. It's easier to search for `Posts::Index` than it is `Index`. Since it's all generated code, I feel it's ok to be a little extra verbose here.

zw963:
> I think the namespaces are actually better to have. These help reduce ambiguity when glancing over code, as well as help searching when doing grep across an entire project. It's easier to search for `Posts::Index` than it is `Index`. Since it's all generated code, I feel it's ok to be a little extra verbose here.

Okay, done.

zw963:
Wow, this PR not review again since a year ago?

src/lucky/tasks/gen/templates/resource/pages/{{folder_name}}/index_page.cr.ecr
```diff
@@ -4,7 +4,7 @@ class <%= pluralized_name %>::IndexPage < MainLayout
 
   def content
     h1 "All <%= pluralized_name %>"
-    link "New <%= resource %>", to: <%= pluralized_name %>::New
+    link "New <%= resource %>", <%= pluralized_name %>::New
```
jwoertink, line 7: I personally prefer there to be the `to:`. I always add it in my projects. But since line 15 is defined without it, I guess we can do this. Or we need to update line 15 to include the `to:` just so they both match.

jwoertink:
Sorry, I just forgot to come back to this. It looks like it's behind now though, so it either needs to be updated or closed out.

zw963:
Sorry, i push forcefully.

i add `to:` to all template, please check.

jwoertink (approved):
I like this change :raised_hands: It's very clear and consistent. Thank you for adding this.

PR #942, opened 2023-05-15
Removes the need for pg client tools

### LINKED ISSUE #926: Investigate removing pg client requirement
Many of the tasks require pg client tools to be installed like `db.create` just runs `createdb ...`, and `db.drop` runs `dropdb ...`... This is super convenient, and really easy to interface with. The downside is the require the install postgres client tools.

I'm pretty sure we can run all of these commands in straight SQL on the server, but I have no clue how. I've always used the client tools even before Lucky, so that's what I'm comfortable with. We will need to dive in to the code and find all the spots where CLI stuff is used. Including the stuff that uses `psql`. For example, how do you do a DB dump and DB restore from within raw SQL?

I really don't want to play the game of "as long as you don't use X, then you won't need client tools". If we can't make it an all or nothing, then I don't think it'll be worth the hassle.

jwoertink:
Fixes #926

This doesn't fully remove the need for it, but it does remove the need for it when doing `db.create` and `db.drop` tasks.

The two main areas still remaining are `db.schema.dump` and `db.schema.restore` tasks which currently require `psql` and `pg_dump` to exist. I would like to somehow catch those and raise some sort of error before you run them. Right now if you don't have pg client tools installed and try to run either of those, you'll just end up with some exceptions saying it couldn't find those commands. Not horrible, but not great...

After doing this, there was some weird things that came up. Previously you could do this:

```crystal
SomeDatabase.run do |db|
  db.exec "SELECT some_sql"
end

DB::Drop.run_task
```

That seemed to be ok maybe because it was running `dropdb` from the CLI? :man_shrugging:  However, now, this raises an exception saying the that DB is currently in use. I guess this is a good thing, but it required adding a way to ensure all connections and the DB were fully closed. I'll add some more comments on this PR

### DIFF
--- a/spec/avram/migrator/runner_spec.cr
+++ b/spec/avram/migrator/runner_spec.cr
@@ -11,4 +11,37 @@ describe Avram::Migrator::Runner do
       end
     end
   end
+
+  describe ".create_db" do
+    context "when the DB doesn't exist yet" do
+      it "creates the new DB" do
+        Avram.temp_config(database_to_migrate: SampleBackupDatabase) do
+          Avram::Migrator::Runner.create_db(quiet?: true)
+
+          DB.open(SampleBackupDatabase.credentials.url) do |db|
+            results = db.query_all("SELECT datname FROM pg_database WHERE datistemplate = false", as: String)
+            results.should contain("sample_backup")
+          end
+          # ensure it's deleted before moving on to another spec
+          Avram::Migrator::Runner.drop_db(quiet?: true)
+        end
+      end
+    end
+  end
+
+  describe ".drop_db" do
+    context "when it already exists" do
+      it "drops the DB" do
+        Avram.temp_config(database_to_migrate: SampleBackupDatabase) do
+          Avram::Migrator::Runner.create_db(quiet?: true)
+          Avram::Migrator::Runner.drop_db(quiet?: true)
+
+          expect_raises(DB::ConnectionRefused) do
+            DB.open(SampleBackupDatabase.credentials.url) do |_|
+            end
+          end
+        end
+      end
+    end
+  end
 end
--- a/spec/avram/tasks/db_schema_restore_spec.cr
+++ b/spec/avram/tasks/db_schema_restore_spec.cr
@@ -16,20 +16,27 @@ describe Db::Schema::Restore do
   end
 
   it "restores from the sample_backup file" do
-    Avram.temp_config(database_to_migrate: SampleBackupDatabase) do
-      Avram::Migrator::Runner.create_db(quiet?: true)
+    swap_database_with_cleanup(SampleBackupDatabase) do
       Db::Schema::Restore.new(SQL_DUMP_FILE).run_task
 
       SampleBackupDatabase.run do |db|
         value = db.scalar("SELECT COUNT(*) FROM sample_records").as(Int64)
         value.should eq 0
       end
-
-      # HACK: This is needed because the `run` command above
-      # opens the DB, and since it's never closed, we can't drop it
-      SampleBackupDatabase.close_connections
-      # make sure this is dropped before another spec runs
-      Avram::Migrator::Runner.drop_db(quiet?: true)
     end
+  end
+end
+
+private def swap_database_with_cleanup(database_to_migrate, &)
+  Avram.temp_config(database_to_migrate: database_to_migrate) do
+    # Create this new DB
+    Avram::Migrator::Runner.create_db(quiet?: true)
+
+    yield
+
+    # Ensure all connections to this DB are closed
+    SampleBackupDatabase.close_connections!
+    # make sure this is dropped before another spec runs
+    Avram::Migrator::Runner.drop_db(quiet?: true)
   end
 end
--- a/spec/avram/tasks/db_schema_restore_spec.cr
+++ b/spec/avram/tasks/db_schema_restore_spec.cr
@@ -17,15 +17,19 @@ describe Db::Schema::Restore do
 
   it "restores from the sample_backup file" do
     Avram.temp_config(database_to_migrate: SampleBackupDatabase) do
-      Db::Drop.new(quiet: true).run_task
-      Db::Create.new(quiet: true).run_task
-
+      Avram::Migrator::Runner.create_db(quiet?: true)
       Db::Schema::Restore.new(SQL_DUMP_FILE).run_task
 
       SampleBackupDatabase.run do |db|
         value = db.scalar("SELECT COUNT(*) FROM sample_records").as(Int64)
         value.should eq 0
       end
+
+      # HACK: This is needed because the `run` command above
+      # opens the DB, and since it's never closed, we can't drop it
+      SampleBackupDatabase.close_connections
+      # make sure this is dropped before another spec runs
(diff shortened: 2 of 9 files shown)


spec/avram/migrator/runner_spec.cr
```diff
@@ -11,4 +11,37 @@ describe Avram::Migrator::Runner do
...
     end
   end
+
+  describe ".create_db" do
+    context "when the DB doesn't exist yet" do
+      it "creates the new DB" do
+        Avram.temp_config(database_to_migrate: SampleBackupDatabase) do
+          Avram::Migrator::Runner.create_db(quiet?: true)
+
+          DB.open(SampleBackupDatabase.credentials.url) do |db|
```
jwoertink, line 21: This fully closes the DB connection when the block ends allowing me to drop the DB.

spec/avram/tasks/db_schema_restore_spec.cr
```diff
@@ -17,15 +17,19 @@ describe Db::Schema::Restore do
...
-      Db::Drop.new(quiet: true).run_task
-      Db::Create.new(quiet: true).run_task
-
+      Avram::Migrator::Runner.create_db(quiet?: true)
       Db::Schema::Restore.new(SQL_DUMP_FILE).run_task
 
       SampleBackupDatabase.run do |db|
         value = db.scalar("SELECT COUNT(*) FROM sample_records").as(Int64)
         value.should eq 0
       end
```
jwoertink, line 26: This does not close the connections or the DB when the block ends. It's still open...

spec/avram/tasks/db_schema_restore_spec.cr
```diff
@@ -17,15 +17,19 @@ describe Db::Schema::Restore do
...
+
+      # HACK: This is needed because the `run` command above
+      # opens the DB, and since it's never closed, we can't drop it
+      SampleBackupDatabase.close_connections
```
jwoertink, line 30: This feels like an odd spot to put this, but I can't drop the DB unless all connections are fully closed.

spec/spec_helper.cr
```diff
@@ -34,18 +34,6 @@ Spec.before_each do
   Fiber.current.query_cache = LuckyCache::NullStore.new
 end
 
-class SampleBackupDatabase < Avram::Database
-end
-
-SampleBackupDatabase.configure do |settings|
```
jwoertink, line 40: I moved this to `config/database.cr` where all the others are. It should have been there in the first place anyway :joy:

src/avram/connection.cr
```diff
@@ -1,10 +1,17 @@
...
 
   def open : DB::Database
-    try_connection!
+    @db = try_connection!
+  end
+
+  def close
+    @db.try(&.close)
+    @db = nil
   end
```
jwoertink, line 15: I don't really understand this. We've never had a `db.close` before, but it feels like there should have been... I think we just open a connection, then let it sit open in the connection pool forever. If you redeploy your app, that kills all of the connections which may explain why we always get IO errors randomly... But what's a better way to handle this?

src/avram/credentials.cr
```diff
@@ -80,6 +80,17 @@ class Avram::Credentials
     @query.try(&.strip).presence
   end
 
+  # This is the full connection string used
+  # to connect to the PostgreSQL server.
+  def connection_string : String
+    String.build do |io|
```
jwoertink, line 86: In order to run SQL before a database exists, we need the connection string `postgres://user:pass@host:5432`

src/avram/database.cr
```diff
@@ -158,6 +158,12 @@ abstract class Avram::Database
     end
   end
 
+  # Close all available connections as well as the DB
+  def self.close_connections
+    connections.values.map(&.close)
+    @@db.try(&.close)
+  end
```
jwoertink, line 165: Not really a fan of this... I'm basically just saying "grab everything and close it all". I'd assume if you ran this in your app, everything would go really bad :grimacing:

src/avram/migrator/runner.cr
```diff
@@ -48,33 +48,35 @@ class Avram::Migrator::Runner
   end
 
   def self.drop_db(quiet? : Bool = false)
-    run "dropdb #{cmd_args}"
+    DB.connect(credentials.connection_string) do |db|
+      db.exec "DROP DATABASE IF EXISTS #{db_name}"
+    end
```
jwoertink, line 53: If you've already dropped it, we can just silently ignore it. There's no real action to be taken.

src/avram/migrator/runner.cr
```diff
@@ -48,33 +48,35 @@ class Avram::Migrator::Runner
...
   def self.create_db(quiet? : Bool = false)
-    run "createdb #{cmd_args}"
+    DB.connect(credentials.connection_string) do |db|
+      db.exec "CREATE DATABASE #{db_name} WITH OWNER DEFAULT"
+    end
     unless quiet?
-      puts "Done creating #{Avram::Migrator::Runner.db_name.colorize(:green)}"
+      puts "Done creating #{db_name.colorize(:green)}"
+    end
+  rescue e : DB::ConnectionRefused
```
jwoertink, line 66: Not sure what level the issue is at, but when you can't connect, a `PQ::PQError` is raised with a nice helpful error... then that gets swallowed somewhere and a `DB::ConnectionRefused` is raised in it's place with no error message... so `e.message` is nil. They just expect you to read through the stack trace.

src/avram/migrator/runner.cr
```diff
@@ -48,33 +48,35 @@ class Avram::Migrator::Runner
...
+    message = e.message.to_s
+    if message.includes?(%("#{self.db_name}" already exists))
       unless quiet?
         puts "Already created #{self.db_name.colorize(:green)}"
       end
-    elsif (message = e.message) && (message.includes?("createdb: not found") || message.includes?("No command 'createdb' found"))
-      raise PGClientNotInstalledError.new(message)
-    elsif (message = e.message) && message.includes?("could not connect to database template")
+    elsif message.includes?("Cannot establish connection")
       raise PGNotRunningError.new(message)
```
jwoertink, line 80: I actually think this won't ever get here now... I'll have to double check, but I believe this is the same as above. It raises a DB::ConnectionRefused error with no message, and then in the stack trace there's a "Caused by..."

jkthorne:
Dis dependency has caused me problems in the past.  I ❤️ this change!

src/avram/credentials.cr
jwoertink: Oh, I was just saying that I moved this logic out so I can differentiate between

```
postgres://user:pass@host:5432
```

and

```
postgres://user:pass@host:5432/some_db
```

spec/avram/tasks/db_schema_restore_spec.cr
jwoertink: I guess that's a better option :+1:

jwoertink:
Alright, had to tell create and drop DB that it *has* to connect to some database... not sure why ONLY that one instance required it, but my guess is that you'll always have a DB named the same as your runner's name. If not, then we'd have to make that configurable as "maintenance_database_name" or something weird...

PR #945, opened 2023-06-23
Suppress logging in migrator

### LINKED ISSUE #937: Logging queries no longer seems to work
Just doing `Avram::Log.info { "testing" }` doesn't seem to print out the logs in a default app.

According to a default generated Lucky app `Avram::QueryLog.dexter.configure(:info)` should be all you do to see all of your SQL queries, but nothing happens. You should also be able to just throw  `Avram::QueryLog.dexter.info { {test: "hi"} }` in any action and see it printed out, but that also doesn't happen. No clue when this broke or how to fix it.

garymardell:
Fixes #937

The initializer of `Avram::Migrator::Runner` re-configures the avram logger to be level `:none`. The runner is instantiated in development during `start_server.cr` to ensure that migrations have been run (after configuration has been loaded). The change was introduced in https://github.com/luckyframework/avram/pull/780

I am assuming the intention was to suppress the logging when the migrator is checking if migrations have run, creating schema tables etc. Instead I have added `suppress_logging` method that uses dexter to temporarily set the level to `:none` and reset after. There might be a nicer way of doing this but there doesn't appear to be a single entrypoint into the migrator that can be configured.

### DIFF
--- a/src/avram/migrator/migration.cr
+++ b/src/avram/migrator/migration.cr
@@ -28,15 +28,17 @@ abstract class Avram::Migrator::Migration::V1
   # @prepared_statements array. Each statement is then executed in  a
   # transaction and tracked upon completion.
   def up(quiet = false)
-    if migrated?
-      puts "Already migrated #{self.class.name.colorize(:cyan)}"
-    else
-      reset_prepared_statements
-      migrate
-      execute_in_transaction @prepared_statements do |tx|
-        track_migration(tx)
-        unless quiet
-          puts "Migrated #{self.class.name.colorize(:green)}"
+    suppress_logging do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
ohh!! Good catch :joy: Thanks for fixing this!

garymardell:
Turns out I wasn't returning the yielded value back from `suppress_logging` leading to `migrated?` not returning a correct value. I also broke the style checker which should be fixed now.

jwoertink:
Sweet. That Ameba error is popping up everywhere, so I'm not worried about that. I'll go ahead and merge this in. Thanks for fixing it!

PR #948, opened 2023-06-25
Fix error message for numerical validation with non integers

### LINKED ISSUE #858: validate_numeric shows wrong message when using floats
https://github.com/luckyframework/avram/blob/549639bb956a3f59d409800714e0021d722b4d66/src/avram/i18n_backend.cr#L17-L18

Using this code:

```crystal
validate_numeric number, greater_than: 0.01, less_than: 0.99
```

Would show the error `"must be less than 0" ` or `"must be greater than 0"` since the string uses `%d` instead of `%f`. Though, even with `%f`, you'll get floating precision like `must be less than 0.990000`

garymardell:
Fixes #858

Was just looking through the issues list and this seemed like an easy fix. Despite the condition being numerical I think using the string based representation of the number makes more sense as a default and will work for all numbers (floats / negative etc) of any precision. A user can pass in a custom specific message with different formatting if required (say, they only wanted specific precision for floats)


jwoertink (approved):
:rocket: great solution!

PR #949, opened 2023-06-25
Adding JSON serialized Array columns

### LINKED ISSUE #845: JSON serialize with Array does not work
When having an array of a type in the JSON::Serializable, it doesn't work to use:
```
column exceptions : Array(Sentry::Exception), serialize: true
```
The one above doesn't work at all as 
> Error: no overload matches 'JSON::Lexer.new' with type Array(String)

With trace to:
> Error: instantiating 'JSON::Serializable::Lucky(Sentry::Exception).class#parse(Array(String))'

When mentioning this error on discord, I tested:
```
column exceptions : Sentry::Exception, serialize: true
```
Which don't complain, and make sense as JSON::Serializable makes some handle under the hood with Array.
But now I can't save, as the type return in the type is `Array(Sentry::Exception)` and not `Sentry::Exception` And get type wrong error from Avram when trying to save. (wrong params)

So I now have to use `JSON::Any` temporarily till this is fixed.

Having the full support of JSON Serialize would simplify JSON handling in the database and the saving of the same.

jwoertink:
Fixes #845

This PR allows you to convert your json arrays in to an array of serialized objects.. One thing to note about this PR is that it's only supporting `'[]'::jsonb` and *not* `[]::jsonb[]`. The distinction here is that the column is still stored as `jsonb`, but it allows that json string to be parsed as an array VS using postgres's ARRAY column to store an array of jsonb objects.

I think we could eventually do that one too by just saying that the column isn't serialized, but that can be for a future PR.

With this, you can now do something like:

```crystal
struct Server
  include JSON::Serializable
  property host : String
end

class Monitor < BaseModel
  table do
    column input_server : Server
    column output_servers : Array(Server)
  end
end

in_server = Server.from_json(%({"host":"x.y.z"}))
out_servers = Array(Server).from_json(%([{"host":"a.b.c"}, {"host":"d.e.f"}]))

SaveMonitor.create!(input_servers: in_server, output_servers: out_servers)

m = MonitorQuery.new.first
m.output_servers.first.host == "a.b.c"
```

### DIFF
--- a/db/migrations/20230624134902_add_json_array_to_blob.cr
+++ b/db/migrations/20230624134902_add_json_array_to_blob.cr
@@ -1,0 +1,13 @@
+class AddJsonArrayToBlob::V20230624134902 < Avram::Migrator::Migration::V1
+  def migrate
+    alter table_for(Blob) do
+      add servers : JSON::Any, default: JSON::Any.new({} of String => JSON::Any)
+    end
+  end
+
+  def rollback
+    alter table_for(Blob) do
+      remove :servers
+    end
+  end
+end
--- a/db/migrations/20230624134902_add_json_array_to_blob.cr
+++ b/db/migrations/20230624134902_add_json_array_to_blob.cr
@@ -1,7 +1,7 @@
 class AddJsonArrayToBlob::V20230624134902 < Avram::Migrator::Migration::V1
   def migrate
     alter table_for(Blob) do
-      add servers : JSON::Any, default: JSON::Any.new({} of String => JSON::Any)
+      add servers : JSON::Any, default: JSON::Any.new([] of JSON::Any)
     end
   end
 
--- a/src/avram/charms/json_extensions.cr
+++ b/src/avram/charms/json_extensions.cr
@@ -16,13 +16,21 @@ module JSON::Serializable
       value
     end
 
+    def parse(value : Array(JS)) forall JS
+      SuccessfulCast(Array(JS)).new value
+    end
+
     def parse(value : JSON::Serializable)
       SuccessfulCast(JSON::Serializable).new value
     end
 
-    def parse(value)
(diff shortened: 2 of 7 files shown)


db/migrations/20230624134902_add_json_array_to_blob.cr
```diff
@@ -0,0 +1,13 @@
+class AddJsonArrayToBlob::V20230624134902 < Avram::Migrator::Migration::V1
+  def migrate
+    alter table_for(Blob) do
+      add servers : JSON::Any, default: JSON::Any.new({} of String => JSON::Any)
```
jwoertink, line 4: oops, just realized this is the wrong default :joy:

src/avram/charms/json_extensions.cr
```diff
@@ -16,13 +16,21 @@ module JSON::Serializable
...
-    def parse(value)
-      SuccessfulCast(JSON::Serializable).new T.from_json(value)
-    end
+    # def parse(value : String)
+    #   if value.starts_with?('[')
+    #     SuccessfulCast(JSON::Serializable).new Array(T).from_json(value)
+    #   else
+    #     SuccessfulCast(JSON::Serializable).new T.from_json(value)
+    #   end
+    # end
```
jwoertink, line 33: this is not needed anymore.

src/avram/charms/json_extensions.cr
```diff
@@ -16,13 +16,21 @@ module JSON::Serializable
       value
     end
 
+    def parse(value : Array(JS)) forall JS
+      SuccessfulCast(Array(JS)).new value
```
jwoertink, line 20: Actually, looks like I can use `T` here... I tried earlier and got a Lexer exception, but just tried again locally and it worked.. :thinking:

matthewmcgarvey (approved):

PR #950, opened 2023-06-25
Adding a new belongs_to association for factories.

### LINKED ISSUE #385: Make it easier to set default associations in a box

jwoertink:
Fixes #385

This PR adds in a new `belongs_to` association macro for Factory that allows you to association another model and build on demand.

```crystal
class ProductFactory < Avram::Factory
  belongs_to store : StoreFactory

  def initialize
    title "Cool Thing"
    create_store(&.name("Example Store"))
  end
end

# You can use it many different ways

product = ProductFactory.create
product.store.name == "Example Store"

product = ProductFactory.create(&.create_store(&.name("Different Store")))
product.store.name == "Different Store"

store = StoreFactory.create(&.name("Evil Store"))
product = ProductFactory.create &.store(store)
product.store_id == store.id
```

One weird thing I ran in to is naming order with requires, but I think that's just a Crystal thing...

For example, with a `require "./factories/**"`, doing this will throw a `undefined constant PostFactory` because in alphabetical order, "PostFactory" comes after "CommentFactory", so the file hasn't been loaded. You'd just need to do an extra `require "./post_factory"` at the top if that's the case...

```crystal
  belongs_to post : PostFactory
end
```

### DIFF
--- a/src/avram/factory.cr
+++ b/src/avram/factory.cr
@@ -17,6 +17,37 @@ abstract class Avram::Factory
     {% end %}
   end
 
+  # Creates a method named from the TypeDeclaration for assigning the foreign key value
+  # from the object passed in. And a second method for creating the association on saves.
+  # ```
+  # class CommentFactory < Avram::Factory
+  #   belongs_to post : PostFactory
+  #
+  #   def initialize
+  #     create_post
+  #   end
+  # end
+  # ```
+  macro belongs_to(type_declaration)
+    {% unless type_declaration.is_a?(TypeDeclaration) %}
+      {% raise "belongs_to expected a type declaration like 'user : UserFactory', instead got: '#{type_declaration}'" %}
+    {% end %}
+    def {{ type_declaration.var }}(%m : {{ type_declaration.type.stringify.gsub(/Factory$/, "").id }})
+      {{ type_declaration.var }}_id(%m.id)
+    end
+    def create_{{ type_declaration.var }}(&blk : {{ type_declaration.type }} ->)
+      before_save do
+        if operation.{{ type_declaration.var }}_id.value.nil?
+          %m = {{ type_declaration.type.resolve }}.create do |inst|
+            blk.call(inst)
+            inst
+          end
+          {{ type_declaration.var }}(%m)
+        end
+      end
+    end
+  end
+
   macro setup_attribute_shortcuts(operation)
     {% for attribute in operation.resolve.constant(:COLUMN_ATTRIBUTES) %}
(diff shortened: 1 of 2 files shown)


src/avram/factory.cr
```diff
@@ -17,6 +17,37 @@ abstract class Avram::Factory
...
+  # end
+  # ```
+  macro belongs_to(type_declaration)
+    {% unless type_declaration.is_a?(TypeDeclaration) %}
+      {% raise "belongs_to expected a type declaration like 'user : UserFactory', instead got: '#{type_declaration}'" %}
+    {% end %}
+    def {{ type_declaration.var }}(%m : {{ type_declaration.type.stringify.gsub(/Factory$/, "").id }})
+      {{ type_declaration.var }}_id(%m.id)
+    end
+    def create_{{ type_declaration.var }}(&blk : {{ type_declaration.type }} ->)
```
matthewmcgarvey, line 38: Does this allow calling the method without a block? I'm not sure it does

jwoertink: Ooo, good call. I'll add that in.

jwoertink:
Putting this in WIP... Just ran in to an issue that I'm not sure how to handle yet.. If your Factory has `create_thing(&.some_default)` set in the initializer, and you want to override that, it's currently not working.

```crystal
class CommentFactory < Avram::Factory
  belongs_to post : PostFactory

  def initialize
    body("Test Comment")
    create_post(&.title("Test Post"))
  end
end

# the post title will be "Test Post" here...
TempCommentFactory.create(&.create_post(&.title("A new post")))
```

This is because it chains the `before_saves`, and runs the first one, then doesn't run the second because the value has already been set.

PR #957, opened 2023-08-04
Fixing upsert so you can update a field to nil.

### LINKED ISSUE #864: Using upsert to change a nilable column value to nil doesn't seem to work
If you have a SaveOperation using `upsert`, and one of the lookup columns is nilable, it doesn't seem to allow you to update that field to `nil`.

jwoertink:
Fixes #864

This fixes an issue when using `upsert` where you may want to update a field's value to `nil` when the record is found.

/cc. @franciscoGPS


franciscoGPS:
Thank you Jeremy! :pray:

PR #962, opened 2023-08-10
Reuse query builder and insert

garymardell:
Just a tiny PR but I noticed that for each update and insert it is calling `update_query` and `insert_sql` respectively twice rather than re-using the same generated object.


jwoertink (approved):
:+1: nice catch!

PR #965, opened 2023-08-16
Code Cleanup

### LINKED ISSUE #964: Avram is borking Benchmark.memory
I need to investigate more, but this is weird...

If you run this program, you should see an output of `32`

```crystal
require "benchmark"
total = Benchmark.memory do
  Array(Int32).new
end

pp total
```

Now run this program
```crystal
require "avram"
require "benchmark"
total = Benchmark.memory do
  Array(Int32).new
end

pp total
```

You should see an output of `0` :raised_eyebrow: 

Removing these lines seems to fix it:

https://github.com/luckyframework/avram/blob/29d89adc4f716b41fa3da87fa3f875188bf8c477/src/avram.cr#L20-L21

I have no clue what's going on

jwoertink:
Closes #964

There was a few random issues that came up while I was investigating. It seems that when you use `require "./avram/**"` Crystal does something weird with the double splat. I had to change it so I could narrow down where the error was being called from, and as soon as I did that, I started getting compilation errors about `Database` being a class and not a module. Then I started getting weird errors like `settings` is not a method on `Database` along with a lot of random type errors. Some that double splat require set things up just perfect to "magically" run... It didn't seem to solve the issue though. I now believe the issue is actually in LuckyTask, but I need more investigating.

For now I'll keep this open while I do more digging.


jwoertink:
This PR was originally an investigation in to #964 but after talking with people in the Crystal discord, they seem to think this isn't an issue, and just a bad use of the method. :man_shrugging:  So I pivoted and just turned this in to some code cleanup. More type returns and ameba updates. The only strange thing I really found was with the `Database::ColumnInfo`,  `Database::DatabaseInfo`, and `Database::TableInfo`. Depending on how you require the files, Crystal may treat `Database` like a module here even though it's actually a class. So to avoid this potential race condition, I've made it explicit.

shard.override.yml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  lucky_task:
+    github: luckyframework/lucky_task
+    version: ~> 0.2.0
```
jwoertink, line 4: This can go away once all of the shards are using this version.

src/avram/database/column_info.cr
```diff
@@ -1,26 +1,28 @@
 module Avram
-  struct Database::ColumnInfo
-    include DB::Serializable
+  class Database
+    struct ColumnInfo
```
jwoertink, line 3: This avoids a potential issue where this file gets required before the `Database` class does causing this namespace to be treated like a module.

PR #971, opened 2023-09-12
Add support for other postgres index types.

### LINKED ISSUE #969: Support GIN indexes
https://github.com/luckyframework/avram/blob/f84e111a80d475611eba5b1a082ca8d8d8df399e/src/avram/migrator/create_index_statement.cr#L31

:man_facepalming:

jwoertink:
Fixes #969

While I was in there already, I just decided to add a few other supported index types. The only one I didn't add was the `sp-gist` which seems to be like `gist` but slightly more limited? I guess if anyone really wants it, we can add it in, but just made the code more complex since I couldn't just `SpGist.to_s.downcase` :grimacing:

### DIFF
--- a/src/avram/migrator/create_index_statement.cr
+++ b/src/avram/migrator/create_index_statement.cr
@@ -28,10 +28,17 @@ require "./index_statement_helpers"
 class Avram::Migrator::CreateIndexStatement
   include Avram::Migrator::IndexStatementHelpers
 
-  ALLOWED_INDEX_TYPES = %w[btree]
+  enum IndexTypes
+    Btree
+    Hash
+    Gist
+    Gin
+    Brin
+  end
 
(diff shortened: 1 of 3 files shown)


src/avram/migrator/create_index_statement.cr
```diff
@@ -28,10 +28,17 @@ require "./index_statement_helpers"
...
+    Btree
+    Hash
+    Gist
+    Gin
+    Brin
+  end
 
-  def initialize(@table : TableName, @columns : Columns, @using : Symbol = :btree, @unique = false, @name : String? | Symbol? = nil)
-    raise "index type '#{using}' not supported" unless ALLOWED_INDEX_TYPES.includes?(using.to_s)
+  def initialize(@table : TableName, @columns : Columns, using : Symbol = :btree, @unique = false, @name : String? | Symbol? = nil)
```
jwoertink, line 39: I was going to make this use the enum instead of a Symbol, but due to how the Crystal autocast works with passing a Symbol, it would have been a breaking change. This can be an optimization for later.

PR #975, opened 2023-10-18
fix handling of bytes outside ASCII range

manveru:


mdwagner:
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE

As of postgres 9.1, the server defaults to not auto escaping backslashes, so this PR makes sense. We should also raise this with pg shard, as I think their implementation is outdated (or could be adjusted).

PR #980, opened 2023-10-31
Fix CockroachDB errors

akadusei:
Fixes #979

PR #984, opened 2023-11-08
Add missing parameters to `Avram::Database.truncate` method

akadusei:
Missed these in #980

### DIFF
--- a/spec/avram/database_cleaner_spec.cr
+++ b/spec/avram/database_cleaner_spec.cr
@@ -9,4 +9,46 @@ describe "DatabaseCleaner" do
       UserQuery.new.select_count.should eq 0
     end
   end
+
+  describe "#truncate" do
+    it "restarts identity", tags: Avram::SpecHelper::TRUNCATE do
+      count = 3
+
+      count.times do
+        UserFactory.create
+        ArticleFactory.create
+      end
+
+      UserQuery.new.select_count.should eq(count)
+      ArticleQuery.new.select_count.should eq(count)
+
+      TestDatabase.truncate(restart_identity: true)
+
+      UserQuery.new.select_count.should eq(0)
+      ArticleQuery.new.select_count.should eq(0)
+
+      UserFactory.create.id.should eq(1)
+      ArticleFactory.create.id.should eq(1)
+    end
+
+    it "does not restart identity", tags: Avram::SpecHelper::TRUNCATE do
+      count = 3
+
+      count.times do
+        UserFactory.create
+        ArticleFactory.create
+      end
+
+      UserQuery.new.select_count.should eq(count)
+      ArticleQuery.new.select_count.should eq(count)
+
+      TestDatabase.truncate(restart_identity: false)
+
+      UserQuery.new.select_count.should eq(0)
+      ArticleQuery.new.select_count.should eq(0)
+
+      UserFactory.create.id.should eq(count + 1)
+      ArticleFactory.create.id.should eq(count + 1)
+    end
+  end
 end
--- a/src/avram/database.cr
+++ b/src/avram/database.cr
@@ -41,8 +41,8 @@ abstract class Avram::Database
   end
 
   # Run a SQL `TRUNCATE` on all tables in the database
-  def self.truncate
-    new.truncate
+  def self.truncate(**named_args)
+    new.truncate(**named_args)
   end
 
   # Run a SQL `DELETE` on all tables in the database
@@ -208,8 +208,8 @@ abstract class Avram::Database
     connection._avram_stack.last?
   end
 
-  protected def truncate
-    DatabaseCleaner.new(self).truncate
+  protected def truncate(**named_args)
+    DatabaseCleaner.new(self).truncate(**named_args)
   end
 
   protected def delete
--- a/src/avram/spec_helper.cr
+++ b/src/avram/spec_helper.cr
@@ -1,5 +1,7 @@
 module Avram::SpecHelper
-  TRUNCATE = "truncate"
+  TRUNCATE            = "truncate"
+  NO_CASCADE          = "no_cascade"
+  NO_RESTART_IDENTITY = "no_restart_identity"
 
   macro use_transactional_specs(*databases)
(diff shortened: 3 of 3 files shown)


jwoertink (commented):
oops! :sweat_smile:  Guess I missed these too... I thought we had specs around truncate, but maybe not? Would you be able to add a spec in just so we have these accounted for in case there's other changes?

akadusei:
> I thought we had specs around truncate, but maybe not?

https://github.com/luckyframework/avram/blob/0b9797b918373615bafea54986849bb886ec239b/spec/avram/queryable_spec.cr#L1244-L1266

There's none for `Avram::Database.truncate`, but I guess it's implicitly tested given that every test tagged with `Avram::SpecHelper::TRUNCATE` performs a truncate?

jwoertink (approved):

akadusei:
Line 13 here produces same error in CockroachDB:

https://github.com/luckyframework/avram/blob/0b9797b918373615bafea54986849bb886ec239b/src/avram/spec_helper.cr#L10-L15

Should the call to `#truncate` factor in `restart_identity` and `cascade` arguments? If yes, what is the best way to do it?

jwoertink:
That probably doesn't matter much for the specs, I would assume. Does Cockroach require this to get changed in order to work?

akadusei:
> Does Cockroach require this to get changed in order to work?

Yes. `Avram::Database::DatabaseCleaner#truncate`  has `restart_identity` set to `true` by default, which inserts `RESTART IDENTITY` into the SQL statement. `RESTART IDENTITY` is not supported by Cockroach DB.

jwoertink:
Gotcha... hmm :thinking:  Would it be weird to have a property on the database that decides if it should have this or not? Maybe it's time to think about an `engine` property that's defaulted to `postgresql`, but has `cockraochdb` as an option.. Maybe this is an enum property or something. Then that line starts to look like

```crystal
databases.each do |db|
    db.truncate(restart_identity: db.engine.postgresql?)
end
```

Or does that sound like too much for this case? Maybe a simple solution is just passing in that option? I'm not sure if there would be a time where you use postgres and don't want the `restart_identity`, but there probably is a case for that...

akadusei:
I'm thinking more along the lines of checking for "no_restart_identity" tag on the spec, and setting `restart_identity: false` if found. (Or checking for "restart_identity" tag and setting `restart_identity: true` if found, depending on what we want as default)

 > Maybe it's time to think about an engine property that's defaulted to postgresql, but has cockraochdb as an option...

If we do this, we should not need to set `restart_identity` to get around the error. In the `#truncate` mehtods, we can simply check the db engine type and exclude "RESTART IDENTITY" from the resulting SQL statement. So the `restart_identity` parameter would still exist, but would do nothing in the case of Cockroach DB.

I believe this raises the question of the broader support for other DB engines as discussed in https://github.com/luckyframework/avram/pull/924#issuecomment-1496520299.

jwoertink:
Ok. Maybe we just add the tag in as a temporary workaround for now with a `TODO` note about updating it later. I'm not sure the best direction to go.

> I believe this raises the question of the broader support for other DB engines

Yeah, I still am of the feeling that I would love to avoid all the extra work of supporting extra engines. It's a bit frustrating that Cockroach says it's postgres compatible, but seems to have an asterisk. The tough part here is that we currently have no way to test if a change will break Cockroach. But I guess we're at the point that we either need to say we can't support it at all, or bite the bullet and just start supporting more things.

So I guess what I'm saying is, let's just hack around this issue for now so you're not blocked on anything, then we can start planning out a path to add more support for other engines as well as figuring out a way to test this to prevent this issue down the road.

akadusei:
> Ok. Maybe we just add the tag in as a temporary workaround for now with a TODO note about updating it later. 

Sure, will see what I can come up with.

> Yeah, I still am of the feeling that I would love to avoid all the extra work of supporting extra engines.

Me too. I figure a shard that extends Avram (mostly via monkey-patching), purposely for Cockroach DB (or whatever engine anyone needs supported) should do the trick. These could be maintained by persons outside the core team.

I added something similar in *Lucille* (https://github.com/GrottoPress/lucille/commit/12cb92955622ffbf6ae40010f1d4ad561d0eaa5e), but removed it after the relevant issues were fixed in #980. The idea is that any issues with any of the DBs would get patched in these shards (which should have shorter release cycles), and may or may not be patched upstream. In any case, for as long as an application does `require "lucille/cockroach"`, for instance, it should be guaranteed to work with Cockroach.

> The tough part here is that we currently have no way to test if a change will break Cockroach

I have added Lucky/Avram `main` branches to the tests in CI for all my shards, so any changes that break Cockroach DB should be caught (and hopefully fixed) before subsequent releases. This is probably not the ideal solution, but an extension shard (as before described) could add its own tests relevant for whatever DB it is supporting.

> It's a bit frustrating that Cockroach says it's postgres compatible, but seems to have an asterisk.

From what I understand from their end, some things are difficult to implement because of the very nature of Cockroach's architecture. The few issues in Avram had to do with spec and migration helpers. It works very well otherwise; one could always drop down to raw SQL if desired.

> So I guess what I'm saying is, let's just hack around this issue for now so you're not blocked on anything...

Don't worry about me :smiley:. I tend to find my way somehow. Whatever we decide should work well for the community at large.

akadusei:
Calling `#truncate` directly inside the spec works even when using transactional specs, so I'm wondering if the whole `.use_truncation?` check should be deprecated/removed instead? That way no one has to worry about how `#truncate` is called inside `.wrap_spec_in_transaction`? :thinking:

jwoertink:
Yeah, you can do your own manual truncate in the specs, but it's currently setup to do it automatically for you so you don't have to call it on every spec. This was added in as part of a way to make specs run faster. I think if we took it out, then you'd have to decide how to clear your data between specs always whether that's with truncate or with transactional rollbacks. That may complicate things more, I think.

jwoertink:
I think this is good, so I'm just gonna merge, but if there's more updates we need for this, let me know. Thanks @akadusei!

PR #995, opened 2023-12-28
Revert and cleanup the verify_connection task

### LINKED ISSUE #902: Issue with the verify_connection task
A user in [Discord](https://discord.com/channels/743896265057632256/743896265057632259/1032919848952545290) ran in to an issue when generating a new Lucky app (both API and Web) where it would tank on `lucky db.verify_connection`.

```
➜  myy_api ✗ lucky -v  
1.0.0-rc1
➜  myy_api git: ✗ crystal -v        
Crystal 1.6.0 [41573fadb] (2022-10-06)

LLVM: 13.0.1
Default target: x86_64-unknown-linux-gnu
➜  myy_api git: ✗ openssl version -a

OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)
```

According to the user, they were using Postgres 14.5 locally on Ubuntu.

After lots of debugging and help from the community, we found that the DB was being created and migrated just fine, it was only the verify_connection that would fail. 

I had the user change this line

https://github.com/luckyframework/avram/blob/1267ffcf2de5aef13c3325eff16c8ca748c1bfa7/src/avram/tasks/db/verify_connection.cr#L23

back to the old way which changed in [this commit](https://github.com/luckyframework/avram/commit/36e2f87b61a6cb7061aaa34035f1ec6dbd6f399c)

to this code:

https://github.com/luckyframework/avram/blob/f6672493918d74771b7b3ef0f7a42eabde02d905/src/avram/tasks/db/verify_connection.cr#L23-L24

And doing this worked. My best guess is the current way calls `open.close` which maybe has some sort of race condition happening somewhere. I think in this instance, we can just revert the code.

jwoertink:
Fixes #902

We've had users in the past that get started with Lucky and have a horrible first experience because their setup script never completes. In this last case, it was only the `verify_connection` task that was failing, but not very apparent. We used to have the code work this way, but changed it to call `open.close`. Somehow, that must be causing some sort of race condition and randomly fails for some people.

I wasn't ever able to recreate the issue, but I'd rather give people a nice first experience, so I'm just reverting this for now.

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -23,7 +23,6 @@ Avram.initialize_logging
 
 Db::Create.new(quiet: true).run_task
 Db::Migrate.new(quiet: true).run_task
-Db::VerifyConnection.new(quiet: true).run_task
 
 Avram::SpecHelper.use_transactional_specs(TestDatabase)
 
--- a/spec/avram/tasks/db_verify_connection_spec.cr
+++ b/spec/avram/tasks/db_verify_connection_spec.cr
@@ -1,11 +1,18 @@
 require "../../spec_helper"
 
 describe Db::VerifyConnection do
+  it "runs a test connection" do
+    task = Db::VerifyConnection.new
+    task.output = IO::Memory.new
+    task.run_task
+    task.output.to_s.should contain "Connection verified"
+  end
+
   it "throws a helpful error" do
     creds = Avram::Credentials.parse("postgres://eat@joes/crab_shack")
     TestDatabase.temp_config(credentials: creds) do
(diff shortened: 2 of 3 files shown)


spec/spec_helper.cr
```diff
@@ -23,7 +23,6 @@ Avram.initialize_logging
 
 Db::Create.new(quiet: true).run_task
 Db::Migrate.new(quiet: true).run_task
-Db::VerifyConnection.new(quiet: true).run_task
```
jwoertink, line 26: This isn't needed. We test this both in the spec, and through the integration test script on CI.

PR #996, opened 2023-12-29
Use DB::Serializable instead of DB.mapping.

### LINKED ISSUE #364: Investigate moving away from DB.mapping

jwoertink:
Fixes #364

The next update for CrystalDB officially deprecates the use of `DB.mapping` https://github.com/crystal-lang/crystal-db/pull/196 

I've labeled this as "Breaking" because if you have any instance variables in your model set by you, you'll have to add the ignore annotation otherwise your app will most likely not compile. However, I don't think it'll be a major breaking change. I was able to get all specs to pass without altering any of them.

The only thing I think may be needed would be a way to add custom converters somehow if you have a custom type, but we do have custom types in the specs, and those work without it, so that could be something added later if someone runs in to an issue.


src/avram/model.cr
```diff
@@ -257,6 +225,21 @@ abstract class Avram::Model
       {% value = nil %}
     {% end %}
 
+    @[DB::Field(
+      ignore: false,
+      key: {{ type_declaration.var }},
+      {% if is_serialized %}
+      converter: JSONConverter({{ data_type }}),
+      {% elsif data_type.id == Array(Float64).id %}
+      converter: PG::NumericArrayFloatConverter,
```
jwoertink, line 234: I'm not sure why this has to be converted now when it didn't need to be with `DB.mapping` :man_shrugging:

src/avram/model.cr
```diff
@@ -8,17 +8,11 @@ abstract class Avram::Model
   include Avram::Polymorphic
   include Avram::SchemaEnforcer
 
-  module JSONConverter(T)
-    def self.from_rs(rs : DB::ResultSet)
-      value = rs.read(JSON::PullParser?)
```
jwoertink, line 13: This just moved to its own file so the `DB::Field` annotation could access it without being namespaced to `Avram::Model`

src/avram/model.cr
mdwagner: It looks like this logic was inside the `setup_db_mapping` macro before

mdwagner (approved):

PR #997, opened 2023-12-29
Remove use of shell in Process.run

### LINKED ISSUE #440: Issues with database config values containing shell special characters
This is a follow-on from #423.

Avram runs `createdb` and `dropdb` commands by string-concatting the config values into a command line, which is passed to `Process.run` with `shell: true`. This means the presence of shell special characters in these values can result in crashes or unintended consequences, e.g. creating/dropping a database other than the one specified.

Examples (these use the database name, but the issue applies to all config values):

* Naming your database `cora"` will crash with the error `--: 1: --: Syntax error: Unterminated quoted string`
* Naming it `good&true` will appear to work, but instead create/drop a database named `good` (with no environment suffix)
* Naming it `#blessed` and creating it will appear to work, but instead create a database with the same name as the current user (the default if no database name is given to `createdb`)

Taking a glance over [the code that calls `Process.run`](https://github.com/luckyframework/avram/blob/998ee826d7674f04ec683d5f9f63395e5b3afadf/src/avram/migrator/runner.cr#L119), it's unclear to me why `shell: true` would be required here, and it seems like it could be easily rewritten without:

```crystal
# using shell
Process.run("createdb -U user -h hostname ...", shell: true)

# using argv
Process.run("createdb", ["-U", "user", "-h", "hostname", ...])
```

But if `shell: true` is necessary for reasons I'm not seeing, each argument could be escaped using [`Process.quote`](https://crystal-lang.org/api/0.35.1/Process.html#quote(args:Enumerable(String)):String-class-method).

jwoertink:
Fixes #440

Now that we removed the pg client tools requirement, the only 2 commands that still require external programs are `restore_db` and `dump_db`. If you used either of these, and your DB config had special shell characters, things could get a little wonky.

This PR adds a bit of safety by removing the `shell: true`, and pushing the args in separately. It also cleans up the shell commands to not use `<` , `>`, and `>>` and just use `File.open`.

### DIFF
--- a/src/avram/migrator/runner.cr
+++ b/src/avram/migrator/runner.cr
@@ -46,6 +46,28 @@ class Avram::Migrator::Runner
     end
   end
 
+  # Returns the DB connection args used
+  # for postgres in an array so you can pass
+  # them to `Process.run`
+  def self.cmd_args_array : Array(String)
+    args = [] of String
+    if user = self.db_user
+      args << "-U"
+      args << user
+    end
+    if host = self.db_host
+      args << "-h"
+      args << host
+    end
+    if port = self.db_port
+      args << "-p"
+      args << port.to_s
+    end
+
+    args << self.db_name
+    args
+  end
+
   def self.drop_db(quiet? : Bool = false)
     DB.connect("#{credentials.connection_string}/#{db_user}") do |db|
       db.exec "DROP DATABASE IF EXISTS #{db_name}"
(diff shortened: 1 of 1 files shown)


src/avram/migrator/runner.cr
```diff
@@ -127,13 +132,17 @@ class Avram::Migrator::Runner
     SQL
   end
 
-  def self.run(command : String, output : IO = STDOUT)
+  def self.run(command : String, output : IO = STDOUT, input : Process::Stdio = Process::Redirect::Close)
+    program, *args = command.split(' ')
```
jwoertink, line 136: I don't think you'd ever have spaces in any of these, but just in case, I'm going to move these over to passing in an array of args directly.

PR #1004, opened 2024-02-11
Add materialized view support helpers.

### LINKED ISSUE #842: Add macro helpers + documentation for supporting materialized views
Per chat in discord:

Current approach:

```
class ModelName < BaseModel
  skip_schema_enforcer

  table do
    ...
  end
```

Ideas:

```
view :materialized do
end
```

jwoertink:
Fixes #842

Avram already supports views and materialized views by way of writing custom SQL for the migration, and specifying `view` in your model.

This PR gives you an easier way to see when a view model is materialized, as well as a handy helper method to refresh the materialized view.

```crystal
class Report < BaseModel
  view materialized: true do
    column count : Int64
  end
end

# refreshes the materialized view
ReportQuery.refresh_view
```


src/avram/base_query_template.cr
```diff
@@ -14,6 +14,12 @@ class Avram::BaseQueryTemplate
         end
       end
 
+      {% if named_args[:materialized_view] %}
+      def self.refresh_view(*, concurrent : Bool = false)
+        {{ type }}.database.exec("REFRESH MATERIALIZED VIEW #{concurrent ? "CONCURRENTLY" : ""} #{{{ type }}.table_name}")
+      end
```
jwoertink, line 20: If you don't pass in this arg, should I still define the method, but raise an exception that tells you the model isn't defined as a materialzed view? :thinking:

spec/support/models/name.cr
```diff
@@ -0,0 +1,5 @@
+class Name < BaseModel
+  view :all_the_names, materialized: true do
+    column name : String
+  end
```
jwoertink, line 4: These technically fail on the SchemaEnforcer... Should materialized views call `skip_schema_enforcer` by default, or would it be better to figure out why they fail?

PR #1006, opened 2024-02-24
Adding new base_query_class arg for associations.

### LINKED ISSUE #85: Allow specifying conditionals in associations

jwoertink:
Fixes #85

This PR adds a new arg that allows you to set the query class used for querying your association. This can be used in conjunction with soft_deleted records to ensure every preload you do only preloads the kept records.

```crystal
class User < BaseModel
  table do
    has_many transactions : Transaction, base_query_class: TransactionQuery
    has_many follows : Follow, foreign_key: :followee_id, base_query_class: FollowQuery
    has_many followers : User, through: [:follows, :follower]
  end
end

class TransactionQuery < Transaction::BaseQuery
  include Avram::SoftDelete::Query

  def initialize
    defaults &.only_kept
  end
end

# This only preloads the kept records
UserQuery.new.preload_transactions
```

The default base_query_class is always `T::BaseQuery`, but with soft_deleted models, you would have to either monkey-patch `T::BaseQuery`, or skip using preloads to avoid cases where you would preload a soft deleted record.

Currently this PR is only adding in `has_many` and `has_many :through` associations. If this works out, we can work out `has_one`, but I'm not sure if `belongs_to` would need this or not :thinking: 

### Side note

There is an edge case that comes up with this. Each association has an extra `_count` method that returns the total count of that association. It uses the `base_query_class` to run a `select_count`. If you use the escape hatch to preload more records than the base query uses, your `_count` method will return more or less than what you've preloaded.

```crystal
u = UserQuery.new.preload_transactions.first
u.transactions_count #=> 10
u.transactions.size #=> 10

# preload ALL transactions, even soft deleted ones
u = UserQuery.new.preload_transactions(Transaction::BaseQuery.new)first
u.transactions_count #=> 10
u.transactions.size #=> 20
```

### Side note 2

How the joins are built when using "has many through" is a bit complex. You generally preload your has_many query, but then that one needs to preload the join internally. If you need an escape hatch out of that default setup, I've added some new args to the preload methods here.

```crystal
# This preload followers and internally the kept follow records
UserQuery.new.preload_followers

# If you need the escape hatch, you have to pass in the base query
# This will now preload all records both kept and soft deleted
UserQuery.new.preload_followers(User::BaseQuery.new, through: Follow::BaseQuery.new)
```

### DIFF
--- a/spec/avram/queryable_spec.cr
+++ b/spec/avram/queryable_spec.cr
@@ -1247,7 +1247,7 @@ describe Avram::Queryable do
       UserQuery.new.select_count.should eq 10
       # NOTE: we don't test rows_affected here because this isn't
       # available with a truncate statement
-      UserQuery.truncate
+      UserQuery.truncate(cascade: true)
       UserQuery.new.select_count.should eq 0
     end
 
--- a/db/migrations/20240224173636_create_transactions.cr
+++ b/db/migrations/20240224173636_create_transactions.cr
@@ -1,0 +1,15 @@
+class CreateTransactions::V20240224173636 < Avram::Migrator::Migration::V1
+  def migrate
+    create table_for(Transaction) do
+      primary_key id : Int64
+      add_timestamps
+      add type : Int32
+      add soft_deleted_at : Time?
+      add_belongs_to user : User, on_delete: :cascade
+    end
+  end
+
+  def rollback
+    drop table_for(Transaction)
+  end
+end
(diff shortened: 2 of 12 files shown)


spec/avram/queryable_spec.cr
```diff
@@ -1247,7 +1247,7 @@ describe Avram::Queryable do
       UserQuery.new.select_count.should eq 10
       # NOTE: we don't test rows_affected here because this isn't
       # available with a truncate statement
-      UserQuery.truncate
+      UserQuery.truncate(cascade: true)
```
jwoertink, line 1250: This had to change because I added the `has_many transactions` to User, and now it was necessary.

PR #1009, opened 2024-03-10
Adds in Array enums.

### LINKED ISSUE #940: Enum array
Is it possible to use an Enum array somehow? This doesn't work:
```crystal
enum MyEnum
  First
  Second
end

class MyModel < BaseModel
  table do
    column enum_array : Array(MyEnum)
  end
end

### PR DESCRIPTION
**Pull request by jwoertink**
Fixes #940

This PR adds the ability to set an array column of enums. It also adds in the ability to specify a custom column converter if you're working with custom types that aren't built in to Avram.

```crystal
class Post < BaseModel
  enum Reaction
    Like
    Heart
    Shocked
    Unlike
  end

  table do
    column reactions : Array(Post::Reaction) = [] of Post::Reaction, converter: PG::EnumArrayConverter(Post::Reaction)
  end
end
```


src/avram/model.cr
```diff
@@ -233,6 +233,8 @@ abstract class Avram::Model
       converter: JSONConverter({{ data_type }}),
       {% elsif data_type.id == Array(Float64).id %}
       converter: PG::NumericArrayFloatConverter,
+      {% elsif converter %}
```
jwoertink, line 236: Originally I was going to use 

```crystal
{% elsif data_type.resolve < Enum %}
```

This would let us pick up the type and set the converter automatically; however, once I did this, it starts to resolve ALL the types. This means that in the `comment` model where we have `belongs_to post : Post`, the `Post` is undefined because it's resolved *before* crystal requires the `post.cr` file. It would end up being a breaking change since any model that references another later in alphabetical order would now have to be required in that file. (or you'd just require each model one by one in the proper order)

src/avram/charms/enum_extensions.cr
```diff
@@ -28,10 +28,37 @@ abstract struct Enum
...
+    end
+
+    # def parse(values : Array(String))
+    #   results = values.map {|s| parse(s)}
+    #   if results.all?(SuccessfulCast)
+    #     parse(results.map(&.value.as(T)))
+    #   else
+    #     FailedCast.new
+    #   end
+    # end
```
jwoertink, line 51: oops. Forgot to remove :joy:  This is handled in `Type`, so it's not needed here.

src/avram/charms/enum_extensions.cr
```diff
@@ -28,10 +28,37 @@ abstract struct Enum
...
+
     def parse(value : T)
       SuccessfulCast.new(value)
     end
 
+    def to_db(values : Array(T))
+      encoded = values.map { |value| to_db(value) }.as(Array(String))
```
jwoertink, line 58: I actually don't know why this needs to be String.... but if you look at the method below, it's also converting the enum value to a string. Maybe a `PQ` thing? :thinking:

PR #1010, opened 2024-03-10
More criteria functions

jwoertink:
I was having a little fun and decided to add just a few more I thought may be helpful.

This adds a few extra Criteria functions

* LENGTH()
* REVERSE()
* ABS()
* CEIL()
* FLOOR()

```crystal
# Find all users that have a username length of 6 characters
UserQuery.new.username.length.eq(6)

# Find all users where the secret in reverse is equal to "terces"
UserQuery.new.secret.reverse.eq("terces")

# Find all transactions where the abs amount is 400 (i.e. matches 400 and -400)
TransactionQuery.new.amount.abs.eq(400)

# Find all purchases where the amount rounded up is equal to 40.0
PurchaseQuery.new.amount.ceil.eq(40.0)

# Find all purchases where the amount rounded down is equal to 30.0
PurchaseQuery.new.amount.floor.eq(30.0)
```

### DIFF
--- a/src/avram/charms/float64_extensions.cr
+++ b/src/avram/charms/float64_extensions.cr
@@ -72,6 +72,7 @@ struct Float64
 
       define_function_criteria(abs, V)
       define_function_criteria(ceil, V)
+      define_function_criteria(floor, V)
     end
   end
 end
--- a/src/avram/charms/float64_extensions.cr
+++ b/src/avram/charms/float64_extensions.cr
@@ -71,6 +71,7 @@ struct Float64
       end
 
       define_function_criteria(abs, V)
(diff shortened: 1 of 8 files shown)


src/avram/charms/float64_extensions.cr
```diff
@@ -69,6 +69,10 @@ struct Float64
       def select_sum! : Float64
         select_sum || 0_f64
       end
+
+      define_function_criteria(abs, V)
+      define_function_criteria(ceil, V)
+      define_function_criteria(floor, V)
```
jwoertink, line 75: Although in postgres the value looks like an integer when using these, the type is `numeric` which maps to Float64 in Crystal.

Issue #968, opened 2023-09-07
Look in to changing how array.includes() criteria method works


jwoertink:
https://github.com/luckyframework/avram/blob/f84e111a80d475611eba5b1a082ca8d8d8df399e/src/avram/criteria_extensions/includes_criteria.cr#L3-L7

Using this method in a query is handy. If one of your columns is an `Array(?)` column, you can get records where that array contains some value

```crystal
# SELECT * FROM users WHERE 'lucky' = ANY(tags)
UseryQuery.new.tags.includes("lucky")
```

However, I just learned that if you put a `GIN` index on `tags`, postgres won't actually use the index in this way...

Using a GIN index
```
=# CREATE INDEX users_tags_index ON users USING gin (tags);
```

Query as-is now
```
=# EXPLAIN SELECT * FROM users WHERE 'lucky' = ANY(tags);
                                    QUERY PLAN                                     
-----------------------------------------------------------------------------------
 Seq Scan on users  (cost=0.00..5748.84 rows=208 width=10)
   Filter: ('lucky'::text = ANY (tags))
(2 rows)

=# SELECT * FROM users WHERE 'lucky' = ANY(tags);
Time: 8.938 ms
```

But there seems to be an alternate way of doing this query which is a ton faster

```
=# EXPLAIN SELECT * FROM users WHERE tags @> '{"lucky"}';
                                         QUERY PLAN                                          
---------------------------------------------------------------------------------------------
 Bitmap Heap Scan on users  (cost=9.61..697.01 rows=208 width=10)
   Recheck Cond: (tags @> '{lucky}'::text[])
   ->  Bitmap Index Scan on users_tags_index  (cost=0.00..9.56 rows=208 width=0)
         Index Cond: (tags @> '{lucky}'::text[])
(4 rows)

=# SELECT * FROM users WHERE tags @> '{"lucky"}';
Time: 0.533 ms
```

This is a pretty big improvement. Though, I should note that without an index, the first way queries twice as fast as the second way...

```
=# SELECT * FROM users WHERE 'lucky' = ANY(tags);
Time: 10.897 ms
=# SELECT * FROM users WHERE tags @> '{"lucky"}';
Time: 23.410 ms
```

jwoertink:
I like the idea of adding some annotations to sort of configure your setup. How would that work here?

Maybe something in your AppDatabase? :thinking: 

```crystal
@[Avram::IncludesCriteria(use_any: false)]
class AppDatabase < Avram::Database
end
[...]

PR #1016, opened 2024-03-23
Adds 2 new JSON critera query methods

### LINKED ISSUE #197: jsonb special query operators

jwoertink:
Fixes #197 

This PR adds 2 more query methods following #1015

* `includes` which performs `WHERE jsonb @> '{"some":"json"}'`
* `in` which performs `WHERE jsonb <@ '{"some":"json"}'`

Also while I was in there I decided to rename the previously added Where classes from `JSON..` to `Jsonb...`. While reading the class names that seemed to make it a little more clear for me.

```crystal
class User < BaseModel
  table do
    column preferences : JSON::Any
  end
end

# query where preferences includes this json in it
UserQuery.new.preferences.includes(JSON::Any.new({"theme" => JSON::Any.new("dark")}))

# query where preferences is included in this json
UserQuery.new.preferences.in(JSON::Any.new({"theme" => JSON::Any.new("dark")}))
```

### DIFF
--- a/src/avram/charms/json_extensions.cr
+++ b/src/avram/charms/json_extensions.cr
@@ -92,13 +92,13 @@ struct JSON::Any
       end
 
       # performs `WHERE jsonb @> other_json`
-      def includes(other_json : JSON::Any)
-        add_clause(Avram::Where::JsonbIncludes.new(column, other_json.to_json))
+      def includes(value) : T
+        add_clause(Avram::Where::JsonbIncludes.new(column, value.to_json))
       end
 
       # performs `WHERE jsonb <@ other_json`
-      def in(other_json : JSON::Any)
-        add_clause(Avram::Where::JsonbIn.new(column, other_json.to_json))
+      def in(value) : T
+        add_clause(Avram::Where::JsonbIn.new(column, value.to_json))
       end
     end
   end
--- a/src/avram/charms/json_extensions.cr
+++ b/src/avram/charms/json_extensions.cr
@@ -96,6 +96,7 @@ struct JSON::Any
         add_clause(Avram::Where::JsonbIncludes.new(column, other_json.to_json))
       end
 
+      # performs `WHERE jsonb <@ other_json`
       def in(other_json : JSON::Any)
         add_clause(Avram::Where::JsonbIn.new(column, other_json.to_json))
       end
--- a/src/avram/charms/json_extensions.cr
+++ b/src/avram/charms/json_extensions.cr
@@ -78,17 +78,17 @@ struct JSON::Any
     class Criteria(T, V) < Avram::Criteria(T, V)
       # performs `WHERE jsonb ? string`
       def has_key(value : String) : T
-        add_clause(Avram::Where::JSONHasKey.new(column, value))
+        add_clause(Avram::Where::JsonbHasKey.new(column, value))
       end
 
       # performs `WHERE jsonb ?| array`
       def has_any_keys(keys : Array(String)) : T
-        add_clause(Avram::Where::JSONHasAnyKeys.new(column, keys))
+        add_clause(Avram::Where::JsonbHasAnyKeys.new(column, keys))
       end
 
       # performs `WHERE jsonb ?& array`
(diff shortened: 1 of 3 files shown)


src/avram/charms/json_extensions.cr
```diff
@@ -78,17 +78,27 @@ struct JSON::Any
...
-        add_clause(Avram::Where::JSONHasAllKeys.new(column, keys))
+        add_clause(Avram::Where::JsonbHasAllKeys.new(column, keys))
+      end
+
+      # performs `WHERE jsonb @> other_json`
+      def includes(other_json : JSON::Any)
+        add_clause(Avram::Where::JsonbIncludes.new(column, other_json.to_json))
+      end
+
+      # performs `WHERE jsonb <@ other_json`
```
robacarp, line 99: Do you think it's possible to add an overload of this method that allows passing in some less verbose type? JSON::Any is so incredibly verbose.

- `def in(other_json : NamedTuple(String, String))` 
- `def in(other_json : Hash(String, String))` 
- maybe even `def in(other_json : T)` -- and do some logic to stringify keys and values?

jwoertink: Totally agree that `JSON::Any` is very verbose. It's also very difficult to construct in many cases without taking a performance hitting shortcut. However, postgres needs the value to look like JSON, so the structure just has to be able to fit in to that. I guess what I could do is leave it untyped, then just call `to_json`. Though, this just means it won't catch some errors at compile-time.

I would say just do `Hash` or `NamedTuple` explicitly, but if you have nested objects that need to be checked, then that wouldn't work. For example

```
# {"user": {"username": "lucky", "preferences": {"theme": "dark"}}}
# This would work with my current change
UserQuery.new.preferences.in(params.from_json)
```

But if we changed this to say `Hash(String, String)` or `NamedTuple`, then the above wouldn't work.

robacarp: It would be nice to be able to specify a duck-type here, wouldn't it?

jwoertink: Any suggestions on what that might look like? It would have to be like `in(value : HasToJsonMethod)` :joy:  though... I think `Object` has `to_json` :thinking:  but that may be too broad. The other thing is that your `jsonb` field may be a json array like `[{"updated": "..."}, {"updated": "..."}]`, so this value would need to be any number of things like JSON::Any, Hash, Array(Hash), NamedTuple, Array(NamedTuple). Can we do `in(value : TellDevinToJustAssumeWhatImThinking)` yet? :rofl:

robacarp: The latter, yes. 😆

src/avram/where.cr
```diff
@@ -245,71 +245,119 @@ module Avram::Where
...
+
+    def prepare(placeholder_supplier : Proc(String)) : String
+      "NOT(#{column} #{operator} #{placeholder_supplier.call})"
+    end
+  end
+
+  class JsonbIncludes < ValueHoldingSqlClause
+    def operator : String
+      "@>"
+    end
```
jwoertink, line 323: One thing I was just thinking about... This operator can be used on Arrays too. [Ref](https://github.com/luckyframework/avram/issues/968)... If this just became the new `Includes` then it would work for both Arrays and Jsonb, but that also means that the old Includes using `val = ANY(col)` would not be an option without some sort of flag or rename or something...

robacarp: If they're equivalent operators, I see no reason not to simplify.

PR #1018, opened 2024-03-30
Added new `Undefined` constant

### LINKED ISSUE #1017: Create a global constant to represent `Avram::Nothing.new`
For every column you define in your app there will be an associated `col : Avram::Nothing = Avram::Nothing.new`.

This does 2 things

1. Creates a ton of instances in memory (it's an empty class, so super minimal)
2. Creates a mess of SaveOperation overloads.

We can't really fix the second point, but maybe by creating a single constant that represents this, it makes that "mess" smaller?

I'm thinking something simple like

```crystal
Undefined = Avram::Nothing.new
``` 

It's 9 characters long instead of 18, and it would only allocate 1 instance instead of one for each column. The only thing I'm hesitant about here is that I didn't want to scope it to Avram. Though, `Avram::IGNORE` is still shorter, it's not shorter by very much. Maybe `Avram::SKIP`?

jwoertink:
Fixes #1017

This PR adds a single `Avram::Nothing.new` instance assigned to a new constant called `IGNORE`. This does 2 things

1. Uses a single allocated instance everywhere to represent `Avram::Nothing` instead of a new instance for every column. While this doesn't give a ton of memory savings, it's still better
2. For those nasty overload errors, it shrinks them up by a few characters.  

Before:
image

After:
image

### DIFF
--- a/src/avram/nothing.cr
+++ b/src/avram/nothing.cr
@@ -4,3 +4,12 @@
 # :nodoc:
 class Avram::Nothing
 end
+
+# Use this value when you want to ignore updating a column
+# in a SaveOperation, for example.
+# ```
+# # Sets value to x.value or ignores it if x.value is nil
+# SaveThing.create!(
+#   value: x.value || Undefined
+# )
+Undefined = Avram::Nothing.new
--- a/src/avram/nothing.cr
+++ b/src/avram/nothing.cr
@@ -6,10 +6,12 @@ class Avram::Nothing
 end
 
 # Use this value when you want to ignore updating a column
-# in a SaveOperation, for example.
+# in a SaveOperation instead of setting the column value to `nil`.
+#
 # ```
-# # Sets value to x.value or ignores it if x.value is nil
+# # Set value to x.value or ignore it if x.value is nil
 # SaveThing.create!(
-#   value: x.value || Undefined
+#   value: x.value || IGNORE
 # )
-Undefined = Avram::Nothing.new
+# ```
+IGNORE = Avram::Nothing.new
--- a/src/avram.cr
+++ b/src/avram.cr
@@ -12,6 +12,7 @@ require "cadmium_transliterator"
 
 require "./ext/db/*"
 require "./ext/pg/*"
+require "./avram/nothing"
 require "./avram/object_extensions"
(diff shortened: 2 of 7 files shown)


src/avram/nothing.cr
```diff
@@ -4,3 +4,12 @@
...
 end
+
+# Use this value when you want to ignore updating a column
+# in a SaveOperation, for example.
+# ```
+# # Sets value to x.value or ignores it if x.value is nil
+# SaveThing.create!(
+#   value: x.value || Undefined
+# )
+Undefined = Avram::Nothing.new
```
robacarp, line 15: I've been spinning on this for a few weeks.

- This is top level global namespace, so if some parser library declares Undefined, it might get weird. Similar to how [monkeypatches are handled](https://github.com/luckyframework/avram/blob/main/src/ext/db/connection.cr)
- The overload error message problem means I want this to be as terse as possible -- I considered proposing `Av::No` but there's really no reason why any single set of characters should be used here.

I don't have any material objection to this, but I don't know about the collision possibilities.

jwoertink: Yup, same. It may be *too* generic to be in the global space. And maybe my dreams of making this shorter is a sunk cost fallacy anyway? `Avram::IGNORE` is still 5 character shorter than `Avram::Nothing.new` so we would still get the benefit of a single instance in memory. It shouldn't really be too much shorter. Or `Avram::SKIP` which is even shorter.

robacarp: Avram::SKIP or IGNORE has the benefit of being more transparent, which I appreciate. I guess I'm considering a specific scenario in my mind when I say that, which is assembling an update piecewise. Maybe there is another scenario where a constant like this might be used which I'm forgetting about.

There'd be no harm in having both SKIP and IGNORE, I suppose, or some other name. 🤷‍♂️ 

I think cleaning up the stack trace is a secondary concern. It would be nice if those were readable in the slightest, but it's okay that they aren't and I don't really think that's on Avram.

jwoertink: I'll give `IGNORE` a try. I think you're right that it is a bit more transparent.

robacarp (approved):
🚀

Issue #353, opened 2020-04-24
Maybe add postgres full text search helpers


paulcsmith:
https://pganalyze.com/blog/full-text-search-ruby-rails-postgres

Sevensidedmarble:
I would be very interested in contributing/working together with someone on this if there's anyone else really into this.

jwoertink:
@Sevensidedmarble I'd love to see this too. The main things when contributing are that it provides a solution in a type safe way, and also allows for an escape hatch for people to get a little crazy. (e.g. `ThingQuery.new.name("")` VS `ThingQuery.new.where("name = ?")`).

I haven't put any thoughts in to this feature yet, but if you'd like to propose an API in this thread, we can sort of work out what may or may not work..

Sevensidedmarble:
Is there anything anyone absolutely doesn't like about the implementation in the [pg_search gem](https://github.com/Casecommons/pg_search#multi-search)? That would be a good starting point for API ideas.

Personally I like their API ideas but I don't like how they've split up 'search scopes' vs 'multi search'. It'd be nice to just have one unified way of defining a document. But maybe I'm biased because my use case is multi-table.

Sevensidedmarble:
I also love the approach https://pganalyze.com/blog/full-text-search-ruby-rails-postgres takes by using generated columns in postgres 12 for single table search. I think that would be a great way to approach searching on a single table here.

PR #1019, opened 2024-03-30
Giving the tsvector and tsquery stuff a shot.

jwoertink:
Ref #353

This adds in a new `match` method along with `to_tsquery` and `to_tsvector` to do a bit of full text searching. The implementation technically generates valid sql, but is this "good enough" for a start? :man_shrugging: 

Before I merge this in, I want to get some feedback and see if this approach will be ok for built-in, or if this will require a lot more planning.

### DIFF
--- a/spec/avram/string_criteria_spec.cr
+++ b/spec/avram/string_criteria_spec.cr
@@ -64,6 +64,12 @@ describe String::Lucky::Criteria do
       name.reverse.eq("revir").to_sql.should eq ["SELECT #{QueryMe::COLUMN_SQL} FROM users WHERE REVERSE(users.name) = $1", "revir"]
     end
   end
+
+  describe "match" do
+    it "uses @@" do
+      name.to_tsvector.match("jeb").to_sql.should eq ["SELECT #{QueryMe::COLUMN_SQL} FROM users WHERE TO_TSVECTOR(users.name) @@ TO_TSQUERY($1)", "jeb"]
+    end
+  end
 end
 
 private def name
--- a/src/avram/charms/string_extensions.cr
+++ b/src/avram/charms/string_extensions.cr
@@ -38,11 +38,17 @@ class String
         add_clause(Avram::Where::Ilike.new(column, value))
       end
 
+      def match(value : String) : T
+        add_clause(Avram::Where::TsMatch.new(column, value))
+      end
+
       define_function_criteria(upper, V)
       define_function_criteria(lower, V)
       define_function_criteria(trim, String)
       define_function_criteria(length, Int64)
       define_function_criteria(reverse, String)
+      define_function_criteria(to_tsvector, String)
+      define_function_criteria(to_tsquery, String)
     end
   end
 end
--- a/src/avram/where.cr
+++ b/src/avram/where.cr
@@ -365,6 +365,34 @@ module Avram::Where
     end
   end
 
+  class TsMatch < ValueHoldingSqlClause
+    def operator : String
+      "@@"
+    end
+
+    def negated : TsNotMatch
+      TsNotMatch.new(column, value)
+    end
+
+    def prepare(placeholder_supplier : Proc(String)) : String
+      "#{column} #{operator} TO_TSQUERY(#{placeholder_supplier.call})"
+    end
+  end
+
+  class TsNotMatch < ValueHoldingSqlClause
+    def operator : String
+      "@@"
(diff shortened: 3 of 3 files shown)


spec/avram/string_criteria_spec.cr
```diff
@@ -64,6 +64,12 @@ describe String::Lucky::Criteria do
       name.reverse.eq("revir").to_sql.should eq ["SELECT #{QueryMe::COLUMN_SQL} FROM users WHERE REVERSE(users.name) = $1", "revir"]
     end
   end
+
+  describe "match" do
+    it "uses @@" do
+      name.to_tsvector.match("jeb").to_sql.should eq ["SELECT #{QueryMe::COLUMN_SQL} FROM users WHERE TO_TSVECTOR(users.name) @@ TO_TSQUERY($1)", "jeb"]
+    end
```
robacarp, line 71: I think you're onto something with this@ It's a nice syntax, and if I'm reading it right I think it means this'll allow me to specify a column type as a tsvec too, which seems necessary to get this to provide reasonably fast searching.

Postgres provides a few variations on to_tsvector that do things a little differently when it comes to operators and escaping. I ended up using `websearch_to_tsquery`, and that may be what the intent here is -- it provides for correctly using boolean, unary, negation, and quotation operators as you might expect a web search to do. Maybe there's room to implement all of this function field or maybe Avram should just pick one? I'm not really sure.

The other thing that seems pretty important is the ts_rank stuff which allows you to actually sort results by relevance. ts_rank + websearch_to_tsquery got me a pretty usable search interface quickly.

jwoertink: I could add `websearch_to_tsquery` easily in the same manner here. I'll have to look at the ranking stuff, that does seem pretty important. The extra options are where I got a little concerned. I *may* be able to add those in, but I wasn't sure how often they're really used. Like, if it's a very small % of people that need the extra options like `to_tsquery('english', 'whatever')` then maybe they just fall back to raw SQL again.

The other thing here is I'm making a huge assumption that the right side of `@@` will always be cast to tsquery although postgres does allow a few variations. It's just that we don't have a way to map to a Crystal side TsQuery object, so I have to make some assumption on it. My main hope is that this is good enough for the 80% until someone comes along and ports [pg_search](https://github.com/Casecommons/pg_search) to crystal :smile:

robacarp: I agree that the parametric index (`ts_vector('english'...)`) is out of reach for this stage. In order to accomplish that I'd probably reach for something more search-domain specific than the general purpose Avram query language.

I've posted this elsewhere, but I'll include it here for posterity.

In a migration and model, I setup a ts_vector column like this:

```crystal
# in the migration:
    execute <<-SQL
      ALTER TABLE channels
      ADD COLUMN search_vector tsvector
      GENERATED ALWAYS AS (
        to_tsvector('english', coalesce(title, '') || ' ' || coalesce(summary, ''))
      ) STORED;
    SQL

# and in the model:
  table :channels do
    column search_vector : String?
  end
```

Then my search query is this:

```crystal
    sql = <<-SQL
      WITH search_results AS (
        SELECT
          channels.id AS channel_id,
          ts_rank_cd(channels.search_vector, websearch_to_tsquery($1)) AS rank
        FROM channels
        WHERE channels.search_vector @@ websearch_to_tsquery($1)
        ORDER BY rank DESC
        LIMIT 25
      )

      SELECT
        channels.*
      FROM search_results
      JOIN channels ON search_results.channel_id = channels.id
    SQL

    AppDatabase.query_all(sql, query_input, as: PoDB::Channel)
```

The ts_rank_cd stuff is "under the hood," I think. I wonder if it would be possible to have a syntax like this: `column.search(query)` which:

- is smart enough to know if the column type is a ts_vector or should be wrapped in a to_tsvector call -- or perhaps issues a warning if it's not?
- wraps query in a websearch_to_tsquery
- runs the match operator
- sets a default order by key

Is that the right thing to do when someone writes that code? Is that even achievable? Is that too much magic?

-----

As an aside, I had to construct this as a CTE because Avram doesn't know what to do with the `rank` column, but I need to SELECT it in order to have postgres calculate the rank in order to sort the results. If Avram could discard the column in the response, it would be unnecessary.

jwoertink: oo nice. That helps seeing the example.

PR #1020, opened 2024-04-10
Updating min Crystal and postgres versions

jwoertink:
While there's no breaking changes back to 1.6.2, I don't think we need to continue supporting the older Crystal versions with 1.12 released, and 1.13 well in the works. 

Since I was in here, I decided to also bump the postgres versions. We've been testing against 12, 13, and 14, but 15 and 16 are both out with 17 in the works.


jwoertink:
ok, so I wasn't expecting that... Turns out `ubuntu-latest` has pg client tools 14 installed. Running against lower versions works fine, but higher versions doesn't work... So the only way we can test this would be to manually install the pg client tools for the same server version installed :frowning_face:

PR #1028, opened 2024-04-21
Replacing Teeplate with LuckyTemplate

### LINKED ISSUE #1026: Remove teeplate
This PR on Lucky https://github.com/luckyframework/lucky/pull/1861 removed Teeplate as a dependency. Avram uses Lucky as a dev dependency for the Lucky extension

I forgot that the generators here
https://github.com/luckyframework/avram/blob/db6ef660d18c378a076bba9d16974f2aa9f69bad/src/lucky/tasks/gen/model.cr#L2

They still use it because Avram uses Lucky v1.0.0 as a dev dependency :man_facepalming: So now I need to remove teeplate from here

jwoertink:
Fixes #1026

This PR is required in order to release v1.2.0 [ref](https://github.com/luckyframework/lucky_cli/actions/runs/8759728427/job/24058332316?pr=860#step:4:15618).

In order to make this work, I needed to update the tasks slightly, but I couldn't make them use the full new format without doing a full rewrite of the tasks.


shard.yml
```diff
@@ -40,6 +40,9 @@ dependencies:
   cadmium_transliterator:
     github: cadmiumcr/transliterator
     branch: master
+  lucky_template:
+    github: luckyframework/lucky_template
+    version: ~> 0.1.0
```
jwoertink, line 45: This can be removed after Lucky 1.2.0 is released.

Issue #1039, opened 2024-05-12
[Migration] Change column type use `change_type`  not change the optional of column?


zw963:
Following is one file script.

```crystal
require "pg"
require "avram"

class AppDatabase < Avram::Database
end

AppDatabase.configure do |settings|
  settings.credentials = Avram::Credentials.new(database: "test_db", username: "postgres", hostname: "localhost")
end

Avram.configure do |settings|
  settings.database_to_migrate = AppDatabase
  settings.lazy_load_enabled = true
end

class CreateFoo::V20240420041991 < Avram::Migrator::Migration::V1
  def migrate
    create table_for(Foo), if_not_exists: true do
      primary_key id : Int64
      add number : Int32?
      add_timestamps
    end
  end

  def rollback
    drop table_for(Foo)
  end
end

class CreateFoo::V20240420041992 < Avram::Migrator::Migration::V1
  def migrate
    alter table_for(Foo) do
      change_type number : Float64, precision: 10, scale: 5
    end
  end

  def rollback
  end
end

Avram::Migrator::Runner.drop_db(true)
Avram::Migrator::Runner.create_db(true)
Avram::Migrator::Runner.new.run_next_migration
Avram::Migrator::Runner.new.run_next_migration

abstract class BaseModel < Avram::Model
  def self.database : Avram::Database.class
    AppDatabase
  end
end

class Foo < BaseModel
  table do
    column number : Float64?
  end
end

Foo::SaveOperation.create!(number: 70884.70074)
Foo::SaveOperation.create!

query = Foo::BaseQuery.new

p! query.select_count # => 2
```

As you can see, i change my `number` column from `Int32?` into `Float64, precision: 10, scale: 5`, but underneath, this column still optional.

Thanks.

davidepaolotua:
I had a quick look at it but I'm kinda perplexed on how to proceed.
I'd honestly add a new statement - like it was done for the default. Three reasons for that:
1. Postgres does not allow for alter column set data .... not null; (and I guess it's a good enough reason)
3. it would be a mess to implement the inversed operation otherwise (not null -> null)

I'll try having a look at it. 
@jwoertink, do u see any reasons not to add a new statement?

jwoertink:
Yeah, this is a tricky one. What are you thinking for a new statement?

jwoertink:
[...]

PR #1041, opened 2024-05-16
Add change nullability statement

davidepaolotua:
This PR attempts to fix two things:

#### provide a way to change the nullability of a column in the migrator.
Related to #1039 
Two new macros have been created:
- allow_nulls_for :column, which allows for the :column to receive null values
- forbid_nulls for :column, which set the not null constraint
it is split from the change_type statement, so that you can theoretically do something like
```
change_type the_column : String
# migrate somehow the null fields present in the table
forbid_nulls_for :the_column
```

#### changes the order of the change_default statement with respect to the change_type statement
Currently, if you had something like:
```
change_type a : String
change_default a : String, default: "Whatever"
```
it would generate first the change of default, then the change of type, which is probably counterintuitive.
Now, it generates first the change of type, then the change of default

### DIFF
--- a/spec/avram/migrator/alter_table_statement_spec.cr
+++ b/spec/avram/migrator/alter_table_statement_spec.cr
@@ -89,6 +89,16 @@ describe Avram::Migrator::AlterTableStatement do
     built.statements[2].should eq "ALTER TABLE ONLY test_defaults ALTER COLUMN money SET DEFAULT '29.99';"
   end
 
+  it "changes defaults after changing the type" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      change_type id : Int64
+      change_default id : Int64, default: 54
+    end
+    built.statements.size.should eq 2
+    built.statements[0].should eq "ALTER TABLE users ALTER COLUMN id SET DATA TYPE bigint;"
+    built.statements[1].should eq "ALTER TABLE ONLY users ALTER COLUMN id SET DEFAULT '54';"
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Nice! Let's give this a shot :rocket:

jwoertink:
@davidepaolotua I was just writing up some documentation and it just occurred to me.... why did we add `allow_nulls_for` and `forbid_nulls` when we have `make_required` and `make_optional` which generate the same SQL? :thinking: 

https://github.com/luckyframework/avram/blob/bec90582181bed458e1c36d0497b8fce3353306f/src/avram/migrator/change_null_statement.cr#L13-L24

https://github.com/luckyframework/avram/blob/bec90582181bed458e1c36d0497b8fce3353306f/src/avram/migrator/alter_table_statement.cr#L104-L106

I'm sure there was some logic behind having a second way to do this, but I'm not seeing why, and I don't remember :sweat_smile:

davidepaolotua:
Uhm... I wish I was as sure as you are... 

imho I just completely missed the existence of that method - I prolly just looked  in the alter_table file, and didn't occur to me that there was actually a separated file only for that case, while the other change_xxx are in alter table ^^' 

we actually could remove those methods imho (just keeping the fix for the ordering of the ddl statements)

jwoertink:
Ok. I'll at least open an issue on it. I'm sure I just forgot about the other methods too at the time :joy:

Issue #1030, opened 2024-04-25
Add && array criteria method


jwoertink:
I don't know what this is called, but it's handy

```
SELECT * FROM users WHERE tags && '{ruby,crystal}'::text[];
```

This returns all users that have either ruby *or* crystal in their `tags` array, but not necessarily both.

It would be like

```crystal
UserQuery.new.tags.has_any(["ruby", "crystal"])
```

jwoertink:
Closed by https://github.com/luckyframework/avram/pull/1042

PR #1042, opened 2024-05-16
Add have any criteria

davidepaolotua:
Implements the feature in #1030 
However, the method was renamed to `have_any?`, instead of `has_any`, just because an array column will prolly be called in the plural form (e.g: `Users.tags`) and have_any just sounded better.

### DIFF
--- a/spec/avram/migrator/alter_table_statement_spec.cr
+++ b/spec/avram/migrator/alter_table_statement_spec.cr
@@ -89,6 +89,27 @@ describe Avram::Migrator::AlterTableStatement do
     built.statements[2].should eq "ALTER TABLE ONLY test_defaults ALTER COLUMN money SET DEFAULT '29.99';"
   end
 
+  it "changes defaults after changing the type" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      change_type id : Int64
+      change_default id : Int64, default: 54
+    end
+    built.statements.size.should eq 2
+    built.statements[0].should eq "ALTER TABLE users ALTER COLUMN id SET DATA TYPE bigint;"
+    built.statements[1].should eq "ALTER TABLE ONLY users ALTER COLUMN id SET DEFAULT '54';"
+  end
+
+  it "can change column nullability" do
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      forbid_nulls_for :id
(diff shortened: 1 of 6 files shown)


spec/avram/migrator/alter_table_statement_spec.cr
```diff
@@ -89,6 +89,27 @@ describe Avram::Migrator::AlterTableStatement do
...
+    built = Avram::Migrator::AlterTableStatement.new(:users).build do
+      forbid_nulls_for :id
+      allow_nulls_for :age
+    end
+
+    built.statements.size.should eq 2
+    built.statements[0].should eq "ALTER TABLE users ALTER COLUMN id SET NOT NULL;"
+    built.statements[1].should eq "ALTER TABLE users ALTER COLUMN age DROP NOT NULL;"
+  end
+
```
jwoertink, line 112: oops! Looks like this snuck in here :joy:

src/avram/criteria.cr
```diff
@@ -154,6 +154,11 @@ class Avram::Criteria(T, V)
     add_clause(Avram::Where::In.new(column, values))
   end
 
+  def have_any?(values) : T
```
jwoertink, line 157: I like the name better than my original idea; however, my only (very very small) concern is would this be confusing that it's a predicate method that doesn't return Bool or denote a nilable value? Maybe it's fine in this case, but I don't want people to confuse it with the `none?` or `any?` [methods](https://github.com/luckyframework/avram/blob/1326a8ea7583b91169993194802bb0875172a9c3/src/avram/queryable.cr#L39-L45) we have.

```crystal
if UserQuery.new.tags.have_any?(["featured"])
  # would always return truethy
end

# would this be strange? lol
UserQuery.new.tags.have_any?(["featured"]).any?
```

But if you don't think it'll be that big of an issue with some proper docs, then I'm cool with it as is :+1:

spec/avram/migrator/alter_table_statement_spec.cr
davidepaolotua: 🤕 did the checkout from the wrong branch 😢

src/avram/criteria.cr
jwoertink: What about just `have_any` without the question mark?

PR #1050, opened 2024-06-22
Added Avram::Join::Raw.

zw963:


src/avram/query_builder.cr
```diff
@@ -318,6 +319,15 @@ class Avram::QueryBuilder
     @joins.uniq(&.to_sql)
   end
 
+  def join(raw_join_clause : Avram::Join::Raw) : self
+    @raw_joins << raw_join_clause
+    self
+  end
+
+  def joins : Array(Avram::Join::Raw)
```
jwoertink, line 327: this method would override the methods on 318 of the same name

jwoertink (changes requested):

jwoertink:
It's a good start! Thanks for taking this on. You're definitely going to need to add a bunch of specs for this. I would suggest to start working through some of those specs to make sure all of the SQL being generated is correct. That may help get through these CI errors.

src/avram/query_builder.cr
zw963: > this method would override the methods on 318 of the same name

I thought `override` is not so accurately, because the return value of both method is different, the issue is, when `joins_sql` invoke `joins` metrhod, it can't distinguish the difference between the two method.

I have no idea about how to fix this, could you please give me a little clue?

zw963:
>I would suggest to start working through some of those specs to make sure all of the SQL being generated is correct. 

Yes, I am trying to see if it is possible to add this new feature but without need change the existing spec. (only add more)

src/avram/query_builder.cr
akadusei: Crystal does not factor in return type when determining a method's signature, at least for overrides. There is no way during method call to determine which of the variants to dispatch to if they only differ in return types. So in this case, yes, the second method will override the first.

How about renaming the method to `#raw_joins`?

zw963: > Crystal does not factor in return type when determining a method's signature, at least for overrides. There is no way during method call to determine which of the variants to dispatch to if they only differ in return types.

I guess it's can?

```crystal
	def hello : String
	  "hello"
	end

	def hello : Int32
	  100
	end

	def say_hello : Int32
	  hello
	end

	p! say_hello # => 100
```

> How about renaming the method to #raw_joins?

I remember i tried it like this, but that #raw_joins method never by invoked some where, there are some(probably macro?) magic happen some where, i didn't found it.

PR #1055, opened 2024-08-10
Run delete operation in a transaction

### LINKED ISSUE #1052: Run delete operation in a transaction
https://github.com/luckyframework/avram/blob/bec90582181bed458e1c36d0497b8fce3353306f/src/avram/delete_operation.cr#L38

I'm wondering if `Avram::DeleteOperation#delete` should use a transaction, like its `SaveOperation#save` counterpart. This should enable an `after_commit` callback counterpart for delete operations.

akadusei:
Closes #1052


jwoertink (approved):
Looks great! :raised_hands:  Thank you

PR #1065, opened 2024-08-31
Adding new validation to validate URL formatted attributes

### LINKED ISSUE #667: Add validator for URLs
Would be awesome to have a built-in validator for handling URLs. I think to start, it would just run through `URI.parse`, and then validate that it has a valid scheme, and hostname. 

This can obviously get a bit tricky since doing relative ftp URLs like `ftp:///upload` could be "valid", but I think it's pretty common to store a standard HTTP url, and having a nice method to validate those would be great.

Maybe something that looks like this?

```crystal
def validates_url_format(
  attribute : Avram::Attribute,
  scheme : String = "https",
  relative : Bool = false,
  message : String = "must be a URL"
)
end
```

jwoertink:
Fixes #667

Adds new `validate_url_format` validator.

### DIFF
--- a/src/avram/validations.cr
+++ b/src/avram/validations.cr
@@ -330,4 +330,20 @@ module Avram::Validations
 
     true
   end
+
+  def validate_url_format(
+    attribute : Avram::Attribute(String),
+    scheme : String = "https",
+    message : Avram::Attribute::ErrorMessage = Avram.settings.i18n_backend.get(:validate_url_format)
+  ) : Bool
+    if url = attribute.value.presence
+      uri = URI.parse(url)
+      if uri.scheme != scheme || uri.host.presence.nil?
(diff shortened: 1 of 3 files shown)


src/avram/validations.cr
```diff
@@ -330,4 +330,20 @@ module Avram::Validations
...
+
+  def validate_url_format(
+    attribute : Avram::Attribute(String),
+    scheme : String = "https",
+    message : Avram::Attribute::ErrorMessage = Avram.settings.i18n_backend.get(:validate_url_format)
+  ) : Bool
+    if url = attribute.value.presence
+      uri = URI.parse(url)
+      if uri.scheme != scheme || uri.host.presence.nil?
+        attribute.add_error(message % scheme)
```
robacarp, line 342: wow this is slick

Issue #1067, opened 2024-09-30
Support Windows natively


jwoertink:
Related: https://github.com/luckyframework/lucky/issues/1746#issuecomment-2381018073

This issue is for full native Windows support. We should be able to add windows to the CI and get full green specs.

jwoertink:
Currently I'm unable to get a test on Windows due to an issue with PG https://github.com/will/crystal-pg/issues/291

jwoertink:
This is now resolved thanks to https://github.com/will/crystal-pg/pull/292

PR #1069, opened 2024-10-12
Windows support

jwoertink:
Ref https://github.com/luckyframework/avram/issues/1067

This is an attempt at adding native Windows support. It was all written on Linux... wish me luck(y) :joy: 

At the moment it seems that we will need to require at least Crystal 1.14 or later for the Windows to work. I've opened an issue on PG https://github.com/will/crystal-pg/issues/291 related to the failure. This may be as close as we can get for now.


.github/workflows/ci.yml
```diff
@@ -29,11 +29,13 @@ jobs:
...
       fail-fast: false
       matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
         shard_file:
           - shard.yml
         postgres_version:
           - 14
-          - 15
```
jwoertink, line 36: No need to check every single version. I think just checking the lowest and the highest versions is fine and assume everything in the middle "just works".

PR #1071, opened 2024-10-31
Testing Windows PG Patch

jwoertink:
Ref: https://github.com/will/crystal-pg/pull/292


jwoertink:
woo! :partying_face:  Specs compile and run on Windows now! Just need to actually fix them :upside_down_face:

PR #1074, opened 2024-11-10
Fixing specs on windows.

### LINKED ISSUE #1072: Specs fail on Windows
Some of the specs fail on Windows because it expects `\r\n` but gets `\n`. We will need to figure out how to handle the line break differences when testing output.

jwoertink:
Fixes #1072

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.cr text eol=lf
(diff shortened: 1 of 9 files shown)


jwoertink:
For some reason I can't get that last spec to pass... It passes locally on my windows machine, but something seems to be different on CI :thinking:

jwoertink:
woot! :partying_face:  all green :green_heart:

PR #1076, opened 2024-12-26
Adding table locking.

### LINKED ISSUE #594: Support locking tables
Hi how i can lock table?

jwoertink:
Fixes #594

This adds some options for doing table and row locks.

```crystal
AppDatabase.with_lock_on(User, mode: :row_exclusive) do
  user = UserQuery.new.id(1).for_update.first
  SaveUser.update!(user, name: "New Name")
end
```

Testing it in a real app, and it seems to work... I'm not sure how to test this other than just ensuring the correct SQL is generated.

### DIFF
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -14,6 +14,7 @@ class Avram::QueryBuilder
   @prepared_statement_placeholder = 0
   @distinct : Bool = false
   @delete : Bool = false
+  @for_update : Bool = false
 
   def initialize(@table)
   end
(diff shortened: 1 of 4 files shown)


src/avram/query_builder.cr
```diff
@@ -158,6 +159,11 @@ class Avram::QueryBuilder
     self
   end
 
+  def for_update : self
+    @for_update = true
```
jwoertink, line 163: Postgres supports more than just `FOR UPDATE` 

> FOR { UPDATE | NO KEY UPDATE | SHARE | KEY SHARE }

So this might need some sort of args, or maybe just different methods like

```crystal
def for_no_key_update
end
def for_share
end
def for_key_share
end

# VS

def for_update(for_method)
end
for_update(:no_key_update)
for_update(:share)
```

Issue #1082, opened 2025-04-09
Crystal 1.16 support


jwoertink:
Looks like variable names with `?` are no longer valid in Crystal 1.16, so we should remove all those. Then whatever formatting changes there are will also need updating.

![Image](https://github.com/user-attachments/assets/1b4c5e00-2edf-4621-ba4b-357dc0207e9f)

jwoertink:
I've removed the good first issue because this is a bit more complex of an issue... After running the Crystal formatter for 1.16, ameba starts failing on a ton of files. After running the `./bin/ameba --fix` it breaks the formatter on Crystal 1.16.

We currently have ameba locked to 1.5.0 because ameba 1.6 doesn't work with Avram due to all of the `Lint/UselessAssign` from macros which ameba doesn't know how to handle.

jwoertink:
Ok, nevermind... Looks like that PR was green. It may have just been my machine then.

jkthorne:
I didnt know crystal had a built in tool for this.  I will have to use this on my projects.

PR #1083, opened 2025-04-11
Support for Crystal 1.16

jwoertink:
Ref #1082


jwoertink:
wow, that all passed?? :raised_eyebrow:  I was sure that was gonna fail... Ok, in that case, I guess this closes #1082

PR #1085, opened 2025-04-27
Fix missing nested param error for nested array params

akadusei:
Even if nested params is not empty, Lucky would still raise `Lucky::MissingNestedParamError` if any array attributes are not set.

You had to fill in the array attribute to avoid the error:

```crystal
response = client.exec(Features::Create, feature: {
  choices: Array(String).new, # Without this, you get "Missing param key: 'feature' (Lucky::MissingNestedParamError)..."
  maximum: 99,
  minimum: 3,
  name: "Number of Kitchens",
  type: :integer
})
```

See <https://github.com/luckyframework/lucky/issues/1949>


jwoertink (approved):
:+1:

PR #1086, opened 2025-04-27
Split callbacks into separate modules

### LINKED ISSUE #973: Running before_delete on a SaveOperation does nothing
Related: https://github.com/luckyframework/avram/issues/898

If you accidentally type `before_delete` instead of `before_save`, your code will compile and run just fine, but it won't run that callback which could lead to weird issues at runtime.

I'd imagine it would be better to raise a compile-time error telling you that you're using the wrong callbacks.

akadusei:
Closes #973

Now you get a compile error when you call `before_delete` in a save operation, for instance:

```
Showing last frame. Use --error-trace for full trace.

In spec/avram/case_insensitive_column_spec.cr:7:3

 7 | before_delete do
     ^------------
Error: undefined local variable or method 'before_delete' for SaveEmailAddress.class
```


jwoertink (commented):
This looks good, but am I missing where the regular operation callbacks are set? 

https://github.com/luckyframework/avram/blob/f366c8dd5ec9c0041d19cd93e85fac1ed0feaea5/src/avram/operation.cr#L2

Seems like this require would need to be update too, right?

akadusei:
> require "./callbacks"

You're right. I'm not even sure why that compiles.

Regular operations used the old `Avram::Callbacks` name, and moved into the `callbacks/` directory. So the require there should be `require "./callbacks/callbacks"` (looks weird). You think I should move it outside the `callbacks/` directory, so it can just be `require "./callbacks"`?

I used `Avram::Callbacks` because the operation is called simply `Avram::Operation`.

jwoertink (approved):

PR #1089, opened 2025-05-06
Allow defining separate reader and writer databases

### LINKED ISSUE #138: Specify read/write databases
Extracted from #9 

```crystal
abstract class BaseModel < Avram::Model
  def database
     PrimaryDatabase
  end


  # Or set separate read/write database
  def read_database
    ReplicaDatabase
  end

  def write_database
    PrimaryDatabase
  end
end
```

This will be super easy to implement except for the specs. Need  to think of a reliable way to do test it

jwoertink:
Fixes #138

This PR adds in two new model methods `read_database` and `write_database`. By default they both point to the same database, so in theory, you would be able to update Avram and everything works like normal. Then when you need to start using a read replica, it becomes as each as just overriding a method.

There are a few things I'm sort of overlooking here as I don't want to get too complicated... The concept of multi-tenant sharding won't exist (yet, at least). I have no clue how we would handle that. Same goes for doing any sort of load balancing. My guess is with it being in a method, that's probably something you could implement if you really needed.

The last bits I really want before I mark this as ready for review would be

* How do I test this?
* How can we do a temp override on the fly?

```crystal
# How would I do this?
Avram.use_database(OtherDB) do
  # uses OtherDB instead of the normal one
  UserQuery.new
end

# uses your normal DB...
UserQuery.new
```

### DIFF
--- a/spec/avram/multi_database_spec.cr
+++ b/spec/avram/multi_database_spec.cr
@@ -10,6 +10,16 @@ private class ModelWithBadDatabase < BaseModel
   end
 end
 
+private class BaseModelWithSeparateReadWrite < Avram::Model
+  def self.read_database : Avram::Database.class
+    TestDatabase
+  end
+
+  def self.write_database : Avram::Database.class
+    TestDatabase
+  end
+end
+
 describe "Configuring and connecting to different databases" do
   it "tries to connect to the configured database" do
     # It will fail to connect which is what we expect since we configured
@@ -21,4 +31,9 @@ describe "Configuring and connecting to different databases" do
       ModelWithBadDatabase::BaseQuery.new.select_count
     end
   end
+
+  # context "when read and write are separate DBs" do
+  #   it "uses the read_database" do
+  #   end
+  # end
 end
--- a/src/avram/base_query_template.cr
+++ b/src/avram/base_query_template.cr
@@ -16,7 +16,7 @@ class Avram::BaseQueryTemplate
 
       {% if named_args[:materialized_view] %}
       def self.refresh_view(*, concurrent : Bool = false)
-        {{ type }}.database.exec("REFRESH MATERIALIZED VIEW #{concurrent ? "CONCURRENTLY" : ""} #{{{ type }}.table_name}")
+        {{ type }}.write_database.exec("REFRESH MATERIALIZED VIEW #{concurrent ? "CONCURRENTLY" : ""} #{{{ type }}.table_name}")
       end
       {% end %}
 
@@ -39,7 +39,7 @@ class Avram::BaseQueryTemplate
           end
         {% end %}
 
-        database.exec(
+        write_database.exec(
           query.statement_for_update(_changes, return_columns: false),
           args: query.args_for_update(_changes)
         ).rows_affected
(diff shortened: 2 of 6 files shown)


jwoertink:
This is failing to compile, but I'm having trouble tracking down where... Something is calling `Model.database` in a way where it's using the abstract class's method and not the overridden method. If I remove the schema enforcer completely, it all compiles and specs (mostly) pass. I'm just not sure why, yet...

jwoertink:
Ah neat... Ok, so in theory, with this PR, I think your code would look moire like this?

```crystal
class BaseModel < Avram::Model
  def self.read_database : Avram::Database.class
    ReadDatabase
  end

  def self.write_database : Avram::Database.class
    if App.use_write_database
        WriteDatabase
      else
        ReadDatabase
      end
  end
end
```

With this PR, I'm using the `write_database` for all writes always. I guess if there was a case where you wanted to write to a different DB depending on the location or whatever, then you'd just need to add that logic in there.

robacarp:
Yes, that makes sense. Fly.io is a little strange in this. They're pretty prescriptive.

They encourage multiple deployments of application nodes across their infrastructure, with each having access to a read-replica of the database. Then, traffic which is write traffic should be directed at application containers which are in the region where the PG primary database is. I suppose write-traffic directed at a read-replica region can still connect to the write-database, but the connection is facilitated on their wireguard backhaul so it'll be slower.

I'm excited to test it out!

akadusei:
> ```crystal
> # How would I do this?
> Avram.use_database(OtherDB) do
>   # uses OtherDB instead of the normal one
>   UserQuery.new
> end
> 
> # uses your normal DB...
> UserQuery.new
> ```

Off the top of my head:

```crystal
module Avram
  protected class_property current_database : Avram::Database.class?

  def self.use_database(database : Avram::Database.class)
    self.current_database = database
    yield
    self.current_database = nil
  end

  module Queryable(T)
    # ...

    delegate :table_name, :primary_key_name, to: T # <= Remove `:database` from here

    def self.database
      Avram.current_database || T.read_database
    end

    # ...
  end
end
```

jwoertink:
```
def self.use_database(database : Avram::Database.class)
    self.current_database = database
    yield
    self.current_database = nil
  end
```

I had the same thought originally or like using Habitat, but in this case it would mean that *all* queries would go to the temp DB until this specific one finished. It's almost like it would need to be patched on the current Fiber instance or something? I'm not sure... 

I'm almost wondering if the whole DB thing should have been dropped down to the instance level instead of using the Class all over the place :thinking:

akadusei:
Yeah, it's probably not safe. Instance-level seems OK:
```crystal
ThingQuery.new(AppDatabase).find(thing_id)
```

It could default to the read database (or write for `#update`) if not passed.

jwoertink:
```
ThingQuery.new(AppDatabase).find(thing_id)
```

This would only work for queries, and not for any operation. Plus if you had the initialize overridden, then you'd no longer have access to set that.

I was more thinking that instead of having

```crystal
def self.database : Avram::Database.class
  #...
end
```

it would become an instance

```crystal
def database : Avram::Database
  #
end
```

This would be a pretty big change though... a breaking one at that. So I think for now maybe what I have here is "good enough" to start. Then I'll open another issue to chat more about how to handle being able to override the selected DB on a per instance level which may include this large refactor.

I appreciate everyone chiming in and helping me work through this :raised_hands:

PR #1090, opened 2025-05-09
Add join method overloads that accept the query condition as argument

### LINKED ISSUE #1088: Deprecate `#where_*` join query methods
Doing anything other than an inner join is verbose:

```crystal
UserQuery.new.left_join_tasks.where_tasks(TaskQuery.new.title("Clean up notes"), auto_inner_join: false)
```

Ideally, calling `#left_join_task` should automatically disable auto inner join. The problem, though, is that nothing stops anyone from calling `#where_tasks` before calling `#left_join_tasks`:

```crystal
UserQuery.new.where_tasks(TaskQuery.new.title("Clean up notes")).left_join_tasks
```

Looking at `Avram::BaseQueryTemplate`, we could add overload to the `#*_join_*` methods that accept the `assoc_query`. This way we can have queries like this:

```crystal
UserQuery.new.left_join_tasks(TaskQuery.new.title("Clean up notes"))
```

If we insist on having `where_tasks` in there (probably because it reads nice?), we can make that the argument name:

```crystal
UserQuery.new.left_join_tasks(where_tasks: TaskQuery.new.title("Clean up notes"))
```

What do you think?

akadusei:
This deprecates `#where_*` join methods to which the query condition was delegated.

Closes #1088

### DIFF
--- a/spec/avram/associations_spec.cr
+++ b/spec/avram/associations_spec.cr
@@ -118,8 +118,8 @@ describe Avram::Model do
         comment = CommentForCustomPost::SaveOperation.create!(body: "bar", post_id: post.id)
         comment.post_with_custom_table.should eq(post)
 
-        CommentForCustomPost::BaseQuery.new.where_post_with_custom_table(PostWithCustomTable::BaseQuery.new.id(post.id)).first.should eq(comment)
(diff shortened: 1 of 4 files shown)


jwoertink (approved):
This definitely looks cleaner. Thanks for adding this!

PR #1091, opened 2025-05-10
Update `updated_at` column in `Avram::BaseQueryTemplate#update`

### LINKED ISSUE #1087: The Query update method doesn't update `updated_at`
This code updates records but doesn't update the `updated_at` field.

```crystal
UserQuery.new.last_active_at.lte(1.year.ago).update(active: false)
```

Somewhere in this mess

https://github.com/luckyframework/avram/blob/23aa237d499aac5c9ca1af317950d036eda42483/src/avram/base_query_template.cr#L29-L45

If you're using triggers in your postgres, then it probably does work, but if you're not, then it just won't update these values. 

Related: https://github.com/luckyframework/avram/issues/393#issuecomment-646807432

akadusei:
Fixes #1087

### DIFF
--- a/spec/avram/queryable_spec.cr
+++ b/spec/avram/queryable_spec.cr
@@ -1369,7 +1369,7 @@ describe Avram::Queryable do
       manager2.employees!.first.name.should eq("Spongebob Alderson")
     end
 
-    it "updates updated_at column" do
+    it "updates updated_at column if not passed in" do
       PostFactory.create &.title("Old title").updated_at(2.days.ago)
       PostFactory.create &.title("Old title").updated_at(2.days.ago)
 
@@ -1382,6 +1382,24 @@ describe Avram::Queryable do
       results.each do |result|
         result.title.should eq("New title")
         result.updated_at.should be_close(Time.utc, 1.minute)
+      end
+    end
+
+    it "updates updated_at column with value passed in" do
+      new_updated_at = 2.days.ago.at_beginning_of_second
+
+      PostFactory.create &.title("Old title").updated_at(Time.utc)
+      PostFactory.create &.title("Old title").updated_at(Time.utc)
+
+      updated_count = PostQuery.new.update(title: "New title", updated_at: new_updated_at)
+
+      updated_count.should eq(2)
+      results = PostQuery.new.results
+      results.size.should eq(2)
+
+      results.each do |result|
+        result.title.should eq("New title")
+        result.updated_at.should eq(new_updated_at)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


spec/avram/queryable_spec.cr
```diff
@@ -1368,6 +1368,22 @@ describe Avram::Queryable do
...
+
+      updated_count = PostQuery.new.update(title: "New title")
+
+      updated_count.should eq(2)
+      results = PostQuery.new.results
+      results.size.should eq(2)
+
+      results.each do |result|
+        result.title.should eq("New title")
+        result.updated_at.should be_close(Time.utc, 1.minute)
```
jwoertink, line 1384: wow! I didn't know about this `be_close` method :joy:  that's neat

src/avram/base_query_template.cr
```diff
@@ -39,6 +39,10 @@ class Avram::BaseQueryTemplate
           end
         {% end %}
 
+        {% if columns.map(&.[:name].stringify).includes?("updated_at") %}
+          _changes[:updated_at] = Time.adapter.to_db(Time.utc) if updated_at == IGNORE
```
jwoertink, line 43: This reads a little confusing... It looks like it's saying "Only change updated_at if I have updated_at set to ignore". Is it just me, or should this not work? :thinking:  I see specs passed so it must be right.

akadusei: Yes. The idea is that if `update_at` is passed in explicitly, then use that. So only set `updated_at` if it was not passed in, in which case its value would be `IGNORE`.

This means `PostQuery.new.update(title: "New title")` would set `updated_at` to current time, but `PostQuery.new.update(title: "New title", updated_at: 2.days.ago)` would set the `updated_at` to 2 days ago. (I should probably add a spec to cover this case)

Don't merge yet, though. I think the condition should be `if updated_at.is_a?(Avram::Ignore)`. Classes compare by identity, so one `Avram::Nothing.new` would not necessarily equal another `Avram::Nothing.new`. I'll add one more commit to address this.

src/avram/base_query_template.cr
```diff
@@ -40,7 +40,7 @@ class Avram::BaseQueryTemplate
         {% end %}
 
         {% if columns.map(&.[:name].stringify).includes?("updated_at") %}
-          _changes[:updated_at] = Time.adapter.to_db(Time.utc) if updated_at == IGNORE
+          _changes[:updated_at] = Time.adapter.to_db(Time.utc) if updated_at.is_a?(Avram::Nothing)
```
jwoertink, line 43: oohhh :joy:  I see what it's doing now. That's funny that changing it to this made it clearer for me. This is just saying that if you pass in your own updated_at value, then this won't run because your own updated_at won't be a `Nothing`, it'll be a `Time`. Good call on that :+1:

jwoertink (approved):

PR #1094, opened 2025-05-19
Adds some missing types

### LINKED ISSUE #1092: Type all the things
If a method, or argument or anything is missing a type, we should add it in unless the type is just absolutely unwieldy (which might be the case in a few... i.e. some massive crazy recursive deep nested union array/hash).

jwoertink:
Fixes #1092

### DIFF
--- a/src/avram/callbacks/callbacks.cr
+++ b/src/avram/callbacks/callbacks.cr
@@ -22,7 +22,7 @@ module Avram::Callbacks
   # end
   # ```
   macro before_run
-    def before_run
+    def before_run : Nil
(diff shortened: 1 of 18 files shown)


src/avram/callbacks/callbacks.cr
```diff
@@ -74,7 +74,7 @@ module Avram::Callbacks
         ERR
       end
     %}
-    def after_run(%object)
+    def after_run(%object : T) : Nil forall T
```
jwoertink, line 77: This might just be `%object : T` and not need the forall... I'll have to double check this.

PR #1095, opened 2025-05-21
Avoid intermediate strings

akadusei:
This should improve performance. [Ref](https://crystal-lang.org/reference/1.16/guides/performance.html#dont-create-intermediate-strings-when-writing-to-an-io)


jwoertink (approved):
:rocket:

PR #1106, opened 2025-06-13
Fix `Avram::SaveOperation#new_record?` for when id is set explicitly

### LINKED ISSUE #1105: `Avram::SaveOperation#new_record?` returns `false` when primary key is set explicitly
The current implementation of `Avram::SaveOperation#new_record?` checks if the primary key column has a `nil` value in the operation:

https://github.com/luckyframework/avram/blob/374559f79f12014a92980f730471e346bd4065d5/src/avram/save_operation.cr#L269-L271

Therefore, when a create operation sets the primary key, `#new_record?` will always return `false`:

```crystal
operation = SaveUser.new(id: 123, name: "Dan", age: 34, joined_at: Time.utc)
puts operation.new_record? # <= `false`
```

Maybe we should be checking `.original_value.nil?` instead of `.value.nil?`?

akadusei:
Fixes #1105

### DIFF
--- a/spec/avram/operations/save_operation_spec.cr
+++ b/spec/avram/operations/save_operation_spec.cr
@@ -913,6 +913,14 @@ describe "Avram::SaveOperation" do
         operation.save.should be_true
         operation.new_record?.should be_true
       end
+
+      it "returns 'true' even if primary key is set" do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
:rocket:

PR #1109, opened 2025-06-26
Updating crystal and postgres in local docker setup.

russ:
- Updating to crystal 1.15.1 and postgres 16
- Using the main branch of lucky_cli


Dockerfile
```diff
@@ -1,18 +1,19 @@
-FROM crystallang/crystal:1.14.0
+FROM crystallang/crystal:1.15.1
```
jwoertink, line 1:
```suggestion
FROM crystallang/crystal:latest
```

Maybe if we just always set this to `latest` then we can ensure that everything works on that and don't have to worry about keeping this up-to-date? The only tricky part is testing the lower bounds. You may write something that works in Crystal 1.17, but fails in 1.15 and wouldn't know until the CI run :thinking:

jwoertink (commented):
I'm good with either way here. What are your thoughts on the Crystal version?

Dockerfile
russ: I think that makes sense for now. Ultimate solution would be a action/test matrix...

PR #1123, opened 2025-11-14
Use `docker compose` instead of `docker-compose` in scripts/

wout:
While I'm at it updating the `docker compose` commands here as well. Similar to https://github.com/luckyframework/lucky/pull/1997.


jwoertink (approved):
Thank you!

Issue #1131, opened 2026-01-10
Add support for `has_one` with `through: ...`


wout:
I just noticed that `has_one` has no `through` option, while `has_many` does. Would it be an idea to add this? Unless there's a valid reason not to of course...

wout:
Great, I'll take a stab at it then! 🙂 It's probably not a very common thing, but I'll need it in at least a handful of models now.

PR #1132, opened 2026-01-11
Add through option for has one

wout:
This implements #1131. I've tried to stick as close as possible to the `has_many` class and reused existing test objects where possible.

### DIFF
--- a/src/avram/associations/has_one.cr
+++ b/src/avram/associations/has_one.cr
@@ -12,9 +12,9 @@ module Avram::Associations::HasOne
       ERROR
       %}
     {% end %}
-    {% if !through.is_a?(NilLiteral) && through.size < 2 %}
+    {% if !through.is_a?(NilLiteral) && through.size != 2 %}
       {% through.raise <<-ERROR
-      'through' on #{@type.name} must be given at least two items. Instead, got: #{through}
+      'through' on #{@type.name} must be given exactly two items. Instead, got: #{through}
 
       Example...
          has_one profile : Profile
--- a/src/avram/associations/has_many.cr
+++ b/src/avram/associations/has_many.cr
@@ -12,9 +12,9 @@ module Avram::Associations::HasMany
       ERROR
       %}
     {% end %}
-    {% if !through.is_a?(NilLiteral) && through.size < 2 %}
+    {% if !through.is_a?(NilLiteral) && through.size != 2 %}
       {% through.raise <<-ERROR
-      'through' on #{@type.name} must be given at least two items. Instead, got: #{through}
+      'through' on #{@type.name} must be given exactly two items. Instead, got: #{through}
 
       Example...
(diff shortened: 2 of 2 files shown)


src/avram/associations/has_one.cr
```diff
@@ -1,5 +1,29 @@
 module Avram::Associations::HasOne
-  macro has_one(type_declaration, foreign_key = nil)
+  macro has_one(type_declaration, through = nil, foreign_key = nil)
```
jwoertink: I'm not sure how common it is, but if someone has code like

```crystal
has_one thing : Thing, :foo_id
```

then this would now break, right? It's probably fine and we can just make note for the `UPGRADE_NOTES` as a step to take when upgrading.

jwoertink (approved):
I think this all looks good. Thanks for taking this on!

src/avram/associations/has_one.cr
wout: Hmm, you're right. I maintained the same order as `has_many`, but maybe the parameters should be reversed so that `through` comes last, in both cases?  I can't imagine people using the `through:` option without the key for `has_many`. What do you think?

wout: I've switched them around and all tests still pass.

src/avram/associations/has_one.cr
```diff
@@ -1,5 +1,29 @@
...
+
+      Example...
+         has_one profile : Profile
+         has_one avatar : Avatar, through: [:profile, :avatar]
+
+      Learn more about associations: https://luckyframework.org/guides/database/models#model-associations
+      ERROR
+      %}
+    {% end %}
+    {% if !through.is_a?(NilLiteral) && through.size < 2 %}
```
matthewmcgarvey: I think this error is a little misleading. This can't handle a through array with more than two items either, can it? It looks like has_many has the same problem

spec/support/models/customer.cr
```diff
@@ -3,5 +3,6 @@ class Customer < BaseModel
     column name : String
     belongs_to employee : Employee
     has_many managers : Manager, through: [:employee, :manager]
+    has_one manager : Manager?, through: [:employee, :manager]
```
matthewmcgarvey: Could you also add this to TaxId? That way it makes sure a singular association also works where this one loads an array and takes the first.

```crystal
has_one email_address : EmailAddress, through: [:business, :email_address]
```

src/avram/associations/has_one.cr
wout: I just followed the same patterns as the `has_many` variant but you're right, that should probably test for exactly two items:
```crystal
{% if !through.is_a?(NilLiteral) && through.size != 2 %}
```
With the message
```crystal
 'through' on #{@type.name} must be given exctly two items. Instead, got: #{through}
```

PR #1138, opened 2026-01-30
Avoid needless heap allocations

akadusei:
General performance enhancements.

Similar to #1137

### DIFF
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -101,7 +101,7 @@ class Avram::QueryBuilder
       else
         value.to_s
       end
-    end.to_a
+    end
   end
 
   private def set_sql_clause(params) : String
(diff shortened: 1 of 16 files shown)


src/avram/query_builder.cr
```diff
@@ -181,12 +181,15 @@ class Avram::QueryBuilder
...
+    @wheres = @wheres.reject do |clause|
+      clause.is_a?(Avram::Where::SqlClause) && clause.column.to_s == column.to_s
+    end
+
     self
   end
 
   def reset_order : Array(Avram::OrderByClause)
-    @orders.clear
+    @orders = Array(Avram::OrderByClause).new
```
akadusei: This is to avoid eventual memory leak in a long running process; `#clear` does not reduce the size of the underlying buffer.

jwoertink (approved):
Wow. It's interesting seeing the code evolve like this. Thanks for working on it!

PR #1142, opened 2026-03-15
Add avram attachment

wout:
This is the Avram counterpart of https://github.com/luckyframework/lucky/pull/2016. It adds the necessary macros and callbacks for models, save operations, and delete operations.

This PR also adds some validations for `StoredFile` attributes:
```crystal
# MIME type
validate_file_mime_type_of avatar, in: %w[image/png image/jpeg]
validate_file_mime_type_of avatar, with: /image\/.*/

# File size
validate_file_size_of avatar, max: 5_000_000
validate_file_size_of avatar, min: 1_000, max: 5_000_000
validate_file_size_of avatar, min: 1_000
```

### DIFF
--- a/db/migrations/20260314174133_create_attachable_items.cr
+++ b/db/migrations/20260314174133_create_attachable_items.cr
@@ -1,0 +1,14 @@
+class CreateAttachableItems::V20260314174133 < Avram::Migrator::Migration::V1
+  def migrate
+    create table_for(AttachableItem) do
+      primary_key id : Int64
+      add_timestamps
+
+      add image : JSON::Any?
+    end
+  end
+
+  def rollback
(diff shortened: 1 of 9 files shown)


jwoertink:
Does this get closed now too since it's gonna move to a shard? Or do we need this?

wout:
Yeah, I meant to ask. My feeling it that it's better to include it in the shard so everything stays. What do you think?

jwoertink:
Yeah, I guess the code in this PR isn't really useful without the code in the shard?

wout:
Yeah, I've included the Avram setup now in the shard: https://github.com/wout/latch. So this is no longer relevant.

PR #1145, opened 2026-03-19
Require migrations to have a rollback method defined

### LINKED ISSUE #1139: Is the migration `rollback` required or not?
Technically, it is required because if you don't define the method, then try to rollback, it'll call this

https://github.com/luckyframework/avram/blob/be8a22f9a7f6f7cc8d94b4becc236668ec22e740/src/avram/migrator/migration.cr#L54

then throw an error about the method not being defined. However, Avram doesn't enforce this method to be created. It's not an abstract def.

Now, with that said, I have quite a few migrations where I update production data, and there's really no way to roll that back. 

We should either

A) Make defining `rollback` required with the `abstract def rollback`
B) Define a default empty `rollback` method so you don't need to define one if it's doing nothing.

jwoertink:
Fixes #1139

### DIFF
--- a/src/avram/migrator/migration.cr
+++ b/src/avram/migrator/migration.cr
@@ -32,6 +32,7 @@ abstract class Avram::Migrator::Migration::V1
   end
 
   abstract def migrate
+  abstract def rollback
   abstract def version : Int64
 
   getter prepared_statements = [] of String


jwoertink:
That CI failure should go away after next crystal nightly build

PR #1149, opened 2026-04-29
Raise compile errors on relevant type node

akadusei:


jwoertink (approved):
You know, I've never really understood what this does 🤔  Does it just make the stack trace contain a better set of line numbers or something? Should all compile-time exceptions always raised on the type node when the error is based around a specific node?

akadusei:
Before:

```
In src/avram/callbacks/after_commit_callback.cr:92:3

 92 | after_commit { }
      ^-----------
Error: The 'after_commit' callback requires exactly 1 block arg to be passed.
```

After:

```
In src/avram/callbacks/after_commit_callback.cr:92:16

 92 | after_commit { }
                   ^
Error: The 'after_commit' callback requires exactly 1 block arg to be passed.
```

Basically *points* to the **exact** node causing the error.

akadusei:
Before:

```
In src/avram/define_attribute.cr:91:3

 91 | attribute aaa : String?
      ^--------
Error: `attribute` in Avram::DefineAttribute must not be called with a type union or nilable type but was called with String | ::Nil
```

After:

```
In src/avram/define_attribute.cr:93:19

 93 | attribute aaa : String?
                      ^
Error: `attribute` in Avram::DefineAttribute must not be called with a type union or nilable type but was called with String | ::Nil
```

The *pointer* points to the `String?` rather than the whole method/macro. Helpful if there's a lot of code in there

jwoertink:
ooohh! Ok, that's awesome. I never realized that. Thanks for explaining.

PR #1150, opened 2026-04-29
Forward original exceptions when re-raised

akadusei:
Do not swallow original exception when re-raising


jwoertink (approved):
This makes a lot of sense 👍

Issue #874, opened 2022-08-12
Checking `new_record?` when using upsert


jwoertink:
I believe that `new_record?` will always be true when using `upsert` because `new_record?` basically does `id.value.nil?`. The `id.value` is only filled in when you pass in `@record` which doesn't happen on upserts.

I'll have to look more in to this, but we had an issue using `new_record?` with the upsert, and it seemed that it was always returning true no matter what.

jinn999:
Uhm, I'm not really sure I understood the problem, but adding the last line in the tests (for an upsert->update operation)
passes
```
    UpsertUserOperation.upsert(
          name: "Rich",
          nickname: nil,
          age: 30,
          joined_at: joined_at
        ) do |operation, user|
          operation.created?.should be_false
          operation.updated?.should be_true
          operation.new_record?.should be_false #=> passes
```
Since I had to put the hands on it quite recently, upsert works this way. 
The id is not filled initially if other columns are used for the lookup (in this case, `name` and `nickname`)
A query is performed to get the record that matches it - and then if the record is present - it is moved into the `@record` variable.
The only possible case in which I can see that being prefilled, is when the lookup columns coincide with the primary key ones (well, you really should directly use `#save` in that case, imho) <- needs more investigation

Issue #790, opened 2022-01-12
Rename current #upsert(!) methods


grepsedawk:
As presented in https://github.com/luckyframework/avram/pull/789 , the upserts in current avram are more of "find_or_create", instead of utilizing `upsert` from postgres: https://www.postgresql.org/docs/9.5/sql-insert.html (INSERT INTO ... ON CONFLICT UPDATE)

This can be confusing to newcomers expecting upsert to be a single UPSERT query.

PR #1151, opened 2026-04-29
Redo operation upserts to use native `ON CONFLICT DO UPDATE`

### LINKED ISSUE #917: If validations set upsert lookup column value, existing record is not found
I have a scenario where my model has a `url : String` field, which is the lookup column `upsert_lookup_columns :url`. This field is set indirectly through an attribute on the save operation:

```crystal
  attribute project : String

  before_save validate_format_of project, with: /^[a-z0-9_-]+\/[a-z0-9_-]+$/i
  before_save validate_project

  def validate_project
    # simplifed example
    url.value = "https://example.com/#{project.value}"
  end
```

For this model `upsert` is not working properly. It always attempts to insert a new record, resulting in a unique key violation.

The issue seems to be that the `upsert` method tries to retrieve the existing record before the validations have been run, and thus in my case `url` is still `nil`:

https://github.com/luckyframework/avram/blob/0ea2a6140f5f9ac770eeaa5badf1d479a5b60f69/src/avram/upsert.cr#L72-L82

It does this query:

```sql
SELECT feeds.id, feeds.created_at, feeds.updated_at, feeds.refreshed_at, feeds.title, feeds.url, feeds.etag, feeds.last_modified, feeds.last_refresh_hash, feeds.source_slug FROM feeds WHERE feeds.url IS NULL ORDER BY feeds.id ASC LIMIT 1
```

Note the `WHERE feeds.url IS NULL` bit.

It gets no record and then proceeds to call save, which attempts to insert a new record. However, as part of `save` it first calls `valid?`, which in turn populates `url` and then when it comes to persist it, a duplicate row is attempted to be inserted.

### LINKED ISSUE #963: Upserting using insert on conflict
I was looking into `upsert` for something I am working on and noticed that it is actually a `find` followed by `create` or `update` rather than a "true" upsert. This means it is prone to race conditions as we may have inserted a conflicting record between the `find` and the subsequent action. 

For my use-case ideally I want to use `INSERT ... ON CONFLICT UPDATE SET foo=$1` to do the upsert in a single query. The downside is that we wouldn't run any of the `after` callbacks as we will no longer know whether it was an `update` or a `save`. This matches the behaviour provided by other frameworks like Rails (https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/upsert). 

Looking for feedback whether this change would be accepted into `Avram` and along those lines should it replace the existing `upsert` method (although this would be a breaking change due to callbacks) or be a new API entirely?

akadusei:
SQL-native upsert, using `ON CONFLICT DO UPDATE`

Closes #963
Fixes #917
Related #790
Related #874

A few caveats:

- Upserts now require actual unique columns. the previous implementation worked without them
- In PostgreSQL, `NULL` is **not** equal to `NULL`, so a nullable composite index may create the same row many times if at least one of the index columns value is null
- `#updated?` && `#created?` return `false` for upserts. (there's currently no way to tell them apart)
- I changed #new_record? to mean create? or upsert?. I'm not sure if that's what's expected

There may be edge cases I did not think about, so scrutinize this very closely.

### DIFF
--- a/db/migrations/20260427115611_make_tokens_next_id_unique.cr
+++ b/db/migrations/20260427115611_make_tokens_next_id_unique.cr
@@ -1,0 +1,9 @@
+class MakeTokensNextIdUnique::V20260427115611 < Avram::Migrator::Migration::V1
+  def migrate
+    create_index table_for(Token), :next_id, unique: true
+  end
+
+  def rollback
+    drop_index table_for(Token), :next_id, if_exists: true
+  end
+end
--- a/db/migrations/20260427120304_make_users_name_unique.cr
+++ b/db/migrations/20260427120304_make_users_name_unique.cr
@@ -1,0 +1,9 @@
+class MakeUsersNameUnique::V20260427120304 < Avram::Migrator::Migration::V1
+  def migrate
+    create_index table_for(Users), [:name, :nickname], unique: true
+  end
+
+  def rollback
+    drop_index table_for(Users), [:name, :nickname], if_exists: true
+  end
+end
--- a/spec/avram/insert_spec.cr
+++ b/spec/avram/insert_spec.cr
@@ -5,15 +5,50 @@ describe Avram::Insert do
     it "inserts with a hash of String" do
       params = {:first_name => "Paul", :last_name => "Smith"}
       insert = Avram::Insert.new(table: :users, params: params)
-      insert.statement.should eq %(insert into users("first_name", "last_name") values($1, $2) returning *)
+
+      insert.statement.should eq %(INSERT INTO users ("first_name", "last_name") \
+        VALUES ($1, $2) RETURNING *)
+
       insert.args.should eq ["Paul", "Smith"]
     end
 
     it "inserts with a hash of Nil" do
       params = {:first_name => nil}
       insert = Avram::Insert.new(table: :users, params: params)
-      insert.statement.should eq %(insert into users("first_name") values($1) returning *)
+      insert.statement.should eq %(INSERT INTO users ("first_name") VALUES ($1) RETURNING *)
       insert.args.should eq [nil]
+    end
+  end
+
+  describe "upserting" do
+    it "updates when keys conflict" do
+      params = {:first_name => "Paul", :last_name => "Smith"}
+
+      insert = Avram::Insert.new(
+        :users,
+        params,
+        conflict_action: :update,
+        conflict_keys: [:first_name, :last_name],
+        conflict_params: [:last_name]
+      )
+
+      insert.statement.should eq %(INSERT INTO users ("first_name", "last_name") \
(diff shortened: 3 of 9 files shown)


akadusei:
check_fomat job failure is from commit 3813ee3f83a1fc03350aa4baeb440c88f8c7671f. Not sure why it's failing now

jwoertink (commented):
Look at that! 😄  I think this is probably the right way to go. I looked over it a little, but I'll need to really dig into how my apps use the upserts to see what all this will change.

I think the main thing I'm a little nervous about is having a SaveOperation that changes semantics internally based on how it's called externally.

Correct me if I'm wrong here, but it seems like now you may run into this?

```crystal
class SaveThing < Thing::SaveOperation
  upsert_lookup_columns :key1, :key2
  permit_columns :key1, :key2

  after_commit(if: :new_record?) do |thing|
    SendNotificationEmail.new(thing).deliver_later
  end
end

# sends the email
SaveThing.create!(params)
# doesn't send the email
SaveThing.update!(thing, params)

# should send the email, but doesn't, and gives no
# warning to let you know until 9 hours of debugging production
SaveThing.upsert(key1: "doesn't exist", key2: "doesn't exist")
```

Or in other words, someone updates Avram, and expects the callbacks to work as they did, but instead of development-time catches indicating a refactor is required, it just goes out to production and quietly misbehaves. 

Now, if that's not the case here, the this is extra great! But we will need to play around with it a bit more to make sure those cases are considered. Thanks for putting this together 🙌

akadusei:
Your specific example should work OK. `#new_record?` is always `true` for upsert. The upsert implementation on `main` branch can distinguish between when a record is created vs updated, I believe. This one doesn't.

Flipping your example around, if `key1` and `key2` exists, upsert will find and update the record, but there is currently no way of knowing that it was an update rather than a create. So `#new_record?` would still return `true` and the email would be delivered.

akadusei:
I am running into a bug where setting the operation outside `.new` call causes updates to `nil` to fail:

```crystal
user = UserFactory.create &.name("test").nickname("sample").total_score(100)

operation = UpsertUserOperation.new(
  name: user.name,
  nickname: user.nickname,
  age: user.age,
  joined_at: user.joined_at
)

operation.total_score.value = nil # <= Set total_score to `nil` here

operation.upsert.should be_true
operation.upserted?.should be_true
operation.saved?.should be_true
operation.record.should_not be_nil

operation.record.try do |saved_user|
  saved_user.name.should eq("test")
  saved_user.total_score.should be_nil # <= This fails
end
```

This applies only when updating to `nil`. Any valid non-nil value passes.

From calling `#compact!` here >> https://github.com/luckyframework/avram/blob/8a48211143e8648ba6636800226ebb698c7f81f4/src/avram/upsert.cr#L98

Attributes do not have a way of differentiating a `nil` value was actually received vs it is `nil` because nothing was received.

PR #1154, opened 2026-05-30
Support partial indexes with a WHERE clause

### LINKED ISSUE #1023: Support creating INDEX with a WHERE
https://github.com/luckyframework/avram/blob/main/src/avram/migrator/create_index_statement.cr

Not sure how this would look, but maybe something like

```crystal
create_index :servers, [:enabled, :kind] where: ServerQuery.new.enabled(true)
# CREATE INDEX ... ON users (enabled, kind)
# WHERE enabled = true
```

The only thing with this is it ties your migrations to `ServerQuery`. If you remove that, then this breaks.

grepsedawk:
Adds a `where:` option to `create_index` and `add_index` so migrations can create PostgreSQL **partial indexes** — including conditional unique indexes (`CREATE UNIQUE INDEX ... WHERE ...`), the original ask.

Fixes #1023

## API

`where:` accepts a raw SQL predicate `String` or an `Avram::Queryable`:

```crystal
# raw SQL — self-contained, cleanest for comparisons
create_index :servers, [:enabled, :kind], unique: true, where: "enabled = true"
# => CREATE UNIQUE INDEX servers_enabled_kind_index ON servers USING btree ("enabled", "kind") WHERE enabled = true;

# query object — values inlined as literals
add_index :enabled, where: ServerQuery.new.enabled(true)
```

DDL predicates can't use bind parameters, so a query is rendered with its values inlined (new `QueryBuilder#to_prepared_where_sql`). Only the query's `where` conditions are used (`select`/`order`/`limit` are ignored); a query with no conditions raises. The raw-`String` form avoids coupling a migration to an app query class; the query form is offered for ergonomics.

## Implementation

- `QueryBuilder#to_prepared_where_sql` returns the `WHERE` conditions as a bare, inlined predicate; `wheres_sql` now builds on the same `where_predicate_sql!`, so there's a single source of truth.
- `CreateIndexStatement` emits ` WHERE <predicate>` after the column list (composes with `unique`, `concurrently`, custom `name`, all index types).
- `create_index` / `add_index` normalize the `String | Avram::Queryable` option through one shared helper.

## Tests

Specs cover both forms through `create_index` and `add_index`, the `QueryBuilder`/`Queryable` predicate extraction (boolean inlining, `OR` conjunctions, `order`/`limit` ignored, raises-on-empty), and the conditional-unique SQL.

🤖 PR Body Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/avram/query_builder.cr
+++ b/src/avram/query_builder.cr
@@ -26,14 +26,20 @@ class Avram::QueryBuilder
   # Prepares the SQL statement by combining the `args` and `statement`
   # in to a single `String`
   def to_prepared_sql : String
+    inline_prepared_args(statement)
+  end
+
+  # Bare WHERE predicate (no "WHERE " prefix), values inlined as literals since
+  # DDL index predicates can't use bind params. Raises if there are no conditions.
+  def to_prepared_where_sql : String
+    predicate = clone.where_predicate_sql!
+    raise Avram::InvalidQueryError.new("Cannot build a partial index predicate: the query has no `where` conditions") unless predicate
+    inline_prepared_args(predicate)
+  end
+
+  private def inline_prepared_args(sql : String) : String
     params = args.map { |arg| "'#{String.new(PQ::Param.encode(arg).slice)}'" }
-    i = 0
-    sql = statement
-    sql.scan(/\$\d+/) do |match|
-      sql = sql.sub(match[0], params[i])
-      i += 1
-    end
-    sql
+    sql.gsub(/\$\d+/) { |placeholder| params[placeholder.lchop('$').to_i - 1] }
   end
 
   # Merges the wheres, joins, and orders from the passed in query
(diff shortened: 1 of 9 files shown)


src/avram/query_builder.cr
```diff
@@ -26,9 +26,22 @@ class Avram::QueryBuilder
...
+    inline_prepared_args(statement)
+  end
+
+  # Bare WHERE predicate (no "WHERE " prefix), values inlined as literals since
+  # DDL index predicates can't use bind params. Raises if there are no conditions.
+  def to_prepared_where_sql : String
+    predicate = clone.where_predicate_sql!
+    if predicate.nil?
+      raise Avram::InvalidQueryError.new("Cannot build a partial index predicate: the query has no `where` conditions")
+    end
```
jwoertink: The raised error here points specifically towards an index, but this is now a public method on Builder which someone could run anywhere.

jwoertink (commented):
Thanks for taking this on! This will be a neat feature to get added in.

I do feel like some of this code is a bit unnecessary though. Since you're allowing both raw strings and a query object into the `where`, you don't know if a `WHERE` statement is being passed in or not, so you're having to move a ton of stuff around just to account for that. Then we now have these new bulky named methods that feel a bit out of place... Also, it sort of breaks a bit of the principle of Avram to try and catch errors at compile-time with the raw strings being front and center like that...

What if we did this... The `where` argument takes the query object and that's it. Then a second arg called `where_raw` takes in a raw string. Now the method knows when you're doing one over the other which could either call `where.to_prepared_sql` or `WHERE #{where_raw}`. With that, all of the changes to `queryable` and `query_builder` would just go away. That makes the primary target something that's compile-time safety, and we keep the escape hatch available while also making it very clear it's a bit more "dangerous". What are your thoughts on that?

